PR #13, opened 2019-03-15
NFT Module

olivdb:


contracts/modules/NftTransfer.sol
```diff
@@ -0,0 +1,136 @@
...
+    // *************** Events *************************** //
+
+    event NonFungibleTransfer(address indexed wallet, address indexed nftContract, uint256 indexed tokenId, address to, bytes data);    
+
+    // *************** Modifiers *************************** //
+
+    /**
+     * @dev Throws if the caller is not the owner or an authorised module.
+     */
+    modifier onlyOwnerOrModule(BaseWallet _wallet) {
```
juniset: Why do we want another module to call this one? I'm afraid it may open the door to attacks so we need to make sure we do it for a good reason.

contracts/modules/NftTransfer.sol
```diff
@@ -0,0 +1,136 @@
...
+     * @dev An ERC721 smart contract calls this function on the recipient contract
+     * after a `safeTransfer`. If the recipient is a BaseWallet, the call to onERC721Received 
+     * will be forwarded to the method onERC721Received of the present module. 
+     * @param operator The address which called `safeTransferFrom` function
+     * @param from The address which previously owned the token
+     * @param tokenId The NFT identifier which is being transferred
+     * @param data Additional data with no specified format
+     * @return bytes4 `bytes4(keccak256("onERC721Received(address,address,uint256,bytes)"))`
+     */
+    function onERC721Received(address operator, address from, uint256 tokenId, bytes calldata data)
```
juniset: Maybe we could emit a Received event here with (wallet, operator, from, tokenId, data) so as to make it easier for clients to track transfers (i.e. they can just monitor the events of the module instead of monitoring all ERC721 transfer and discarding the non Argent ones).

contracts/modules/NftTransfer.sol
```diff
@@ -0,0 +1,136 @@
...
+                "safeTransferFrom(address,address,uint256,bytes)", address(_wallet), _to, _tokenId, _data);
+        } else {
+            methodData = abi.encodeWithSignature(
+                "transferFrom(address,address,uint256)", address(_wallet), _to, _tokenId);
+        }
+
+        _wallet.invoke(_nftContract, 0, methodData);
+        emit NonFungibleTransfer(address(_wallet), _nftContract, _tokenId, _to, _data);
+    }
+
```
juniset: Shouldn't we add a method to approve transfer by another operator? I know we did not do it for ERC20, so this is more an open question then a statement.

contracts/modules/NftTransfer.sol
```diff
@@ -0,0 +1,136 @@
...
+import "./common/BaseModule.sol";
+import "./common/RelayerModule.sol";
+import "../storage/GuardianStorage.sol";
+
+/**
+ * @title NftTransfer
+ * @dev Module to transfer NFTs (ERC721),
+ * @author Olivier VDB - <olivier@argent.xyz>
+ */
+contract NftTransfer is BaseModule, RelayerModule {
```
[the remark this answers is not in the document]
olivdb: Yes, absolutely. I forgot we had that module :)

contracts/modules/NftTransfer.sol
olivdb: Yes, you are right. And likewise, I think there is no reason for `TokenTransfer.transferToken()` to have this modifier either (maybe it is a remnant from when `DappManager.callContract()` was calling `TokenTransfer.transferToken()` ?). Shall I also change it for `TokenTransfer` so the change is there for the next time we redeploy `TokenTransfer`?

contracts/modules/NftTransfer.sol
olivdb: We could but the event would only be emitted for `safeTransfer` so the client would anyways need to monitor ERC721 Transfer events for when unsafe transfers are performed. Maybe we should leave it like that for now and only add this Received event after the client guys confirm that this would indeed make their lives easier?

contracts/modules/NftTransfer.sol
olivdb: I thought about this but my feeling is that, like for the `ERC20.approve()`, this is not typically something that the token owner would be required to do manually in the mobile app. Instead, this is something that is more likely to be requested by a Dapp (e.g. a DEX) so it is best left in DappManager for now in my opinion. The way I think of it is that the [Token|Nft]Transfer modules should only contain methods that are going to be directly called from native integrations in the mobile app UI. Everything else should be left to the DappManager IMO.

contracts/modules/NftTransfer.sol
juniset: Yes.

contracts/modules/NftTransfer.sol
juniset: Agreed

contracts/modules/NftTransfer.sol
juniset: Agreed.

PR #14, opened 2019-03-28
TMP: Uniswap and Compound

juniset:
Addition UniswapManager.sol


contracts/modules/UniswapManager.sol
```diff
@@ -0,0 +1,159 @@
+pragma solidity ^0.5.4;
+import "../wallet/BaseWallet.sol";
+import "./common/BaseModule.sol";
+import "./common/RelayerModule.sol";
+import "./common/OnlyOwnerModule.sol";
+import "../exchange/ERC20.sol";
+import "../utils/SafeMath.sol";
+import "../utils/SafeMath.sol";
```
olivdb: double import

contracts/modules/UniswapManager.sol
```diff
@@ -0,0 +1,159 @@
...
+     * @param _poolToken The address of the ERC20 token of the pair.
+     * @param _ethAmount The amount of ETH available.
+     * @param _tokenAmount The amount of ERC20 token available.
+     * @return the number of liquidity shares minted. 
+     */
+    function addLiquidityToUniswap(
+        BaseWallet _wallet, 
+        address _poolToken, 
+        uint256 _ethAmount, 
+        uint256 _tokenAmount
```
olivdb: why don't we allow the user to only provide a total value `_totalAmountInEth` that they are willing to deposit into Uniswap, and let the contract take care of calculating how this `_totalAmountInEth` should be divided into an `_ethAmount` and a `_tokenAmount` instead of requiring the user to provide both an `_ethAmount` and a `_tokenAmount`?

contracts/modules/UniswapManager.sol
```diff
@@ -0,0 +1,159 @@
...
+        require(pool != address(0), "UM: The target token is not traded on Uniswap");
+
+        uint256 ethPoolSize = address(pool).balance;
+        uint256 tokenPoolSize = ERC20(_poolToken).balanceOf(pool);
+
+        uint256 tokenInEth = getInputToOutputPrice(_tokenAmount, tokenPoolSize, ethPoolSize);
+        if(_ethAmount >= tokenInEth) {
+            // swap some eth for tokens
+            (uint256 ethSwap, uint256 ethPool, uint256 tokenPool) = computePooledValue(ethPoolSize, tokenPoolSize, _ethAmount, _tokenAmount, tokenInEth);
+            _wallet.invoke(pool, ethSwap, abi.encodeWithSignature("ethToTokenSwapInput(uint256,uint256)", 1, block.timestamp));
```
olivdb: shouldn't we do the trade via our kyber module instead, to offer better rates to our users?

contracts/modules/UniswapManager.sol
```diff
@@ -0,0 +1,159 @@
...
+            // add liquidity
+            _wallet.invoke(pool, ethPool, abi.encodeWithSignature("addLiquidity(uint256,uint256,uint256)",1, tokenPool, block.timestamp + 1));
+        }
+        else {
+            // swap some tokens for eth
+            (uint256 tokenSwap, uint256 tokenPool, uint256 ethPool) = computePooledValue(tokenPoolSize, ethPoolSize, _tokenAmount, _ethAmount, 0);
+            _wallet.invoke(_poolToken, 0, abi.encodeWithSignature("approve(address,uint256)", pool, tokenSwap + tokenPool));
+            _wallet.invoke(pool, 0, abi.encodeWithSignature("tokenToEthSwapInput(uint256,uint256,uint256)", tokenSwap, 1, block.timestamp));
+            // add liquidity
+            _wallet.invoke(pool, ethPool - 1, abi.encodeWithSignature("addLiquidity(uint256,uint256,uint256)",1, tokenPool, block.timestamp + 1));
```
olivdb: why `ethPool - 1` ?

contracts/modules/UniswapManager.sol
```diff
@@ -0,0 +1,159 @@
...
+        view 
+        returns(uint256 _majorSwap, uint256 _majorPool, uint256 _minorPool) 
+    {
+        if(_minorInMajor == 0) {
+            _minorInMajor = getInputToOutputPrice(_minorAmount, _minorPoolSize, _majorPoolSize); 
+        }
+        _majorSwap = (_majorAmount.sub(_minorInMajor)).mul(1003).div(2000);
+        uint256 minorSwap = getInputToOutputPrice(_majorSwap, _majorPoolSize, _minorPoolSize);
+        _majorPool = _majorAmount.sub(_majorSwap);
+        _minorPool = _majorPool.mul(_minorPoolSize.sub(minorSwap)).div(_majorPoolSize.add(_majorSwap)) + 1;
```
olivdb: why the `+ 1` ?

contracts/modules/UniswapManager.sol
```diff
@@ -0,0 +1,159 @@
...
+     * @dev Computes the amount of output tokens that can be obtained by swapping the provided amoutn of input.
+     * @param _inputAmount The amount of input token.
+     * @param _inputPoolSize The size of the input pool.
+     * @param _outputPoolSize The size of the output pool.
+     */
+    function getInputToOutputPrice(uint256 _inputAmount, uint256 _inputPoolSize, uint256 _outputPoolSize) internal view returns(uint256) {
+        if(_inputAmount == 0) {
+            return 0;
+        }
+        uint256 inputWithFee = _inputAmount.mul(997);
```
olivdb: maybe clearer to call this `inputAfterFee` or `inputMinusFee`?

contracts/test/TestERC20.sol
```diff
@@ -1,80 +1,32 @@
...
-        return true;
+    function burnFrom(address account, uint256 amount) public {
+        _burnFrom(account, amount);
     }
 
-    function approve(address spender, uint tokens) public returns (bool success) {
-        allowed[msg.sender][spender] = tokens;
-        emit Approval(msg.sender, spender, tokens);
-        return true;
+    function approveInternal(address owner, address spender, uint256 value) public {
```
olivdb: why do we need that?

contracts/test/TestERC20.sol
```diff
@@ -1,80 +1,32 @@
...
+        _approve(owner, spender, value);
```
olivdb: this method doesn't exist in my version of openzeppelin/.../ERC20.sol

contracts/modules/UniswapManager.sol
```diff
@@ -0,0 +1,159 @@
...
+        uint256 _minorInMajor
+    ) 
+        internal 
+        view 
+        returns(uint256 _majorSwap, uint256 _majorPool, uint256 _minorPool) 
+    {
+        if(_minorInMajor == 0) {
+            _minorInMajor = getInputToOutputPrice(_minorAmount, _minorPoolSize, _majorPoolSize); 
+        }
+        _majorSwap = (_majorAmount.sub(_minorInMajor)).mul(1003).div(2000);
```
olivdb: I think the correct coefficient here is `1000/1997` instead of `1003/2000`:

We are looking for `_majorSwap` such that `_majorAmount - _majorSwap =  _minorInMajor + _majorSwap*.997`. Solving for `_majorSwap` gives `_majorSwap = (_majorAmount - _minorInMajor)/(1+.997)`

contracts/modules/UniswapManager.sol
juniset: Indeed. Fixed.

contracts/modules/UniswapManager.sol
juniset: The idea is to have maximum flexibility so the user can provide any amount of ETH and token and the contract will take care of the rest. So `_ethAmount` or `_tokenAmount` can be zero (which is the scenario you describe) but it doesn't need to be.

Note that we may choose to only allow the user to provide only ETH or token at the client level but at least the contracts can handle all scenarios.

contracts/modules/UniswapManager.sol
juniset: It's a fair point and I thought about it.

Kyber: simpler math (the swap does not affect the size of the pool) and potentially a better price (although arbitrage should always mitigate that).  

Uniswap: much cheaper in terms of gas (~40k vs 150k)... and I kind of liked the idea of doing everything with Uniswap :-).

contracts/modules/UniswapManager.sol
juniset: Agreed and changed.

contracts/test/TestERC20.sol
juniset: To be honest I just copied their mock contract https://github.com/OpenZeppelin/openzeppelin-solidity/blob/master/contracts/mocks/ERC20Mock.sol without giving it much thought. I'll remove it.

contracts/test/TestERC20.sol
juniset: Same as above, I will remove it.

contracts/modules/UniswapManager.sol
juniset: Yes you are right indeed. This is what I had and for some reason I decided to change it.

contracts/modules/UniswapManager.sol
juniset: We are trying to satisfy the condition at L57 of uniswap_exchange.vy which translates to:
```tokenPool >= ethPool * token_reserve / eth_reserve + 1``` 
The `+1` in the expression above comes from the integer division.

The `ethPool - 1` and `ethPool + 1` where naive attempts to take this `+1` into account. 

To be more consistent I have now updated the code to replace `tokenPool` by `tokenPool + 1` when the quantities have been computed.

File: contracts/modules/UniswapManager.sol (hunk 5 above)
Comment (juniset) [reply contracts/modules/UniswapManager.sol]: Same as above.

File: contracts/modules/UniswapManager.sol (hunk 3 above)
Comment (olivdb) [reply contracts/modules/UniswapManager.sol]: Ok fair enough.
Another thing: I think the call to `_wallet.invoke(pool, ethSwap, abi.encodeWithSignature("ethToTokenSwapInput(uint256,uint256)", 1, block.timestamp)); ` should be wrapped in an `if(ethSwap > 0){}` because in the (rare) cases where `ethSwap == 0`,   `ethToTokenSwapInput` will revert (see L128 in uniswap_exchange.vy).

File: contracts/modules/UniswapManager.sol [hunk 10]
```diff
@@ -0,0 +1,158 @@
...
+            _wallet.invoke(pool, ethSwap, abi.encodeWithSignature("ethToTokenSwapInput(uint256,uint256)", 1, block.timestamp));
+            _wallet.invoke(_poolToken, 0, abi.encodeWithSignature("approve(address,uint256)", pool, tokenPool + 1));
+            // add liquidity
+            _wallet.invoke(pool, ethPool, abi.encodeWithSignature("addLiquidity(uint256,uint256,uint256)",1, tokenPool + 1, block.timestamp + 1));
+        }
+        else {
+            // swap some tokens for eth
+            (uint256 tokenSwap, uint256 tokenPool, uint256 ethPool) = computePooledValue(tokenPoolSize, ethPoolSize, _tokenAmount, _ethAmount, 0);
+            _wallet.invoke(_poolToken, 0, abi.encodeWithSignature("approve(address,uint256)", pool, tokenSwap + tokenPool + 1));
+            _wallet.invoke(pool, 0, abi.encodeWithSignature("tokenToEthSwapInput(uint256,uint256,uint256)", tokenSwap, 1, block.timestamp));
```
olivdb: I think the call to `_wallet.invoke(pool, 0, abi.encodeWithSignature("tokenToEthSwapInput(uint256,uint256,uint256)", tokenSwap, 1, block.timestamp)); ` should be wrapped in an `if(tokenSwap > 0){}` because in the (rare) cases where `tokenSwap == 0`,   `tokenToEthSwapInput ` will revert (see L203 in uniswap_exchange.vy)

contracts/modules/UniswapManager.sol
olivdb: One problem I have with this approach is that a swap is almost always required even if the user owns enough ETH and tokens to add liquidity without having to do any swap. 

Example:
Assume 1 ETH = 100 DAI. User owns 100 ETH and 10,000 DAI. User wants to put ~2ETH in Uniswap. Client calls `addLiquidityToUniswap` with `_poolToken=100 DAI` and `_ethAmount=1ETH`. There will most likely be some swap going on, even though the user could have just used his existing ETH and DAI balances to do the operation.

The only case where a swap is not required is (assuming we add the `if` wrapping iI mentioned in my other comments) if the client manages to perfectly calculate `_ethAmount` and `_tokenAmount` such that `ethSwap` or `tokenSwap` is zero. But that would make the client implementation complicated and not very robust to changes given all the fees, rounding up, etc.

olivdb: Another approach could be:
1. client just specifies how much ETH he wants to add to the pool
2. contract calculates how many tokens are required
3. if user's token balance is lower than amount of tokens required, convert some ether to get the required balance
4. perform addLiquidity

juniset: As discussed you have a fair point. In the end it boils down to the the UX we want to offer to our users. Since that UX is not properly defined yet I'm in favor of an implementation that covers as many use cases as possible.

To keep the current implementation while addressing your comment I have added a boolean flag `preventSwap` to the `addLiquidityWithUniswap()` method that can prevent swaps from happening by finding the lowest in value between `_ethAmount` and `_tokenAmount` and add liquidity based on that reference.

contracts/modules/UniswapManager.sol
juniset: Agreed and done.

contracts/modules/UniswapManager.sol
juniset: Same as above. Agreed and done.

PR #16, opened 2019-04-18
Feat upgrade

juniset:
Addition of scripts to enable wallet upgrade by deploying/registering the new modules and deploying/registering the associated Upgraders.


deployment/6_register_modules_v1.js
```diff
@@ -61,8 +62,23 @@ const deploy = async (network, secret) => {
 
     for (let idx = 0; idx < wrappers.length; idx++) {
```
olivdb: can be removed I guess

deployment/7_upgrade_v2.js
```diff
@@ -0,0 +1,133 @@
...
+        await multisigExecutor.executeCall(ModuleRegistryWrapper, "registerModule", [wrapper.contractAddress, utils.asciiToBytes32(wrapper._contract.contractName)]);
+    }
+
+    ////////////////////////////////////
+    // Deploy and Register upgraders
+    ////////////////////////////////////
+
+    const toAdd = newModuleWrappers.map((wrapper) => {
+        return {
+            'name': wrapper._contract.contractName,
```
olivdb: I would put 'address' first and 'name' second to produce a more consistent json file (see L73 of 6_register_modules_v1.js)

deployment/7_upgrade_v2.js
```diff
@@ -0,0 +1,133 @@
...
+            'address': wrapper.contractAddress
+        };
+    });
+    let fingerprint; 
+
+    const versions = await versionUploader.load(BACKWARD_COMPATIBILITY);
+    for (let idx = 0; idx < versions.length; idx++) {
+        const version = versions[idx];
+        const moduleNames = MODULES_TO_DISABLE.concat(MODULES_TO_ENABLE);
+        const toRemove = version.modules.filter(module => moduleNames.includes(module.name));
```
olivdb: shouldn't we use `MODULES_TO_DISABLE` instead of `moduleNames` here? (even though it makes not difference for that particular upgrade)

deployment/7_upgrade_v2.js
```diff
@@ -0,0 +1,133 @@
...
+        const toKeep = version.modules.filter(module => !moduleNames.includes(module.name));
```
olivdb: shouldn't we use `MODULES_TO_DISABLE` instead of `moduleNames` here? (even though it makes not difference for that particular upgrade)

deployment/7_upgrade_v2.js
```diff
@@ -0,0 +1,133 @@
...
+
+    const toAdd = newModuleWrappers.map((wrapper) => {
+        return {
+            'name': wrapper._contract.contractName,
+            'address': wrapper.contractAddress
+        };
+    });
+    let fingerprint; 
+
+    const versions = await versionUploader.load(BACKWARD_COMPATIBILITY);
```
olivdb: What exactly does this load?

deployment/7_upgrade_v2.js
```diff
@@ -0,0 +1,133 @@
...
+    for (let idx = 0; idx < versions.length; idx++) {
+        const version = versions[idx];
+        const moduleNames = MODULES_TO_DISABLE.concat(MODULES_TO_ENABLE);
+        const toRemove = version.modules.filter(module => moduleNames.includes(module.name));
+        const toKeep = version.modules.filter(module => !moduleNames.includes(module.name));
+        if(idx == 0) {
```
olivdb: Why this if?

deployment/6_register_modules_v1.js
```diff
@@ -61,8 +62,23 @@ const deploy = async (network, secret) => {
...
         let wrapper = wrappers[idx];
-        await multisigExecutor.executeCall(ModuleRegistryWrapper, "registerModule", [wrapper.contractAddress, utils.asciiToBytes32(wrapper._contract.contractName)]);
+        //await multisigExecutor.executeCall(ModuleRegistryWrapper, "registerModule", [wrapper.contractAddress, utils.asciiToBytes32(wrapper._contract.contractName)]);
```
olivdb: is this intentionally commented out?

juniset: No, it's a left-over from the debug phase. I will uncomment.

deployment/7_upgrade_v2.js
juniset: Done

deployment/7_upgrade_v2.js
juniset: It loads the Version object from the `BACKWARD_COMPATIBILITY` previous versions. We need them to construct and deploy the Upgrader contract that will upgrade wallets from a past version (within `BACKWARD_COMPATIBILITY` range) to the target version.

deployment/7_upgrade_v2.js
juniset: No because we also need to remove the modules that we are adding (if they are present). 

Say we are in `Test` and the target version adds the NftModule. We go from 1.0.x to 1.1.0. We then realise that there is a bug in the NftModule so we update it and make a patch to go to 1.1.1. Wallets that did not upgrade to 1.1.0 will need to add the new version of the NftModule, while wallets that upgraded to 1.1.0 need to remove their current NftModule then add the new version. The current script takes care of these 2 scenarios by concatenating `MODULES_TO_DISABLE` with `MODULES_TO_ENABLE`.

deployment/7_upgrade_v2.js
juniset: Same as above.

deployment/7_upgrade_v2.js
juniset: Because we are potentially loading multiple version objects (if `BACKWARD_COMPATIBILITY` > 1). `idx == 0` means we are working on the latest version and we use that version to create the next version object that will be uploaded to S3.

For the older versions (with `idx > 0`) we only need to create and deploy the Uploader contract.

deployment/7_upgrade_v2.js
```diff
@@ -0,0 +1,133 @@
...
+    // Deploy new modules
+    ////////////////////////////////////
+
+    const NftTransferWrapper = await deployer.deploy(
+        NftTransfer,
+        {},
+        config.contracts.ModuleRegistry,
+        config.modules.GuardianStorage
+    );
+    newModuleWrappers.push(NftTransferWrapper);
```
olivdb: I feel that there is a bit of a mix of generic code and code specific to the NFT update. I would probably try to put the generic code (at least L69-L127, and possibly also the code in L51-L69) in one function that is called with the the names of the modules to enable, the names of the modules to disable, and a dictionary with the module names as keys and the moduleWrappers as values. Something like that...

deployment/7_upgrade_v2.js
olivdb: ok

deployment/7_upgrade_v2.js
olivdb: ok

deployment/7_upgrade_v2.js
juniset: I fully agree, but I want to make sure the entire upgrade process is working end-to-end before investing time into extracting the generic code. I was planning to do that for the next upgrade (pnormally defi).

PR #17, opened 2019-05-23
Supporting CryptoKitty transfers

olivdb:


contracts/modules/NftTransfer.sol
```diff
@@ -110,4 +110,29 @@ contract NftTransfer is BaseModule, RelayerModule, OnlyOwnerModule {
...
 
+    /**
+    * @dev lets the owner transfer a CryptoKitty from a wallet.
+    * @param _wallet The target wallet.
+    * @param _nftContract The CryptoKitty contract address.
+    * @param _to The recipient.
+    * @param _tokenId The kitty id
+    * @param _data The data to pass with the transfer.
+    */
+    function transferCryptoKitty(
```
juniset: This method creates a security breach in our model. Someone could provide the address of an ERC20 instead of the address of the CK contract and `_tokenId` will be interpreted as an amount.... yielding in an ERC20 transfer bypassing our daily limit.

What I think we should do is pass the address of the CK contract in the constructor and update the `transferNFT` method:

``` 
function transferNFT(
        BaseWallet _wallet, 
        address _nftContract, 
        address _to, 
        uint256 _tokenId,
        bool _safe,
        bytes calldata _data
    ) 
        external 
        onlyOwner(_wallet) 
        onlyWhenUnlocked(_wallet)
    {
        bytes memory methodData;
        if(_nftContract == CKADDRESS) {
            methodData = abi.encodeWithSignature("transfer(address,uint256)", _to, _tokenId);
        } else {
           if(_safe) {
               methodData = abi.encodeWithSignature(
                   "safeTransferFrom(address,address,uint256,bytes)", address(_wallet), _to, _tokenId, _data);
           } else {
               methodData = abi.encodeWithSignature(
                   "transferFrom(address,address,uint256)", address(_wallet), _to, _tokenId);
           }
        }
        _wallet.invoke(_nftContract, 0, methodData);
        emit NonFungibleTransfer(address(_wallet), _nftContract, _tokenId, _to, _data);
    }
``` 

This is secure and also easier to use by the clients.

olivdb: Agreed, will change

PR #19, opened 2019-06-20
Defi

juniset:
- InvestManager with Compound and Uniswap
- LoanManager with Compound and Maker
- LeverageManager with Maker/Uniswap


contracts/modules/common/ProviderModule.sol
```diff
@@ -0,0 +1,36 @@
...
+    function removeProvider(address _provider) public onlyOwner {
+        delete providers[_provider];
+    } 
+
+    function isProvider(address _provider) public view returns (bool) {
+        return providers[_provider].exists;
+    }
+
+    function delegateToProvider(address _provider, bytes memory _methodData) internal returns (bool, bytes memory) {
+        require(_provider != address(0), "ProviderModule: Unknown provider");
```
olivdb: This check seems a bit useless given that all the modules seem to always call `isProvider()` before calling `delegateToProvider()`. Why not do `require(isProvider(_provider))` here instead of `require(_provider != address(0))` and remove the `require(isProvider(_provider))` at the beginning of each `openLoan`, `addInvestment`, etc.

contracts/defi/provider/MakerProvider.sol
```diff
@@ -0,0 +1,817 @@
...
+
+   /**
+     * @dev Opens a collateralized loan.
+     * @param _wallet The target wallet.
+     * @param _collateral The token used as a collateral.
+     * @param _collateralAmount The amount of collateral token provided.
+     * @param _debtToken The token borrowed.
+     * @param _debtAmount The amount of tokens borrowed.
+     * @param _oracles (optional) The address of one or more oracles contracts that may be used by the provider to query information on-chain.
+     * @return (optional) An ID for the loan when the provider enables users to create multiple distinct loans.
```
olivdb: * @return (optional) An ID for the loan when the provider enables users to create multiple distinct loans.

->

* @return An ID for the loan

contracts/defi/provider/MakerProvider.sol
```diff
@@ -0,0 +1,817 @@
...
+    /* ********************************** Implementation of Loan ************************************* */
+
+   /**
+     * @dev Opens a collateralized loan.
+     * @param _wallet The target wallet.
+     * @param _collateral The token used as a collateral.
+     * @param _collateralAmount The amount of collateral token provided.
+     * @param _debtToken The token borrowed.
+     * @param _debtAmount The amount of tokens borrowed.
+     * @param _oracles (optional) The address of one or more oracles contracts that may be used by the provider to query information on-chain.
```
olivdb: * @param _oracles (optional) The address of one or more oracles contracts that may be used by the provider to query information on-chain.
->
* @param _oracles An array containing the address of MakerDAO's Tub contract as single element

contracts/defi/provider/MakerProvider.sol
olivdb: idem for other interface methods: i would specify in the documentation how the particular method should be called instead of repeating  the description of the generic parameters introduced in the interface

contracts/defi/provider/MakerProvider.sol
```diff
@@ -0,0 +1,817 @@
...
+     * @return false if the CDP is in danger of being liquidated.
+     */
+    function isSafe(bytes32 _cup, IMakerCdp _makerCdp) public returns (bool) { 
+        return _makerCdp.safe(_cup);
+    }
+
+    /**
+     * @dev Checks if a CDP exists.
+     * @param _cup The id of the CDP.
+     * @return false if the CDP is in danger of being liquidated.
```
olivdb: `ink` is just the amount of collateral in the CDP. Why does `ink == 0` implies a danger of liquidation?

olivdb: if `tab==0` (no debt) and `ink==0` (no collateral), then the CDP is safe

contracts/defi/provider/MakerProvider.sol
```diff
@@ -0,0 +1,817 @@
...
+    ) 
+        external 
+        view 
+        returns (uint8 _status, uint256 _ethValue)
+    {
+        IMakerCdp makerCdp = IMakerCdp(_oracles[0]);
+        if(exists(_loanId, makerCdp)) {
+            return (3,0);
+        }
+        return (0,0);
```
olivdb: I don't understand this function. From what I understand, if the CDP has some collateral, we return "no info", otherwise, we return "no loan", why?

contracts/defi/provider/MakerProvider.sol
```diff
@@ -0,0 +1,817 @@
...
+contract MakerProvider is Loan, Leverage {
+
+    // Mock token address for ETH
+    address constant internal ETH_TOKEN_ADDRESS = 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE;
+
+    // Method signatures to reduce gas cost at depoyment
+    bytes4 constant internal CDP_DRAW = bytes4(keccak256("draw(bytes32,uint256)"));
+    bytes4 constant internal CDP_WIPE = bytes4(keccak256("wipe(bytes32,uint256)"));
+    bytes4 constant internal CDP_SHUT = bytes4(keccak256("shut(bytes32)"));
+    bytes4 constant internal CDP_DEPOSIT = bytes4(keccak256("deposit"));
```
olivdb: that's a method on WETH, so I would call it `WETH_DEPOSIT`

contracts/defi/provider/MakerProvider.sol
```diff
@@ -0,0 +1,817 @@
...
+    bytes4 constant internal CDP_JOIN = bytes4(keccak256("join(uint256)"));
+    bytes4 constant internal CDP_LOCK = bytes4(keccak256("lock(bytes32,uint256)"));
+    bytes4 constant internal CDP_FREE = bytes4(keccak256("free(bytes32,uint256)"));
+    bytes4 constant internal CDP_EXIT = bytes4(keccak256("exit(uint256)"));
+    bytes4 constant internal ERC20_APPROVE = bytes4(keccak256("approve(address,uint256)"));
+    bytes4 constant internal ERC20_WITHDRAW = bytes4(keccak256("withdraw(uint256)"));
```
olivdb: that's a method on WETH, so I would call it `WETH_WITHDRAW`

contracts/defi/provider/MakerProvider.sol
```diff
@@ -0,0 +1,817 @@
...
+
+    /**
+     * @dev Draw more DAI from the CDP and exchange it to collateral.
+     * @param _wallet The target wallet
+     * @param _leverageId The id of the CDP.
+     * @param _availableCollateral The amount of collateral available in the CDP.
+     * @param _daiPerPethRatio The ratio of DAI that can be drawn from 1 PETH.
+     * @param _makerCdp The Maker CDP contract
+     * @param _uniswapFactory The Uniswap Factory contract used for the exchange.
+     * @return the amount of DAI that where drawn from the CDP and the amount of collateral that where added
```
olivdb: where > were/was

contracts/modules/common/ProviderModule.sol
juniset: Agreed and changed.

contracts/defi/provider/MakerProvider.sol
juniset: Agreed and changed.

contracts/defi/provider/MakerProvider.sol
juniset: Agreed and changed.

contracts/defi/provider/MakerProvider.sol
juniset: agreed and changed

contracts/defi/provider/MakerProvider.sol
juniset: Agreed and changed

contracts/defi/provider/MakerProvider.sol
juniset: We cannot wrap the desired methods as the `getLoan` method is `view` and the target methods modify the state. The only info we can provide is if the loan (aka the CDP) exists. If yes we return `status=3` (exists but no info), if no  we return `status=0` (no loan). I've updated the comment to make that clearer.

contracts/defi/provider/MakerProvider.sol
juniset: There is an error in the comment (lousy cut and paste...). Here we only check if the CDP exists, and we do so by checking if there is collateral (no collateral means no CDP). 

I've updated the comment according to the method :-)

contracts/defi/provider/MakerProvider.sol
juniset: Updated.

juniset:
I've addressed all your comments.

I've also commented out the implementation of `Leverage` for `MakerProvider` as it is not part of the next release and should not be audited at this point.

contracts/modules/LoanManager.sol
```diff
@@ -0,0 +1,256 @@
...
+        emit DebtRemoved(address(_wallet), _provider, _loanId, _debtToken, _debtAmount);
+    }
+
+    /**
+     * @dev Gets information about a loan identified by its ID.
+     * @param _wallet The target wallet.
+     * @param _provider The address of the provider to use.
+     * @param _loanId The ID of the loan if any, 0 otherwise.
+     * @return a status [0: no loan, 1: loan is safe, 2: loan is unsafe and can be liquidated, 3: unable to provide info] and the estimated ETH value of the loan
+     * combining all collaterals and all debts. When status = 1 it represents the value that could still be borrowed, while with status = 2
```
olivdb: "and the estimated ETH value of the loan combining all collaterals and all debts" -> "and a value (in ETH) representing the value that could still be borrowed when status = 1; or the value of the collateral that should be added to avoid liquidation when status = 2"

contracts/defi/provider/MakerProvider.sol
```diff
@@ -0,0 +1,818 @@
...
+    }
+
+    /**
+     * @dev Checks if a CDP exists.
+     * @param _cup The id of the CDP.
+     * @param _makerCdp The Maker CDP contract
+     * @return true if the CDP exists, false otherwise.
+     */
+    function exists(bytes32 _cup, IMakerCdp _makerCdp) public view returns (bool) { 
+        return _makerCdp.ink(_cup) != 0;
```
olivdb: I think it's better to check that `_makerCdp.lad(_cup) != address(0)` here instead of `_makerCdp.ink(_cup) != 0` because it's possible for the CDP to have no collateral left (`ink == 0`) but still be opened (it won't be closed until `shut()` is called).

contracts/defi/provider/MakerProvider.sol
olivdb: Ok

contracts/defi/provider/MakerProvider.sol
```diff
@@ -0,0 +1,818 @@
...
+        removeDebt(_wallet, _loanId, _debtAmount, makerCdp, UniswapFactory(_oracles[1]));
+    }
+
+    /**
+     * @dev Gets information about a loan identified by its ID.
+     * @param _wallet The target wallet.
+     * @param _loanId The ID of the target CDP.
+     * @param _oracles An array containing the address of MakerDAO's Tub contract as first element.
+     * @return a status [0: no loan, 1: loan is safe, 2: loan is unsafe and can be liquidated, 3: loan exists but we are unable to provide info] and the estimated ETH value of the loan
+     * combining all collaterals and all debts. When status = 1 it represents the value that could still be borrowed, while with status = 2
```
olivdb: "and the estimated ETH value of the loan combining all collaterals and all debts" -> "and a value (in ETH) representing the value that could still be borrowed when status = 1; or the value of the collateral that should be added to avoid liquidation when status = 2"

contracts/defi/Loan.sol
```diff
@@ -0,0 +1,129 @@
...
+    ) 
+        external;
+
+    /**
+     * @dev Gets information about a loan identified by its ID.
+     * @param _wallet The target wallet.
+     * @param _loanId The ID of the loan if any, 0 otherwise.
+     * @param _oracles (optional) The address of one or more oracles contracts that may be used by the provider to query information on-chain.
+     * @return a status [0: no loan, 1: loan is safe, 2: loan is unsafe and can be liquidated, 3: unable to provide info] and the estimated ETH value of the loan
+     * combining all collaterals and all debts. When status = 1 it represents the value that could still be borrowed, while with status = 2
```
olivdb: "and the estimated ETH value of the loan combining all collaterals and all debts" -> "and a value (in ETH) representing the value that could still be borrowed when status = 1; or the value of the collateral that should be added to avoid liquidation when status = 2"

contracts/defi/provider/CompoundV2Provider.sol
```diff
@@ -0,0 +1,405 @@
...
+        exitMarketIfNeeded(_wallet, dToken, comptroller);
+    }
+
+    /**
+     * @dev Gets information about a loan identified by its ID.
+     * @param _wallet The target wallet.
+     * @param _loanId The ID of the loan if any, 0 otherwise.
+     * @param _oracles (optional) The address of one or more oracles contracts that may be used by the provider to query information on-chain.
+     * @return a status [0: no loan, 1: loan is safe, 2: loan is unsafe and can be liquidated] and the estimated ETH value of the loan
+     * combining all collaterals and all debts. When status = 1 it represents the value that could still be borrowed, while with status = 2
```
olivdb: "and the estimated ETH value of the loan combining all collaterals and all debts" -> "and a value (in ETH) representing the value that could still be borrowed when status = 1; or the value of the collateral that should be added to avoid liquidation when status = 2"

contracts/defi/provider/MakerProvider.sol
```diff
@@ -0,0 +1,818 @@
...
+        BaseWallet _wallet, 
+        address _collateral, 
+        uint256 _collateralAmount, 
+        address _debtToken, 
+        uint256 _debtAmount, 
+        address[] calldata _oracles
+    ) 
+        external 
+        returns (bytes32 _loanId)
+    {
```
olivdb: Would be nice to add `require(_oracles.length == 1)` as you did for the other providers

contracts/defi/provider/MakerProvider.sol
```diff
@@ -0,0 +1,818 @@
...
+    /**
+     * @dev Closes a collateralized loan by repaying all debts (plus interest) and redeeming all collateral (plus interest).
+     * @param _wallet The target wallet.
+     * @param _loanId The ID of the target CDP.
+     * @param _oracles An array containing the address of MakerDAO's Tub contract as first element and Uniswap's Factory as second.
+     */
+    function closeLoan(
+        BaseWallet _wallet, 
+        bytes32 _loanId, 
+        address[] calldata _oracles
```
olivdb: Would be nice to add `require(_oracles.length == 2)` as you did for the other providers

contracts/defi/provider/MakerProvider.sol
```diff
@@ -0,0 +1,818 @@
...
+     */
+    function addCollateral(
+        BaseWallet _wallet, 
+        bytes32 _loanId, 
+        address _collateral, 
+        uint256 _collateralAmount, 
+        address[] calldata _oracles
+    ) 
+        external
+    {
```
olivdb: Would be nice to add `require(_oracles.length == 1)` as you did for the other providers

contracts/defi/provider/MakerProvider.sol
```diff
@@ -0,0 +1,818 @@
...
+     */
+    function removeCollateral(
+        BaseWallet _wallet, 
+        bytes32 _loanId, 
+        address _collateral, 
+        uint256 _collateralAmount, 
+        address[] calldata _oracles
+    ) 
+        external 
+    {
```
olivdb: Would be nice to add `require(_oracles.length == 1)` as you did for the other providers

contracts/defi/provider/MakerProvider.sol
```diff
@@ -0,0 +1,818 @@
...
+     */
+    function addDebt(
+        BaseWallet _wallet, 
+        bytes32 _loanId, 
+        address _debtToken, 
+        uint256 _debtAmount, 
+        address[] calldata _oracles
+    ) 
+        external
+    {
```
olivdb: Would be nice to add `require(_oracles.length == 1)` as you did for the other providers

contracts/defi/provider/MakerProvider.sol
```diff
@@ -0,0 +1,818 @@
...
+     */
+    function removeDebt(
+        BaseWallet _wallet, 
+        bytes32 _loanId, 
+        address _debtToken, 
+        uint256 _debtAmount, 
+        address[] calldata _oracles
+    ) 
+        external
+    {
```
olivdb: Would be nice to add `require(_oracles.length == 2)` as you did for the other providers

contracts/defi/provider/MakerProvider.sol
```diff
@@ -0,0 +1,818 @@
...
+     */
+    function getLoan(
+        BaseWallet _wallet, 
+        bytes32 _loanId, 
+        address[] calldata _oracles
+    ) 
+        external 
+        view 
+        returns (uint8 _status, uint256 _ethValue)
+    {
```
olivdb: Would be nice to add `require(_oracles.length == 1)` as you did for the other providers

contracts/defi/provider/UniswapProvider.sol
```diff
@@ -0,0 +1,155 @@
...
+        address[] calldata _oracles
+    ) 
+        external 
+        view
+        returns (uint256 _tokenValue, uint256 _periodEnd) 
+    {
+        address tokenPool = UniswapFactory(_oracles[0]).getExchange(_token);
+        uint256 tokenPoolSize = ERC20(_token).balanceOf(tokenPool);
+        uint shares = ERC20(tokenPool).balanceOf(address(_wallet));
+        _tokenValue = shares.mul(tokenPoolSize).mul(2);
```
olivdb: Shouldn't we divide this by `ERC20(tokenPool).totalSupply()` to get the total invested value in token?

contracts/defi/provider/UniswapProvider.sol
```diff
@@ -0,0 +1,155 @@
...
+    )
+        internal 
+        returns (uint256)
+    {
+        address tokenPool = UniswapFactory(_factory).getExchange(_token);
+        require(tokenPool != address(0), "Uniswap: target token is not traded on Uniswap");
+
+        uint256 tokenBalance = ERC20(_token).balanceOf(address(_wallet));
+        if(_amount > tokenBalance) {
+            uint256 ethToSwap = UniswapExchange(tokenPool).getEthToTokenOutputPrice(tokenBalance - _amount);
```
olivdb: `tokenBalance - _amount` -> `_amount - tokenBalance`

contracts/defi/provider/UniswapProvider.sol
```diff
@@ -0,0 +1,155 @@
...
+
+/**
+ * @title Uniswap
+ * @dev Wrapper contract to integrate Uniswap.
+ * The first item of the oracles array is the Uniswap Factory contract.
+ * @author Julien Niset - <julien@argent.xyz>
+ */
+contract UniswapProvider is Invest {
+
+    address constant internal ETH_TOKEN_ADDRESS = 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE;
```
olivdb: Not used

contracts/defi/provider/CompoundV2Provider.sol
```diff
@@ -0,0 +1,405 @@
...
+    {
+        require(_oracles.length == 2, "CompoundV2: invalid oracles length");
+        address cToken = CompoundRegistry(_oracles[1]).getCToken(_token);
+        mint(_wallet, cToken, _token, _amount);
+        _invested = _amount;
+    }
+
+    /**
+     * @dev Exit invested postions.
+     * @param _wallet The target wallet.s
```
olivdb: .s

contracts/defi/provider/CompoundV2Provider.sol
```diff
@@ -0,0 +1,405 @@
...
+     */
+    function mint(BaseWallet _wallet, address _cToken, address _token, uint256 _amount) internal {
+        require(_cToken != address(0), "Compound: No market for target token");
+        require(_amount > 0, "Compound: amount cannot be 0");
+        if(_token == ETH_TOKEN_ADDRESS) {
+            _wallet.invoke(_cToken, _amount, abi.encodeWithSignature("mint()"));
+        }
+        else {
+            _wallet.invoke(_token, 0, abi.encodeWithSignature("approve(address,uint256)", _cToken, _amount));
+            _wallet.invoke(_cToken, 0, abi.encodeWithSignature("mint(uint256)", _amount));
```
olivdb: Shall we maybe check the return value and revert on error?

contracts/defi/provider/CompoundV2Provider.sol
```diff
@@ -0,0 +1,405 @@
...
+    /**
+     * @dev Redeems underlying tokens from a cToken contract.
+     * @param _wallet The target wallet.
+     * @param _cToken The cToken contract.
+     * @param _amount The amount of cToken to redeem.
+     */
+    function redeem(BaseWallet _wallet, address _cToken, uint256 _amount) internal {     
+        require(_cToken != address(0), "Compound: No market for target token");   
+        require(_amount > 0, "Compound: amount cannot be 0");
+        _wallet.invoke(_cToken, 0, abi.encodeWithSignature("redeem(uint256)", _amount));
```
olivdb: Shall we maybe check the return value and revert on error?

contracts/defi/provider/CompoundV2Provider.sol
```diff
@@ -0,0 +1,405 @@
...
+    /* ********************************** Implementation of Loan ************************************* */
+
+    /**
+     * @dev Opens a collateralized loan.
+     * @param _wallet The target wallet.
+     * @param _collateral The token used as a collateral.
+     * @param _collateralAmount The amount of collateral token provided.
+     * @param _debtToken The token borrowed.
+     * @param _debtAmount The amount of tokens borrowed.
+     * @param _oracles (optional) The address of one or more oracles contracts that may be used by the provider to query information on-chain.
```
olivdb: I would specify here that `_oracles[0]` should be the Comptroller and `_oracles[1]` should be the Compound Registry

contracts/defi/provider/CompoundV2Provider.sol
```diff
@@ -0,0 +1,405 @@
...
+     * @dev Repays some borrowed underlying tokens to a cToken contract.
+     * @param _wallet The target wallet.
+     * @param _cToken The cToken contract.
+     * @param _amount The amount of underlying to repay.
+     */
+    function repayBorrow(BaseWallet _wallet, address _cToken, uint256 _amount) internal {
+        require(_cToken != address(0), "Compound: No market for target token");
+        require(_amount > 0, "Compound: amount cannot be 0");
+        string memory symbol = CToken(_cToken).symbol();
+        if(keccak256(abi.encodePacked(symbol)) == keccak256(abi.encodePacked("cETH"))) {
```
olivdb: Maybe it's simpler (and more consistent with what we do in `mint()`) to do:
```
address token = CToken(_cToken).underlying()
if(token == ETH_TOKEN_ADDRESS) { ... }
```

olivdb: Ah sorry, I just realised cETH does not implement `underlying()`, so ignore my comment

contracts/defi/provider/CompoundV2Provider.sol
```diff
@@ -0,0 +1,405 @@
...
+        markets[0] = CompoundRegistry(_oracles[1]).getCToken(_collateral);
+        markets[1] = CompoundRegistry(_oracles[1]).getCToken(_debtToken);
+        address comptroller = _oracles[0];
+        _wallet.invoke(comptroller, 0, abi.encodeWithSignature("enterMarkets(address[])", markets));
+        mint(_wallet, markets[0], _collateral, _collateralAmount);
+        borrow(_wallet, markets[1], _debtAmount);
+    }
+
+    /**
+     * @dev Closes a collateralized loan by repaying all debts (plus interest). Note that it does not redeem the collateral.
```
olivdb: Probably worth clarifying that this closes *all* Compound loans, for all markets

contracts/defi/provider/CompoundV2Provider.sol
```diff
@@ -0,0 +1,405 @@
...
+        address comptroller = _oracles[0];
+        address[] memory markets = Comptroller(comptroller).getAssetsIn(address(_wallet));
+        for(uint i = 0; i < markets.length; i++) {
+            address cToken = markets[i];
+            uint debt = CToken(cToken).borrowBalanceCurrent(address(_wallet));
+            if(debt > 0) { 
+                repayBorrow(_wallet, cToken, debt);
+                uint collateral = CToken(cToken).balanceOf(address(_wallet));
+                if(collateral == 0) { 
+                    _wallet.invoke(comptroller, 0, abi.encodeWithSignature("exitMarket(address)", address(cToken)));
```
olivdb: I understand that it does not hurt to come back to the initial state of not being in the market but is there any other benefit to exiting the market when there is no collateral left in it?

contracts/defi/provider/MakerProvider.sol
```diff
@@ -0,0 +1,818 @@
...
+    bytes4 constant internal ETH_TOKEN_SWAP_INPUT = bytes4(keccak256("ethToTokenSwapInput(uint256,uint256)"));
+    bytes4 constant internal TOKEN_ETH_SWAP_INPUT = bytes4(keccak256("tokenToEthSwapInput(uint256,uint256,uint256)"));
+
+    using SafeMath for uint256;
+
+    /* *************** Events *************************** */
+
+    event CdpOpened(address indexed wallet, bytes32 cup, uint256 pethCollateral, uint256 daiDebt);    
+    event CdpUpdated(address indexed wallet, bytes32 cup, uint256 pethCollateral, uint256 daiDebt);    
+    event CdpClosed(address indexed wallet, bytes32 cup);
```
olivdb: I think we can remove those events now that we have `LoanOpened`, `LoanClosed`, `CollateralAdded`, etc in `LoanManager`. I confirmed with Matt that the client does not need them.

contracts/defi/provider/CompoundV2Provider.sol
```diff
@@ -0,0 +1,405 @@
...
+
+    /**
+     * @dev Exits a cToken market if there is no more collateral and debt.
+     * @param _wallet The target wallet.
+     * @param _cToken The cToken contract.
+     * @param _comptroller The comptroller contract.
+     */
+    function exitMarketIfNeeded(BaseWallet _wallet, address _cToken, address _comptroller) internal {
+        uint collateral = CToken(_cToken).balanceOf(address(_wallet));
+        uint debt = CToken(_cToken).borrowBalanceStored(address(_wallet));
```
olivdb: I think we should use `borrowBalanceCurrent(address(_wallet))` here instead.

contracts/defi/provider/CompoundV2Provider.sol
```diff
@@ -0,0 +1,405 @@
...
+
+    /**
+     * @dev Enters a cToken market if it was not entered before.
+     * @param _wallet The target wallet.
+     * @param _cToken The cToken contract.
+     * @param _comptroller The comptroller contract.
+     */
+    function enterMarketIfNeeded(BaseWallet _wallet, address _cToken, address _comptroller) internal {
+        bool isEntered = Comptroller(_comptroller).checkMembership(address(_wallet), CToken(_cToken));
+        if(!isEntered) {
```
olivdb: Not sure this check is necessary as `enterMarkets(address[])` just skips markets that have already been entered.

contracts/defi/provider/CompoundV2Provider.sol
```diff
@@ -0,0 +1,405 @@
...
+        address comptroller = _oracles[0];
+        (uint error, uint liquidity, uint shortfall) = Comptroller(comptroller).getAccountLiquidity(address(_wallet));
+        require(error == 0, "Compound: failed to get account liquidity");
+        if(liquidity > 0) {
+            return (1, liquidity);
+        }
+        if(shortfall > 0) {
+            return (2, shortfall);
+        }
+        return (0,0);
```
olivdb: I think we are missing the (rare) case where the loan exists but both `liquidity` and `shortfall` are `0`, in which case we should return `(1, 0)`.

contracts/defi/provider/MakerProvider.sol
```diff
@@ -0,0 +1,818 @@
...
+     * @dev Draw more DAI from the CDP and exchange it to collateral.
+     * @param _wallet The target wallet
+     * @param _leverageId The id of the CDP.
+     * @param _availableCollateral The amount of collateral available in the CDP.
+     * @param _daiPerPethRatio The ratio of DAI that can be drawn from 1 PETH.
+     * @param _makerCdp The Maker CDP contract
+     * @param _uniswapFactory The Uniswap Factory contract used for the exchange.
+     * @return the amount of DAI that was drawn from the CDP and the amount of collateral that was added.
+     */
+    function drawMoreDai(
```
olivdb: Let's comment this out as well for now?

contracts/defi/provider/MakerProvider.sol
```diff
@@ -0,0 +1,818 @@
...
+        _daiPerMKR = uint256(daiPerMKR_);
+    }
+
+    /**
+     * @dev Gives the additional amount of DAI that can be drawn from a CDP, given an additional amount of PETH collateral
+     * @param _conversionRatio The conversion ratio to use (must be greater than 1.5)
+     * @param _makerCdp The Maker CDP contract
+     * @return The amount of DAI that can be drawn from the CDP per unit of PETH
+     */
+    function availableDaiPerPeth(uint256 _conversionRatio, IMakerCdp _makerCdp) internal returns (uint256 _availableDaiPerPeth) {
```
olivdb: Let's comment this out as well for now?

contracts/defi/provider/MakerProvider.sol
```diff
@@ -0,0 +1,818 @@
...
+     * @dev Converts a given amount of ETH collateral into DAI and MKR in proportion 
+     * to their requirements as debt and fee repayments
+     * @param _wallet The target wallet
+     * @param _cup The id of the CDP.
+     * @param _collateralAmount The amount of ETH collateral to convert
+     * @param _tab The total amount of DAI debt in the CDP
+     * @param _makerCdp The Maker CDP contract
+     * @return the amount of converted DAI
+     */
+    function convertEthCollateralToDaiAndMkr(
```
olivdb: Let's comment this out as well for now?

contracts/modules/LoanManager.sol
juniset: Done.

contracts/defi/provider/MakerProvider.sol
juniset: agreed and done

contracts/defi/provider/MakerProvider.sol
juniset: Done.

contracts/defi/Loan.sol
juniset: Done

contracts/defi/provider/CompoundV2Provider.sol
juniset: Done.

contracts/defi/provider/MakerProvider.sol
juniset: Agreed. Note that the module always provides the same oracle array for a provider (in the case of Maker it is a length 2), so I've added the condition `require(_oracles.length >= 1` for methods that only use the first entry and `require(_oracles.length == 2` for methods that use both.

contracts/defi/provider/MakerProvider.sol
juniset: same as above

contracts/defi/provider/MakerProvider.sol
juniset: same as above

contracts/defi/provider/MakerProvider.sol
juniset: same as above

contracts/defi/provider/MakerProvider.sol
juniset: same as above

contracts/defi/provider/UniswapProvider.sol
juniset: Agreed and done.

contracts/defi/provider/UniswapProvider.sol
juniset: Changed.

contracts/defi/provider/UniswapProvider.sol
juniset: Removed.

contracts/defi/provider/CompoundV2Provider.sol
juniset: Done.

contracts/defi/provider/MakerProvider.sol
juniset: Agreed.

contracts/defi/provider/MakerProvider.sol
juniset: Agreed.

contracts/defi/provider/MakerProvider.sol
juniset: Agreed.

contracts/defi/provider/MakerProvider.sol
juniset: Agreed and done.

contracts/defi/provider/CompoundV2Provider.sol
juniset: For the moment the wallets do not return anything when invoked (it was only introduced in solidity 5.0). I'm only adding a return in this PR to be exploited in the future but we will need to be careful as there will be 2 types of wallets. I don't want to introduce this complexity now.

So for now we need to ignore the return of the `invoke` method.

contracts/defi/provider/CompoundV2Provider.sol
juniset: Same as above.

contracts/defi/provider/CompoundV2Provider.sol
juniset: Done.

contracts/defi/provider/CompoundV2Provider.sol
juniset: Agreed and done.

contracts/defi/provider/CompoundV2Provider.sol
juniset: Yes, because the Comptroller checks the status of each entered market when the loan is modified (collateral/debt increased/decreased) which costs gas. So exiting an empty market reduces the gas of every operation.

contracts/defi/provider/CompoundV2Provider.sol
juniset: In principle yes, but `borrowBalanceCurrent` just accrues all interests first then calls `borrowBalanceStored` which costs a lot of gas. In our case `exitMarketIfNeeded` is always called after an action that already accrues interests so both methods return the same value while `borrowBalanceStored` is much cheaper in terms of gas.

contracts/defi/provider/CompoundV2Provider.sol
juniset: True but it removes an unnecessary `invoke` to the wallet which costs more then a simple read.

contracts/defi/provider/CompoundV2Provider.sol
juniset: Yes probably, but this is a limitation from Compound. If there is no loan (i.e. no entered markets) the Comptroller will return (0,0,0) and we have no way to discriminate between 'no loan' and 'a loan but no liquidity and no shortfall'. I believe the first scenario is much more likely in our case hence we must miss the edge case you describe.

PR #23, opened 2019-07-29
Improving Upgrade Mechanism

olivdb:
…without user consent + fixing upgrade scripts


contracts/upgrade/ModuleRegistry.sol
```diff
@@ -46,27 +42,6 @@ contract ModuleRegistry is Owned {
         emit ModuleDeRegistered(_module);
     }
 
-        /**
```
juniset: I think we should not modify the ModuleRegistry contract at all as the ` UpgraderRegistered` event is used by the client to load the list of possible upgraders from the blockchain. Also, all current modules point to the actual ModuleRegistry so updating it will make the upgrades more complicated.

deployment/7_upgrade_v1_1.js
```diff
@@ -109,16 +116,22 @@ const deploy = async (network) => {
...
-            toRemove.map((module) => {return module.address;}),
-            toAdd.map((module) => {return module.address;})
+            config.contracts.ModuleRegistry,
+            toRemove.map(module => module.address),
+            toAdd.map(module => module.address)
         );
-        const upgraderName = version.fingerprint + '_' + fingerprint; 
-        await multisigExecutor.executeCall(ModuleRegistryWrapper, "registerUpgrader", [UpgraderWrapper.contractAddress, utils.asciiToBytes32(upgraderName)]);
+        const upgraderName = version.fingerprint + '_' + fingerprint;
+        await multisigExecutor.executeCall(ModuleRegistryWrapper, "registerModule", [UpgraderWrapper.contractAddress, utils.asciiToBytes32(upgraderName)]);
```
juniset: Following my earlier comment about keeping the current ModuleRegistry, we need to make sure that adding the upgrader triggers the `UpgraderRegistered` event so we should register the upgrader as module and as an upgrader.

deployment/8_upgrade_v1_2.js
```diff
@@ -127,45 +127,51 @@ const deploy = async (network) => {
...
         const UpgraderWrapper = await deployer.deploy(
             Upgrader,
             {},
-            toRemove.map((module) => {return module.address;}),
-            toAdd.map((module) => {return module.address;})
+            config.contracts.ModuleRegistry,
+            toRemove.map(module => module.address),
+            toAdd.map(module => module.address)
         );
-        const upgraderName = version.fingerprint + '_' + fingerprint;
```
juniset: Same as above. We should register the upgrader module as a module and a register.

PR #24, opened 2019-07-31
Improving TokenPriceProvider

olivdb:


contracts/exchange/TokenPriceProvider.sol
```diff
@@ -14,19 +15,17 @@ contract TokenPriceProvider {
 
     mapping(address => uint256) public cachedPrices;
```
juniset: We should pass the address of Kyber in the constructor.

contracts/exchange/TokenPriceProvider.sol
```diff
@@ -14,19 +15,17 @@ contract TokenPriceProvider {
...
 
-    function syncPrice(ERC20 token) public {
-        uint256 expectedRate;
-        (expectedRate,) = kyberNetwork().getExpectedRate(token, ERC20(ETH_TOKEN_ADDRESS), 10000);
-        cachedPrices[address(token)] = expectedRate;
+    function setPrice(ERC20 _token, uint256 _price) external onlyManager {
+        cachedPrices[address(_token)] = _price;
     }
```
juniset: We should add a method to set a list of token prices in one call.

contracts/exchange/TokenPriceProvider.sol
```diff
@@ -14,19 +15,17 @@ contract TokenPriceProvider {
...
 
-    //
-    // Convenience functions
-    //
+    function syncPrice(ERC20 _token) external onlyManager {
```
juniset: Why restrict this function to managers? It's ok if external parties synch the price of Kyber tokens so we should remove the `onlyManager` modifier.

contracts/exchange/TokenPriceProvider.sol
```diff
@@ -14,19 +15,17 @@ contract TokenPriceProvider {
...
+        _syncPrice(_token);
+    }
 
-    function syncPriceForTokenList(ERC20[] memory tokens) public {
-        for(uint16 i = 0; i < tokens.length; i++) {
-            syncPrice(tokens[i]);
+    function syncPriceForTokenList(ERC20[] calldata _tokens) external onlyManager {
```
juniset: Same comment as above. We should remove the `onlyManager` modifier.

contracts/exchange/TokenPriceProvider.sol
olivdb: What would be the benefits of doing that? I feel that if we open price changes to any address we increase the attack surface of this contract for no good reason. We would also need to make sure in `_syncPrice()` to reject changes of prices that try to set a price to zero (otherwise prices of tokens not on Kyber can be set to 0 using `syncPrice`). That would prevent us from handling tokens whose value is zero (probably never going to happen anyway but who knows). This also makes the contract less predictable: we set a price to 10 and shortly after someone execute `syncPrice` to set it to 10.9. I had left `syncPrice` here just to facilitate the transition from the previous TokenPriceProvider but I'm now thinking that it would make more sense to remove all Kyber-syncing methods.

juniset: At first I did not see the point of keeping the possibility to synch with Kyber, but then I realized it could be a way to make the system more trustless for our users, i.e. if a user does not trust the prices that we push or if we are corrupted he can make the oracle synch with Kyber.

But I am equally fine with removing it. So I think we should either remove this method or make it public.

contracts/exchange/TokenPriceProvider.sol
```diff
@@ -36,17 +20,39 @@ contract TokenPriceProvider {
...
+    }
+
+    function syncPrice(ERC20 token) public {
+        require(address(kyberNetwork) != address(0), "Kyber sync is disabled");
+        uint256 expectedRate;
+        (expectedRate,) = kyberNetwork.getExpectedRate(token, ERC20(ETH_TOKEN_ADDRESS), 10000);
+        cachedPrices[address(token)] = expectedRate;
+    }
+
+    function syncPriceForTokenList(ERC20[] memory tokens) public {
```
juniset: Can we have the logic of `syncPrice` in `syncPriceForTokenList` and add the require check in `syncPriceForTokenList`? I know it duplicates the code but it will save some gas by avoiding checking the value of `kyberNetwork` for every token.

contracts/exchange/TokenPriceProvider.sol
```diff
@@ -36,17 +20,39 @@ contract TokenPriceProvider {
...
+    // Address of the KyberNetwork contract
+    KyberNetwork public kyberNetwork;
+
+    constructor(KyberNetwork _kyberNetwork) public {
+        kyberNetwork = _kyberNetwork;
+    }
+
+    function setKyberNetwork(KyberNetwork _kyberNetwork) external onlyManager {
+        kyberNetwork = _kyberNetwork;
+    }
```
juniset: Can we have this at the beginning of the contract? I know it's for backward compatibility with the backend but having the constructor and the storage in the middle makes the contract unclean.

contracts/exchange/TokenPriceProvider.sol
olivdb: Done. Sorry, I had missed that comment.

contracts/exchange/TokenPriceProvider.sol
olivdb: Done. But maybe `syncPrice()` and `setPrice()` could both be `internal` ? @petar-argent do you ever call `syncPrice()` in addition to `syncPriceForTokenList()`?

contracts/exchange/TokenPriceProvider.sol
olivdb: Done

PR #30, opened 2019-09-02
New BaseWallet returning call result and revert reason

olivdb:
Closes #27


contracts/modules/common/BaseModule.sol
```diff
@@ -83,4 +83,26 @@ contract BaseModule is Module {
...
+     * @param _wallet The target wallet.
+     * @param _to The target address for the transaction.
+     * @param _value The value of the transaction.
+     * @param _data The data of the transaction.
+     */
+    function invokeWallet(address _wallet, address _to, uint256 _value, bytes memory _data) internal returns (bytes memory _res) {
+        bool success;
+        // solium-disable-next-line security/no-call-value
+        (success, _res) = _wallet.call.value(0)(
+            abi.encodeWithSignature("invoke(address,uint256,bytes)", _to, _value, _data)
```
juniset: Maybe it is worth defining
`bytes4 constant internal INVOKE = bytes4(keccak256("invoke(address,uint256,bytes)"));`
and use `abi.encodeWithSelector(INVOKE, _to, _value, _data)` to save some gas.

contracts/wallet/BaseWallet.sol
```diff
@@ -101,16 +101,23 @@ contract BaseWallet {
...
+        bool success;
         // solium-disable-next-line security/no-call-value
-        (bool success, ) = _target.call.value(_value)(_data);
-        require(success, "BW: call to target failed");
+        (success, _result) = _target.call.value(_value)(_data);
+        if(!success) {
+            // solium-disable-next-line security/no-inline-assembly
+            assembly {
+                // _result = {result_length:32}{sig:4}{pos:32}{revert_reason_length:32}{revert_reason}
+                revert(add(_result, 0x64), add(_result, 0x44))
```
juniset: Maybe we could use
```
assembly {
                returndatacopy(0, 0, returndatasize)
                revert(0, returndatasize)
            }
```
as we do in the fallback method as it is more readable.

contracts/test/basewallet/OldTestModule.sol
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    // *************** External/Public Functions ********************* //
+
+    function callDapp(OldBaseWallet _wallet)
+        external
+    {
+        _wallet.invoke(address(dapp), 0, abi.encodeWithSignature("noReturn()", 0));
+    }
+
```
juniset: I think we should also add `callDapp2` to test the new wallet interacting with a dapp that returns a value using an old module.

test/baseWallet.js
```diff
@@ -13,55 +16,94 @@ describe("Test BaseWallet", function () {
...
+            assert.equal(uintVal, 42, "should have the correct uint");
+            assert.equal(addressVal, nonowner.address, "should have the address");
+        });
+
+    })
+
+    describe("New BaseWallet", () => {
+        it("should work with old modules", async () => {
+            await wallet.init(owner.address, [oldModule.contractAddress]);
+            await oldModule.callDapp(wallet.contractAddress);
```
juniset: As mentioned above I think we should test the new wallet interacting with a dapp that returns a value using an old module.

contracts/modules/common/BaseModule.sol
```diff
@@ -83,4 +83,26 @@ contract BaseModule is Module {
...
+        );
+        if(success && _res.length > 0) { //_res is empty if _wallet is an "old" BaseWallet that can't return output values
+            (_res) = abi.decode(_res, (bytes));
+        } else if (_res.length > 0) {
+            revert(string(_res));
+        } else if(!success) {
```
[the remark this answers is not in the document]
olivdb: We need it because we need to make sure we are not in the 4th possibility (`success == true && _res.length == 0`) for which we do not want to do anything

contracts/test/basewallet/OldTestModule.sol
olivdb: Done

contracts/modules/common/BaseModule.sol
olivdb: Why do you think it should save some gas? In my tests, this makes both the module deployment cost and the method execution cost slightly higher.

contracts/wallet/BaseWallet.sol
olivdb: That would return the reason message prefixed with some junk because of the way reason messages are returned (see EIP 838). We could skip the junk using:
```
let len := sub(returndatasize, 0x44)
returndatacopy(0, 0x44, len)
revert(0, len)
```
I thought that this would be less gas-efficient than what I did (because of the additional copy) but it turns out that it's pretty much the same gas cost, so I can use that expression instead.

test/baseWallet.js
olivdb: Done

contracts/modules/common/BaseModule.sol
juniset: Really? That's surprising. I would expect that having the signature computed at compile time and stored would be cheaper than computing it at run time, but I guess this statement is making a lot of assumptions about how the compiler works.

contracts/modules/common/BaseModule.sol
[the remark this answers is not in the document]
juniset: Indeed, 2 conditions makes 4 possibilities :-)

contracts/wallet/BaseWallet.sol
```diff
@@ -101,16 +101,25 @@ contract BaseWallet {
...
-        require(success, "BW: call to target failed");
+        (success, _result) = _target.call.value(_value)(_data);
+        if(!success) {
+            // solium-disable-next-line security/no-inline-assembly
+            assembly {
+                // returndata = {sig:4}{pos:32}{revert_reason_length:32}{revert_reason} as per EIP838
+                let len := sub(returndatasize, 0x44)
+                returndatacopy(0, 0x44, len)
+                revert(0, len)
+            }
```
juniset: I still don't understand why we don't do:
```
assembly {
                returndatacopy(0, 0, returndatasize)
                revert(0, returndatasize)
            }
``` 
and return the entire byte array from the error we are bubbling up?

juniset: Sorry, I didn't see your previous response at it was outdated and hidden.

Still, isn't it safer to return exactly what the lower call error is returning and let the client deal with any potential junk?

olivdb: You're right. I got confused by an incorrect idea that every subsequent revert would recursively add some more junk... (that's not the case using `assembly{revert()}`). Changed.

contracts/modules/common/BaseModule.sol
```diff
@@ -83,4 +83,30 @@ contract BaseModule is Module {
...
+     * @dev Helper method to invoke a wallet.
+     * @param _wallet The target wallet.
+     * @param _to The target address for the transaction.
+     * @param _value The value of the transaction.
+     * @param _data The data of the transaction.
+     */
+    function invokeWallet(address _wallet, address _to, uint256 _value, bytes memory _data) internal returns (bytes memory _res) {
+        bool success;
+        // solium-disable-next-line security/no-call-value
+        (success, _res) = _wallet.call.value(0)(
```
juniset: Why do we explicitely set the value to zero?

olivdb: No good reason. Removed.

PR #31, opened 2019-09-19
Transfer manager

juniset:
Replacing TokenTransfer module by TransferManager with:
- support for EIP 1271
- addition of approveTransfer() and callContract() methods to support safe WalletConnect integration


contracts/modules/TransferManager.sol
```diff
@@ -0,0 +1,532 @@
...
+    * @param _id The pending transfer ID.
+    * @return the epoch time at which the pending transfer can be executed.
+    */
+    function getPendingTransfer(BaseWallet _wallet, bytes32 _id) external view returns (uint64 _executeAfter) {
+        _executeAfter = uint64(configs[address(_wallet)].pendingActions[_id]);
+    }
+
+    /**
+    * @dev Implementation of EIP 1271.
+    * Should return whether the signature provided is valid for the provided data.
```
olivdb: I would say "should return a magic value if the provided signature is valid or revert otherwise".

contracts/modules/TransferManager.sol
```diff
@@ -0,0 +1,532 @@
...
+    */
+    function isValidSignature(bytes memory _data, bytes memory _signature) public view returns (bytes4) {
+        bytes32 msgHash = keccak256(abi.encodePacked(_data));
+        isValidSignature(msgHash, _signature);
+        return ERC721_ISVALIDSIGNATURE_BYTES;
+    }
+
+    /**
+    * @dev Implementation of EIP 1271.
+    * Should return whether the signature provided is valid for the provided data.
```
olivdb: I would say "should return a magic value if the provided signature is valid or revert otherwise".

olivdb:
All good for me!

PR #33, opened 2019-10-16
Allowing owner to transfer ownership of their wallet

olivdb:


contracts/modules/RecoveryManager.sol
```diff
@@ -80,13 +117,18 @@ contract RecoveryManager is BaseModule, RelayerModule {
...
+        uint256 _securityPeriod,
+        uint256 _securityWindow
     )
         BaseModule(_registry, _guardianStorage, NAME)
         public
     {
+        guardianStorage = _guardianStorage;
         recoveryPeriod = _recoveryPeriod;
         lockPeriod = _lockPeriod;
+        securityPeriod = _securityPeriod;
```
juniset: I think we should add a condition for `recoveryPeriod` `securityPeriod` and `securityWindow` in the constructor as these 3 parameters must be related for the wallet to be secure. In particular they must satisfy `recoveryPeriod >= securityPeriod + securityWindow`.

contracts/modules/RecoveryManager.sol
```diff
@@ -129,33 +171,85 @@ contract RecoveryManager is BaseModule, RelayerModule {
...
-        bytes memory _signatures
+        address _newOwner
+    )
+        external
+        onlyWalletOwner(_wallet)
+        onlyWhenUnlocked(_wallet)
+        notWhenOwnershipTransfer(_wallet)
+    {
+        require(_newOwner != address(0), "RM: new owner address cannot be null");
+        require(!isOwner(_wallet, _newOwner), "RM: new owner address already owner");
```
juniset: Is this check necessary? Avoiding it will save some gas.

contracts/modules/RecoveryManager.sol
```diff
@@ -115,12 +157,12 @@ contract RecoveryManager is BaseModule, RelayerModule {
      * @param _wallet The target wallet.
      */
     function finalizeRecovery(BaseWallet _wallet) external onlyExecute onlyWhenRecovery(_wallet) {
```
juniset: We can remove the `onlyExecute` modifier. Is is part of another PR but it makes more sense to have it here.

PR #34, opened 2019-10-22
Allowing Relayed calls to OnlyOwnerModule.addModule + adding ApprovedTransfer.callContract + allowing test run with ganache-cli

olivdb:
- allowing relayed updates to be performed using OnlyOwnerModule.addModule
- adding tests for relayed update
- cleaning deploy scripts to match the upgrade pattern on prod
- fixing deploy scripts to add the TransferManager and the new ApprovedTransfer and to remove the ModuleManager, DappManager and TokenManager using the correct update scheme (legacy or new) based on the origin version and the environment.


contracts/modules/TransferManager.sol
```diff
@@ -502,25 +461,6 @@ contract TransferManager is BaseModule, RelayerModule, LimitManager {
...
-    )
-        internal
-        view
-        returns (bool)
-    {
-        address signer = recoverSigner(_signHash, _signatures, 0);
-        return isOwner(_wallet, signer); // "TT: signer must be owner"
-    }
-
     function getRequiredSignatures(BaseWallet /* _wallet */, bytes memory _data) internal view returns (uint256) {
```
juniset: We should sync with @petar-argent and use the default implementation provided by `OnlyOwnerModule`.

contracts/modules/common/BaseModule.sol
```diff
@@ -60,7 +60,7 @@ contract BaseModule is Module {
      * @param _wallet The target wallet.
      * @param _module The modules to authorise.
      */
-    function addModule(BaseWallet _wallet, Module _module) external strictOnlyWalletOwner(_wallet) {
+    function addModule(BaseWallet _wallet, Module _module) public strictOnlyWalletOwner(_wallet) {
```
juniset: We have to be careful when we change the visibility as this method should implement the `addModule` method of the interface `Module` which is declared external. With this change we are no longer implementing the `Module` interface.

juniset:
The last test of lockManager.js fails at L119. Not related to this PR but it should be fixed.

contracts/modules/TransferManager.sol
olivdb: Yes, will be updating the three contracts that need to be simplified once the backend is ready.

@petar-argent for reference the full list of methods that it'd be great for the backend to call directly instead of via execute is:

GuardianManager:
confirmGuardianAddition(address,address)
confirmGuardianRevokation(address,address)

RecoveryManager:
finalizeRecovery(address)

TransferManager:
executePendingTransfer(address,address,address,uint256,bytes,uint256)"

contracts/modules/common/BaseModule.sol
olivdb: Technically I think we are still implementing `Module` because adding a public method is like adding an external + an internal method :) But that was a mistake. Fixed.

contracts/modules/common/BaseTransfer.sol
```diff
@@ -0,0 +1,72 @@
...
+
+    // *************** Events *************************** //
+
+    event Transfer(address indexed wallet, address indexed token, uint256 indexed amount, address to, bytes data);
+    event Approved(address indexed wallet, address indexed token, uint256 amount, address spender);
+    event CalledContract(address indexed wallet, address indexed to, uint256 amount, bytes data);
+
+    // *************** Constructor ********************** //
+
+    constructor(ModuleRegistry _registry, bytes32 _name) BaseModule(_registry, _name) public {}
```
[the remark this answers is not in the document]
olivdb: Done

contracts/modules/common/OnlyOwnerModule.sol
```diff
@@ -11,6 +11,28 @@ import "../../wallet/BaseWallet.sol";
...
+        return this.isOnlyOwnerModule.selector;
+    }
+
+    /**
+     * @dev Adds a module to a wallet. First checks that the module is registered.
+     * Unlike its overrided parent, this method can be called via the RelayerModule's execute()
+     * @param _wallet The target wallet.
+     * @param _module The modules to authorise.
+     */
+    function addModule(BaseWallet _wallet, Module _module) public onlyWalletOwner(_wallet) {
```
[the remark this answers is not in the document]
olivdb: Done

contracts/modules/TransferManager.sol
olivdb: Adding @dappcoder to this thread

PR #43, opened 2019-11-19
Fix/transfer refund

juniset:


contracts/modules/TransferManager.sol
```diff
@@ -108,35 +103,36 @@ contract TransferManager is BaseModule, RelayerModule, OnlyOwnerModule, BaseTran
...
         // migrate daily pending if we are within a rolling period
         (uint256 unspent, uint64 periodEnd) = oldLimitManager.getDailyUnspent(_wallet);
         if(periodEnd > now) {
-            limits[address(_wallet)].dailySpent = DailySpent(uint128(currentLimit.sub(unspent)), periodEnd);
+            limits[address(_wallet)].dailySpent = DailySpent(uint128(current.sub(unspent)), periodEnd);
         }
+
+        // setup static calls
+        _wallet.enableStaticCall(address(this), ERC721_ISVALIDSIGNATURE_BYTES);
+        _wallet.enableStaticCall(address(this), ERC721_ISVALIDSIGNATURE_BYTES32);
```
olivdb: Why moving this down here? These lines won't be executed for new wallets or new deployments.

test/recoveryManager.js
```diff
@@ -65,7 +65,7 @@ describe("RecoveryManager", function () {
     function testExecuteRecovery(guardians) {
         it("should let a majority of guardians execute the recovery procedure (relayed transaction)", async () => {
             let majority = guardians.slice(0, Math.ceil((guardians.length + 1) / 2));
-            await manager.relay(recoveryManager, 'executeRecovery', [wallet.contractAddress, newowner.address], wallet, sortWalletByAddress(majority));
+            let txReceipt = await manager.relay(recoveryManager, 'executeRecovery', [wallet.contractAddress, newowner.address], wallet, sortWalletByAddress(majority));
```
olivdb: `txReceipt` is not used

contracts/modules/TransferManager.sol
juniset: Yes, I have revereted.

PR #47, opened 2019-11-29
Clean deployment scripts

juniset:


deployment/1_setup_test_environment.js
```diff
@@ -65,7 +66,13 @@ const deploy = async (network, secret) => {
         // Deploy Kyber Network if needed
         const address = await deployKyber(deployer);
         configurator.updateKyberContract(address);
-    }
+	}
+	
+	if (config.defi.maker.deployOwn) {
+        // Deploy Kyber Network if needed
```
olivdb: Comment should be "Deploy Maker's mock Migration contract if needed"

juniset: done

PR #50, opened 2019-12-17
MakerV2 Module

olivdb:
Enabling: 
- Borrowing DAI via MCD Vaults 
- DAI investments into the DSR pot

The `MakerV2Manager` module is replacing the `MakerManager` module.


contracts/modules/maker/MakerV2Invest.sol
```diff
@@ -0,0 +1,178 @@
...
+        grantVatAccess(_wallet, address(daiJoin));
+        // Check the actual balance of DAI in the vat after the pot exit
+        uint bal = vat.dai(address(_wallet));
+        // It is necessary to check if due to rounding the exact amount can be exited by the adapter.
+        // Otherwise it will do the maximum DAI balance in the vat
+        uint256 amount = pot.chi().mul(pie) / RAY;
+        uint256 withdrawn = bal >= amount.mul(RAY) ? amount : bal / RAY;
+        invokeWallet(address(_wallet), address(daiJoin), 0, abi.encodeWithSelector(ADAPTER_EXIT, address(_wallet), withdrawn));
+    }
+}
```
juniset: You have removed the `exitAllDsr` which was present in the previous version of the module. Don't the clients need it to remove all collateral?

contracts/modules/maker/MakerV2Loan.sol
```diff
@@ -0,0 +1,626 @@
...
+    // The address of the Jug
+    JugLike internal jug;
+    // The address of the CDP Manager
+    ManagerLike internal cdpManager;
+    // The address of the SCD Tub
+    SaiTubLike internal tub;
+    // The Maker Registry in which all supported collateral tokens and their adapters are stored
+    MakerRegistry internal makerRegistry;
+    // The Uniswap Exchange contract for DAI
+    IUniswapExchange internal daiUniswap;
```
juniset: We know that Uniswap is working on a V2 that should be released around february or march. We need to anticipate that change.

contracts/modules/maker/MakerV2Invest.sol
olivdb: They will be able to use `exitDsr(_wallet, 10000)`, which is equivalent. 

Note that, because on the current `prod` deployment they need to call `exitAllDsr()` (`exitDsr(_wallet, 10000)` would leave some dust in the pot), I've also temporarily deployed a version of MakerV2 on `test` with an `exitAllDsr()` function defined as:
```
    function exitAllDsr(
        BaseWallet _wallet
    )
        external
        onlyWalletOwner(_wallet)
        onlyWhenUnlocked(_wallet)
    {
        exitDsr(_wallet, 10000);
    }
```
The idea is to change the clients from calling `exitAllDsr()` to calling `exitDsr(10000)` after we are ready to the deploy CDP-powered `MakerV2Manager` (in which `exitDsr(10000)` no longer leaves dust in the pot).

contracts/modules/common/BaseModule.sol
```diff
@@ -98,14 +98,30 @@ contract BaseModule is Module {
...
+     }
+
+    /**
+     * @dev Verify that the caller is the module or the wallet owner.
+     * @param _wallet The target wallet.
+     */
+    function verifyWalletOwner(BaseWallet _wallet) internal view {
+        require(msg.sender == address(this) || isOwner(_wallet, msg.sender), "BM: must be wallet owner");
+    }
+
```
juniset: Why did you move the `verifyUnlocked ` and `verifyWalletOwner ` logic to internal methods? They seem to be used only once. It is more gas efficient?

contracts/modules/maker/MakerV2Loan.sol
juniset: For the record we have decided to leave it as it is. Users will be able to use the atomic swap of ETH/MKR for CDP migration and ETH/DAI to repay their debt for as long as there will be liquidity on Uniswap. We anticipate that:
1. all our CDP users will have migrated their CDP by the time the ETH/MKR liquidity is too low on Uniswap.
2. the client will handle the ETH/DAI swap before repaying debt when the ETH/DAI Uniswap pool is empty.

contracts/modules/maker/MakerV2Loan.sol
```diff
@@ -0,0 +1,626 @@
...
+interface IUniswapExchange {
+    function getEthToTokenOutputPrice(uint256 _tokens_bought) external view returns (uint256);
+    function getEthToTokenInputPrice(uint256 _eth_sold) external view returns (uint256);
+    function getTokenToEthOutputPrice(uint256 _eth_bought) external view returns (uint256);
+    function getTokenToEthInputPrice(uint256 _tokens_sold) external view returns (uint256);
+}
+
+/**
+ * @title MakerV2Loan
+ * @dev Module to migrate old CDPs and open and manage new CDPs.
```
juniset: Could you had some explanation/context on why we decided to have the Maker module act as a proxy for user Vaults instead of letting them be the owner directly.

contracts/modules/maker/MakerV2Loan.sol
```diff
@@ -0,0 +1,626 @@
...
+        saveLoanOwner(_wallet, _loanId);
+    }
+
+    /**
+     * @dev Lets a SCD CDP owner migrate their CDP to use the new MCD engine.
+     * Requires MKR or ETH to pay the SCD governance fee
+     * @param _wallet The target wallet.
+     * @param _cup id of the old SCD CDP to migrate
+     */
+    function migrateCdp(
```
juniset: I think it would make the code easier to read if we use the word `Vault` everywhere for the new CDPs as it is the official name introduced by Maker, e.g. `openVault` `giveVault` etc. Would make it easier to differentiate between V1 and V2 positions (they are both CDP in the code now).

contracts/modules/maker/MakerV2Invest.sol
```diff
@@ -0,0 +1,178 @@
...
+    /**
+     * @dev Exit invested postions.
+     * @param _wallet The target wallet.
+     * @param _token The token address.
+     * @param _fraction The fraction of invested tokens to exit in per 10000.
+     */
+    function removeInvestment(
+        BaseWallet _wallet,
+        address _token,
+        uint256 _fraction
```
juniset: Are we sure we want to stick with `fraction` instead of `amount`? Given our current strategy there is no need for a generic interface. We may want to check with the clients if it would solve problems or if they are fine as it is.

contracts/modules/common/BaseModule.sol
olivdb: Yes, it saves quite a bit of gas on deployment. Modifiers are inlined in the code so calling an internal method within them reduces the amount of code duplicated in each function that uses the modifier.

contracts/modules/maker/MakerV2Loan.sol
olivdb: Yes, will do the change but it's worth noting that Maker keeps referring to their vaults as "CDPs" in their codebase (e.g. see their CdpManager contract).

contracts/modules/maker/MakerV2Loan.sol
olivdb: Done

contracts/modules/maker/MakerV2Loan.sol
olivdb: Done. Though I still kept the name 'CdpManager' to be consistent with Maker's naming.

contracts/modules/maker/MakerV2Invest.sol
olivdb: Being discussed with @DarthMike and @sgc-code. Note that if we use `_amount`, we'll have to also add a `removeAllInvestment()` external method.

DarthMike: Happy to have `removeInvestment` use exact `amount` (and event to return an amount as well). Currently users have issue with Compound and DSR. 

My concern was not changing Compound API as it should be the same as well, and we would fix all rounding issues we've seen.

.circleci/config.yml
```diff
@@ -41,12 +41,12 @@ jobs:
...
           command: npm run ganache >/dev/null 2>&1 & npm run test
-      - run:
-          name: "Running coverage"
-          command: npm run test:coverage
       - run:
           name: "Testing deployment scripts"
           command: npm run ganache >/dev/null 2>&1 & npm run test:deployment
+      - run:
+          name: "Running coverage"
+          command: npm run test:coverage
```
elenadimitrova: Why has the order here changed?

utils/test-manager.js
```diff
@@ -65,40 +74,58 @@ class TestManager {
...
     const block = await this.provider.getBlockNumber();
-    const timestamp = (new Date()).getTime();
-    return `0x${ethers.utils.hexZeroPad(ethers.utils.hexlify(block), 16)
-      .slice(2)}${ethers.utils.hexZeroPad(ethers.utils.hexlify(timestamp), 16).slice(2)}`;
+    const timestamp = new Date().getTime();
+    return `0x${ethers.utils
+      .hexZeroPad(ethers.utils.hexlify(block), 16)
+      .slice(2)}${ethers.utils
+      .hexZeroPad(ethers.utils.hexlify(timestamp), 16)
+      .slice(2)}`;
```
elenadimitrova: There's a lot of reformatting that has gone in this file, maybe to do with prettier settings in editor or something like that. Please reformat everything back.

contracts/modules/maker/MakerV2Loan.sol
```diff
@@ -0,0 +1,642 @@
...
+    address constant internal ETH_TOKEN_ADDRESS = 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE;
+
+    // Method signatures to reduce gas cost at depoyment
+    bytes4 constant internal ERC20_TRANSFER = bytes4(keccak256("transfer(address,uint256)"));
+    bytes4 constant internal WETH_DEPOSIT = bytes4(keccak256("deposit()"));
+    bytes4 constant internal WETH_WITHDRAW = bytes4(keccak256("withdraw(uint256)"));
+    bytes4 constant internal VAT_HOPE = bytes4(keccak256("hope(address)"));
+    bytes4 constant internal CDP_GIVE = bytes4(keccak256("give(uint256,address)"));
+    bytes4 constant internal TUB_GIVE = bytes4(keccak256("give(bytes32,address)"));
+    bytes4 constant internal ETH_TOKEN_SWAP_OUTPUT = bytes4(keccak256("ethToTokenSwapOutput(uint256,uint256)"));
```
elenadimitrova: These are used in `abi.encodeWithSelector(bytes4(keccak256(bytes(signature))), ...)` which is faster to do with `abi.encodeWithSignature(string memory signature, ...)`

https://solidity.readthedocs.io/en/latest/units-and-global-variables.html#abi-encoding-and-decoding-functions

contracts/modules/maker/MakerV2Base.sol
```diff
@@ -0,0 +1,100 @@
+pragma solidity ^0.5.4;
```
elenadimitrova: Is this missing a `Copyright (C) 2018  Argent Labs` at the top?

utils/test-manager.js
```diff
@@ -48,7 +52,12 @@ class TestManager {
...
         chainId: 3,
       };
-      return new etherlime.InfuraPrivateKeyDeployer(this.accounts[0].signer.privateKey, "ropsten", APIKEY, defaultConfigs);
+      return new etherlime.InfuraPrivateKeyDeployer(
+        this.accounts[0].signer.privateKey,
+        "ropsten",
+        APIKEY,
+        defaultConfigs,
+      );
     }
```
juniset: I had problems running the tests on Kovan (all tests with relayed transactions failing). I tracked the problem to the provider not being properly configured.

It is fixed by creating an InfuraProvider when the network is Kovan:
```
    else if (this.network === "kovan") {
      const defaultConfigs = {
        gasPrice: 20000000000,
        gasLimit: 4700000,
        chainId: 42,
      };
      return new etherlime.InfuraPrivateKeyDeployer(
        this.accounts[0].signer.privateKey,
        "kovan",
        process.env.KOVAN_INFURA_KEY,
        defaultConfigs,
      );
    }
```

contracts/modules/maker/MakerV2Base.sol
juniset: When we add the copyright we should set the year to 2019.

utils/test-manager.js
olivdb: Changed. I decided to have the kovan test scripts pass the correct Infura-based deployer to the constructor of `TestManager` to avoid creating it twice.

contracts/modules/maker/MakerV2Invest.sol
```diff
@@ -0,0 +1,180 @@
...
+    * @param _wallet The target wallet.
+    * @param _amount The amount of DAI to deposit
+    */
+    function joinDsr(
+        BaseWallet _wallet,
+        uint256 _amount
+    )
+        internal
+    {
+        if (daiToken.balanceOf(address(_wallet)) < _amount) {
```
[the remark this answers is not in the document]
olivdb: I'd say it all depends on how this is handled by the clients but I'll check with the clients that this is in fact useful. It may well be the case that we only allow DSR deposits from positive DAI balances anyways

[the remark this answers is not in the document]
olivdb: Confirmed with Miguel that this is not used and probably not useful. Will remove.

contracts/modules/maker/MakerV2Invest.sol
```diff
@@ -0,0 +1,180 @@
...
+        address _token,
+        uint256 _amount,
+        uint256 _period
+    )
+        external
+        onlyWalletOwner(_wallet)
+        onlyWhenUnlocked(_wallet)
+        returns (uint256 _invested)
+    {
+        require(_token == address(daiToken), "DM: token should be DAI");
```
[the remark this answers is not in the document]
olivdb: Done

contracts/modules/maker/MakerV2Invest.sol
```diff
@@ -0,0 +1,180 @@
...
+
+    bytes4 constant internal ADAPTER_JOIN = bytes4(keccak256("join(address,uint256)"));
+    bytes4 constant internal ADAPTER_EXIT = bytes4(keccak256("exit(address,uint256)"));
+    bytes4 constant internal VAT_HOPE = bytes4(keccak256("hope(address)"));
+    bytes4 constant internal POT_JOIN = bytes4(keccak256("join(uint256)"));
+    bytes4 constant internal POT_EXIT = bytes4(keccak256("exit(uint256)"));
+
+    // *************** Constructor ********************** //
+
+    constructor(
```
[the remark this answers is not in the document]
olivdb: Probably a vestige from pre-refactoring. Changed.

contracts/modules/common/BaseModule.sol
```diff
@@ -65,7 +64,8 @@ contract BaseModule is Module {
      * @dev Throws if the sender is not the owner of the target wallet or the module itself.
      */
     modifier onlyWalletOwner(BaseWallet _wallet) {
-        require(msg.sender == address(this) || isOwner(_wallet, msg.sender), "BM: must be an owner for the wallet");
+        // Wrapping in an internal method reduces deployment cost by avoiding duplication of inlined code
+        verifyWalletOwner(_wallet);
```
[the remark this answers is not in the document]
olivdb: Yes, I can deploy it without. FYI the gas cost difference is:
4,955,012 gas with  `verifyWalletOwner(_wallet);`
5,160,221 gas with `require(msg.sender == address(this) || isOwner(_wallet, msg.sender), "BM: must be an owner for the wallet");`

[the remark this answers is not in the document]
olivdb: Yes, I can deploy it without the internal methods. FYI the gas cost difference is:
4,954,988 gas with the internal methods
5,612,048 gas without the internal methods

contracts/modules/maker/MakerV2Base.sol
olivdb: Done

contracts/modules/maker/MakerV2Loan.sol
olivdb: We made that choice in the past because we had measured a reduction of deployment gas cost for `--runs=200` when storing a single `constant bytes4 ` to be used by multiple `abi.encodeWithSelector` instead of inlining multiple times the same string signature in `abi.encodeWithSignature`. Interestingly, I redid the comparison and I now find that though our approach was indeed cheaper for `--runs=200`, using `abi.encodeWithSignature` is actually cheaper for `--runs=999`, even when inlining the signature multiple times in the code. I'll make the change to use `abi.encodeWithSignature` everywhere.

.circleci/config.yml
olivdb: Because coverage is the slowest task to run and I find it convenient to be able to identify an issue in the deployment script earlier.

utils/test-manager.js
olivdb: Is that still the case? `eslint .` does not show me any warning or error for this file.

contracts/modules/maker/MakerV2Invest.sol
```diff
@@ -0,0 +1,180 @@
...
+    function pie(address) public view returns (uint);
+    function drip() public;
+}
+
+/**
+ * @title MakerV2Invest
+ * @dev Module to lock/unlock MCD DAI into/from Maker's Pot
+ * @author Olivier VDB - <olivier@argent.xyz>
+ */
+contract MakerV2Invest is Invest, MakerV2Base {
```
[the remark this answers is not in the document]
olivdb: Done. I was holding off on that so the clients wouldn't have to deal with two different interfaces but we now agreed with Miguel to do a phased change.

utils/test-manager.js
olivdb: Reformatted.

PR #52, opened 2020-01-22
addition of approveTokenAndCallContract to ApproveTransfer module

juniset:
Allowing to call `approveTokenAndCallContract` method with guardians. Required to allow a user to approve a token amount above its daily limit and calling a contract method that spends that amount in one atomic Tx.


contracts/modules/ApprovedTransfer.sol
```diff
@@ -77,6 +77,31 @@ contract ApprovedTransfer is BaseModule, RelayerModule, BaseTransfer {
...
+    */
+    function approveTokenAndCallContract(
+        BaseWallet _wallet,
+        address _token,
+        address _contract,
+        uint256 _amount,
+        bytes calldata _data
+    )
+        external
+        onlyWalletOwner(_wallet)
```
th3m477: onlyExecute?

PR #54, opened 2020-02-04
Support the ENS Registry migration

elenadimitrova:
Following the [ENS Registry update](https://docs.ens.domains/ens-migration/technical-description#ens-registry), here we implement updates to the Argent Registrar and the integration with the [new Registry](https://etherscan.io/address/0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e#code) 
  
For facilitating testing and maintenance, the ENS contracts which are external to the repo have been moved to a `/lib` folder to differentiate them from Argent-specific logic. These can be compiled using `npm run compile:lib` only necessary when an integrated contract changes.

We've also improved the code coverage for ENS logic, although without a code coverage report, it's difficult to say by how much exactly. TBC as part of a bigger refactoring to enable code coverage.


deployment/7_upgrade_ens.js
```diff
@@ -0,0 +1,31 @@
+const TestENSRegistry = require('../build/TestENSRegistry');
+const TestReverseRegistrar = require('../build/TestReverseRegistrar');
+const ENSResolver = require('../build/ENSResolver');
+const ENSManager = require('../build/ENSManager');
+
+const deploy = async (network) => {
+    const config = configurator.config;
```
juniset: We need to create the `configurator`, the `deployer` and the `abiUploader` first using:
```
    const manager = new DeployManager(network);
    await manager.setup();
    const configurator = manager.configurator;
    const deployer = manager.deployer;
    const abiUploader = manager.abiUploader;
```

deployment/7_upgrade_ens.js
```diff
@@ -0,0 +1,31 @@
...
+
+	// ENS domain
+	const parts = domain.split('.');
+	const extension = parts[1];
+	const domainName = parts[0];
+
+    // Instantiate the ENS Registry
+    const ensRegistry = await deployer.wrapDeployedContract(TestENSRegistry, config.ENS.ensRegistry);
+
+	// Create the 'argentx.xyz' wallet ENS namespace
```
juniset: This step is not necessary since we are using the official ENSRegistry. We need to claim ownership of `argent.xyz` for the new ENSManager using the Registrar for .xyz domains (links to DNS record owner).

deployment/7_upgrade_ens.js
```diff
@@ -0,0 +1,31 @@
...
+	const setSubnodeOwnerArgent = await ensRegistry.contract.setSubnodeOwner(utils.namehash(extension), utils.sha3(domainName), deploymentAccount);
+	await ensRegistry.verboseWaitForTransaction(setSubnodeOwnerArgent, `Setting Subnode Owner for ${domainName}.${extension}`);
+
+    const ensResolver = await deployer.deploy(ENSResolver);
```
juniset: We do not need to deploy new versions of the `ENSResolver` and the `ReverseRegistrar` as they are not affected by the update.

deployment/7_upgrade_ens.js
```diff
@@ -0,0 +1,31 @@
...
+    const ensManager = await deployer.deploy(ENSManager, {}, domain, walletNode, config.ENS.ensRegistry, ensResolver.contractAddress);
```
juniset: Once the new ENSManager is deployed we need to add the WalletFactory as a manager so that the factory can register subdomains to the wallets it creates:
```
    const ENSManagerAddManagerTx = await ENSManagerWrapper.contract.addManager(config.contracts.WalletFactory);
    await ENSManagerWrapper.verboseWaitForTransaction(ENSManagerAddManagerTx, 'Set the WalletFactory as the manager of the ENS Manager');
```
We also need to change the address of the ENSMAnager in the factory:
```
        const WalletFactoryChangeENSManagerTx = await WalletFactoryWrapper.contract.changeENSManager(ENSManagerWrapper.contractAddress);
        await WalletFactoryWrapper.verboseWaitForTransaction(WalletFactoryChangeENSManagerTx, `Change the ENSManager in the WalletFactory`);
```
Then we need to et the Multisig as the owner of the ENSManager:
```
        const changeOwnerTx = await ENSManagerWrapper.contract.changeOwner(config.contracts.MultiSigWallet);
        await ENSManagerWrapper.verboseWaitForTransaction(changeOwnerTx, `Set the MultiSig as the owner of ENSManagerWrapper}`);
```

deployment/7_upgrade_ens.js
```diff
@@ -0,0 +1,31 @@
...
+
+    const ensReverse = await deployer.deploy(TestReverseRegistrar, {}, ensRegistry.contractAddress, ensResolver.contractAddress);
+}
```
juniset: Finally we need to upload the ABI and the address of the new ENSManager to AWS:
```
    configurator.updateInfrastructureAddresses({
        ENSManager: ENSManagerWrapper.contractAddress
    });
    await configurator.save();

    await Promise.all([
        abiUploader.upload(ENSManagerWrapper, "contracts")
    ]);
```

elenadimitrova: What is the abi actually used for on AWS?

deployment/7_upgrade_ens.js
elenadimitrova: There's these changes https://github.com/argentlabs/argent-contracts/pull/54/files#diff-5f255ee701d7d41347fe8dac7c4b29e2R147

deployment/7_upgrade_ens.js
juniset: The ABIs are exposed via a public end-point and some of them are used by the backend (e.g. the WalletFactory) or the clients (the modules, as part of the build process).

deployment/7_upgrade_ens.js
juniset: I believe the `TestReverseRegistrar.sol` contract is only used for tests. On both Ropsten and Mainnet we use the official ReversRegistrar deployed by the ENS.

deployment/7_upgrade_ens.js
elenadimitrova: The `WalletFactory` itself would have changed as well right?

juniset: No. We anticipated that the ENSManager might need to be changed in the future so we can update  the address of the ENSManager in the factory without redeploying it.

deployment/7_upgrade_ens.js
elenadimitrova: Can we add config for it then, I believe it's at `0x084b1c3C81545d370f3634392De611CaaBFf8148` on both mainnet and ropsten

deployment/7_upgrade_ens.js
elenadimitrova: It's misleading because the `ArgentENSManager` is imported but only the interface definition `IENSManager` inside that file is actually used. I have split the interface into its own file to avoid future confusion.

contracts/test/ENSTest.sol
```diff
@@ -103,14 +127,86 @@ contract TestENSRegistry is ENSRegistry {
...
+        }
+
+        if(_ttl != records[_node].ttl) {
+            records[_node].ttl = _ttl;
+            emit NewTTL(_node, _ttl);
+        }
+    }
+}
+
+contract ENSRegistryWithFallback is ENSRegistry {
```
juniset: I believe it should extend `TestENSRegistry` as `ENSRegistry is just an interface.

contracts/ens/ENSReverseRegistrar.sol
```diff
@@ -0,0 +1,26 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.5.4;
+
+/**
+ * @dev ENS Reverse Registrar interface.
+ */
+contract ENSReverseRegistrar {
```
juniset: We could mark all the functions as `external` and make it an `interface`.

contracts/ens/ENSResolver.sol
```diff
@@ -0,0 +1,29 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.5.4;
+
+/**
+ * @dev ENS Resolver interface.
+ */
+contract ENSResolver {
```
juniset: We could mark all the functions as `external` and make it an `interface`.

deployment/1_setup_test_environment.js
```diff
@@ -15,7 +15,7 @@ const BYTES32_NULL = '0x00000000000000000000000000000000000000000000000000000000
 // For development purpose
 async function deployENSRegistry(deployer, owner, domain) {
     // Deploy the public ENS registry
-    const ENSWrapper = await deployer.deploy(ENS);
+    const ENSWrapper = await deployer.deploy(ENSRegistryWithFallback);
```
juniset: `ENSRegistryWithFallback` takes an argument in the constructor. I believe etherlime will complain if we don't explicitly pass `address(0)` but I could be wrong.

deployment/7_upgrade_ens.js
```diff
@@ -0,0 +1,46 @@
...
+    const ENSManagerAddManagerTx = await ENSManagerNew.contract.addManager(config.contracts.WalletFactory);
+    await ENSManagerNew.verboseWaitForTransaction(ENSManagerAddManagerTx, 'Set the WalletFactory as the manager of the ENS Manager');
+
+    const WalletFactoryChangeENSManagerTx = await WalletFactoryWrapper.contract.changeENSManager(ENSManagerNew.contractAddress);
+    await WalletFactoryWrapper.verboseWaitForTransaction(WalletFactoryChangeENSManagerTx, `Change the ENSManager on the WalletFactory`);
+
+    const ChangeENSManagerOwnerTx = await ENSManagerNew.contract.changeOwner(config.contracts.MultiSigWallet);
+    await ENSManagerNew.verboseWaitForTransaction(ChangeENSManagerOwnerTx, `Set the MultiSig as the owner of ENSManager}`);
+
+    const ChangeDomainOwnerTx = await ENSManagerWrapper.contract.changeRootnodeOwner("0x0000000000000000000000000000000000000000");
```
juniset: Is this what we want? If we throw away ownership in the old registry there will be an interruption of service between the moment we execute that transaction and the moment we claim ownership in the new registry. During that interruption of service we won't be able to create wallets and register subdomains... which means we need to coordinate with the backend team (so they disable the creation of wallets).

On the other hand if we properly transfer ownership to the new Manager in the old registry there is no interruption of service (we can always create wallets and register subdomains). The old registry will just be ignored once the new one is setup.

lib/ens/ENS.sol
```diff
@@ -0,0 +1,37 @@
+pragma solidity >=0.4.24;
+
+/**
+ * ENS Registry interface.
+ * Reference: https://github.com/ensdomains/ens/blob/master/contracts/ENS.sol
+ */
+
+interface ENS {
```
juniset: Great idea to separate non-Argent contracts in a `/lib` folder!

deployment/1_setup_test_environment.js
elenadimitrova: You're right. I fixed it in the tests but not here as these aren't run during testing. How is it we verify the deployment scripts?

contracts/ens/ArgentENSManager.sol
```diff
@@ -127,10 +119,18 @@ contract ArgentENSManager is IENSManager, Owned, Managed, ENSConsumer {
...
             return true;
         }
         return false;
     }
+
+    /**
+    * @dev Gets the official ENS reverse registrar. 
+    * @return Address of the ENS reverse registrar.
+    */
+    function getENSReverseRegistrar() external view returns (address) {
```
juniset: Now that we no longer use `ENSConsumer` this method should be internal or removed all-together (I prefer that option). It would save some gas and make the code cleaner.

L106 could be
```
ENSReverseRegistrar reverseRegistrar = ENSReverseRegistrar(ensRegistry.owner(ADDR_REVERSE_NODE));
```

PR #55, opened 2020-02-05
Add Circle CI config

elenadimitrova:
As a way of safeguarding the repo, we add Circle CI, see https://app.circleci.com/github/argentlabs/argent-contracts/pipelines

Currently this only runs the unit tests, but the plan is to extend it to include linting, integration testing, security testing, gas cost reporting, code coverage checks and more.

Following the merge here, we will mandate all checks pass before merging to release branches, e.g. `develop`.


juniset:
This is not directly related to this commit but I just realised that we compile the contracts for deployment with `runs = 999` (see deploy.sh) but run the tests with `npm run compile` which uses `runs=200`.

Can you update package.json to use 999 as well?

PR #56, opened 2020-02-07
Add linting for Solidity code

elenadimitrova:
Add `ethlint` a.k.a. `solium` to the Circle CI checks and also pre-commit checks locally. 

I've used the opportunity of the upcoming upgrade to do some potentially breaking change fixes to the contracts that are going to be upgraded including moving the `SafeMath` and `string` utils contracts out of `contracts` and into the `lib\utils` folder. 
The remaining contract changes are non-breaking (e.g. white spaces changes). 
Any breaking changes needed are ignored for now in `.soliumignore` and will be dealt with on their next upgrade.

Also here we include a small fix to the Circle CI node version to the correct v8.12.


.soliumignore
```diff
@@ -0,0 +1,51 @@
+node_modules
+migrations
+openzeppelin-solidity
+lib
+contracts/Migrations.sol
```
juniset: We are not using a `Migrations.sol` contract so we can remove it.

.soliumignore
```diff
@@ -0,0 +1,51 @@
...
+contracts/MultiSigWallet.sol
+contracts/modules/common/RelayerModule.sol
```
juniset: Why are we ignoring these 2 contracts?

PR #58, opened 2020-02-12
Create2

juniset:
- Enabling the `WalletFactory` to create wallets counterfactually with CREATE2.
- Adding the first guardian as part of the wallet creation.


contracts/wallet/WalletFactory.sol
```diff
@@ -34,60 +38,161 @@ contract WalletFactory is Owned, Managed {
...
+     * The wallet is created using the CREATE opcode.
      * @param _owner The account address.
      * @param _modules The list of modules.
-     * @param _label Optional ENS label of the new wallet (e.g. franck).
+     * @param _label (Optional) ENS label of the new wallet, e.g. franck.
      */
-    function createWallet(address _owner, address[] calldata _modules, string calldata _label) external onlyManager {
-        require(_owner != address(0), "WF: owner cannot be null");
-        require(_modules.length > 0, "WF: cannot assign with less than 1 module");
-        require(ModuleRegistry(moduleRegistry).isRegisteredModule(_modules), "WF: one or more modules are not registered");
```
elenadimitrova: I did some gas cost testing for the difference of failing a `require` before and after a contract creation call - in our case the `new Proxy` call. The results were `21,754` vs `114,018` gas with a simple contract in Remix. Need to investigate further but would prefer to abstract the `require`s in a modifier instead of moving them until after the core logic executes.

contracts/wallet/WalletFactory.sol
```diff
@@ -34,60 +38,161 @@ contract WalletFactory is Owned, Managed {
...
-            // initialise the wallet with the owner and the extended modules
-            BaseWallet(wallet).init(_owner, extendedModules);
-            // register ENS
-            registerWalletENS(wallet, _label);
-            // remove the factory from the authorised modules
-            BaseWallet(wallet).authoriseModule(address(this), false);
-        } else {
-            // initialise the wallet with the owner and the modules
-            BaseWallet(wallet).init(_owner, _modules);
+        configureWallet(BaseWallet(wallet), _owner, _modules, _label, address(0));
```
elenadimitrova: Just so I understand the change of logic between here and in `configureWallet`, why has the condition of having and ENS label changed? Also why does the WalletFactory need to be a temporarily registered module in the new wallet? And if it needs to be why does it have to be the first?

contracts/wallet/WalletFactory.sol
```diff
@@ -34,60 +38,161 @@ contract WalletFactory is Owned, Managed {
...
+    )
+        external
+        onlyManager
+    {
+        bytes32 newsalt = keccak256(abi.encodePacked(_salt, _owner, _modules));
+        bytes memory code = abi.encodePacked(type(Proxy).creationCode, uint256(walletImplementation));
+        address payable wallet;
+        // solium-disable-next-line security/no-inline-assembly
+        assembly {
+            wallet := create2(0, add(code, 0x20), mload(code), newsalt)
```
elenadimitrova: Since 0.6.2 there is no need to drop to the assembly level https://solidity.readthedocs.io/en/v0.6.2/control-structures.html#salted-contract-creations-create2 as we can call `new` contract with `CREATE2` in Solidity itself.

contracts/wallet/WalletFactory.sol
```diff
@@ -34,60 +38,161 @@ contract WalletFactory is Owned, Managed {
...
```
elenadimitrova: Same as above.

contracts/wallet/WalletFactory.sol
```diff
@@ -34,60 +38,161 @@ contract WalletFactory is Owned, Managed {
...
+        configureWallet(BaseWallet(wallet), _owner, _modules, _label, address(0));
+    }
+
+    /**
+     * @dev Lets the manager create a wallet for an owner account at a specific address.
+     * The wallet is initialised with a list of modules, a first guardian, and an optional ENS.
+     * The wallet is created using the CREATE2 opcode.
+     * @param _owner The account address.
+     * @param _modules The list of modules.
+     * @param _label (Optional) ENS label of the new wallet, e.g. franck.
```
elenadimitrova: For optional parameters can we have function overloads instead? Passing default type values is not very clean

contracts/wallet/WalletFactory.sol
juniset: I have extracted all the `require` in the internal method `validateInputs(address _owner, address[] memory _modules, string memory _label)`. That method is called before the `Proxy` is created.

contracts/wallet/WalletFactory.sol
```diff
@@ -34,60 +38,161 @@ contract WalletFactory is Owned, Managed {
...
-        require(ModuleRegistry(moduleRegistry).isRegisteredModule(_modules), "WF: one or more modules are not registered");
-        // create the proxy
+    function createWallet(
+        address _owner,
+        address[] calldata _modules,
+        string calldata _label
+    )
+        external
+        onlyManager
+    {
```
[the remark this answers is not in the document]
juniset: I usually use 150 as well for max length. However here we have 4 methods that have the same functionality but differ in there parameter list and implementation; I personally find that it makes the code more readable if they have the same display (easier to see their differences). Since some of them are above the 150 threshold I prefer to display them all with breaking lines.

contracts/wallet/WalletFactory.sol
juniset: The WalletFactory needs to be temporarily registered as a module so that:
- it can invoke the wallet and do the reverse ENS registration
- call the `GuardianStorage` contract and add the first guardian

contracts/wallet/WalletFactory.sol
juniset: The current pragma version on our contracts is 0.5.4. so we should not use 0.6.x features right?

contracts/wallet/WalletFactory.sol
juniset: I have decided to remove the optionality of the ENS label which must now be always defined. 

This was required a year ago for a trick we had in the backend but this trick is no longer there.

contracts/wallet/WalletFactory.sol
elenadimitrova: 0.5.4 is a year old version and we should move to 0.6.x at some point. This is a worthwhile feature to push us to upgrade.

juniset: Agree that we should migrate to 0.6.x at some point but let's do it after the next contract deployment (ENS migration and this PR).

test/factory.js
```diff
@@ -122,24 +133,30 @@ describe("Test Wallet Factory", function () {
...
         it("should fail to create when there is no modules", async () => {
+            let label = "wallet" + index; 
             let modules = [];
-            await assert.revert(factory.from(deployer).createWallet(owner.address, modules, NO_ENS, { gasLimit: 200000 }), "should fail when modules is empty");
+            await assert.revert(factory.from(deployer).createWallet(owner.address, modules, label), "should fail when modules is empty");
+        });
+
+        it("should fail to create when there is no ENS", async () => {
+            let modules = [module1.contractAddress, module2.contractAddress];
+            await assert.revert(factory.from(infrastructure).createWallet(owner.address, modules, NO_ENS), "should fail when ENS is already used");
```
elenadimitrova: All `revert` failures should be checked for the respective error message back, rather than a general failure of the transaction. It's the only way to check that the failure we're testing for is the one that actually occurred.

test/factory.js
```diff
@@ -272,27 +300,51 @@ describe("Test Wallet Factory", function () {
...
+            let modules = [module1.contractAddress, module2.contractAddress]; 
+            // we get the future address
+            let futureAddr = await factory.getAddressForCounterfactualWallet(owner.address, modules, salt); 
+            // We send ETH to the address
+            await infrastructure.sendTransaction({ to: futureAddr, value: ethers.utils.bigNumberify('1000000000000000000') });
+            // we create the wallet
+            let tx = await factory.from(infrastructure).createCounterfactualWallet(owner.address, modules, label, salt);
+            let txReceipt = await factory.verboseWaitForTransaction(tx);
+            let wallet = deployer.wrapDeployedContract(Wallet, futureAddr);
+            assert.isTrue(await utils.hasEvent(txReceipt, wallet, "Received"), "should have generated Received event");
```
elenadimitrova: There are 4 different `Received` events in contracts, how do I know `BaseWallet.init()` was the one that was triggered?

contracts/wallet/WalletFactory.sol
```diff
@@ -56,38 +69,135 @@ contract WalletFactory is Owned, Managed {
...
+        validateInputs(_owner, _modules, _label);
+        bytes32 newsalt = keccak256(abi.encodePacked(_salt, _owner, _modules));
+        bytes memory code = abi.encodePacked(type(Proxy).creationCode, uint256(walletImplementation));
+        address payable wallet;
+        // solium-disable-next-line security/no-inline-assembly
+        assembly {
+            wallet := create2(0, add(code, 0x20), mload(code), newsalt)
+            if iszero(extcodesize(wallet)) { revert(0, returndatasize) }
```
elenadimitrova: Instead of checking the code at address, can we simply have a `require(wallet != address(0), "error message"`. `create2` returns an address when successful and `0` if not.

contracts/wallet/WalletFactory.sol
```diff
@@ -56,38 +69,135 @@ contract WalletFactory is Owned, Managed {
...
+     * @dev Lets the manager create a wallet for an owner account at a specific address.
+     * The wallet is initialised with a list of modules, a first guardian, and an ENS.
+     * The wallet is created using the CREATE2 opcode.
+     * @param _owner The account address.
+     * @param _modules The list of modules.
+     * @param _label ENS label of the new wallet, e.g. franck.
+     * @param _guardian The guardian address.
+     * @param _salt The salt.
+     */
+    function createCounterfactualWalletWithGuardian(
```
elenadimitrova: Just a final thing on overloads, I think it makes sense for the following to be overloads to at most 2 functions:
`createWallet(address _owner, address[] calldata _modules, string calldata _label)`
`function createWalletWithGuardian(address _owner, address[] calldata _modules, string calldata _label, address _guardian)`
`function createCounterfactualWallet(address _owner, address[] calldata _modules, string calldata _label, bytes32 _salt)`
`createCounterfactualWalletWithGuardian(address _owner, address[] calldata _modules, string calldata _label,         address _guardian, bytes32 _salt)`

test/factory.js
juniset: Done. I replaced `assert.revert` by `assert.revertWith` which is now available with etherlime.

We should propagate this change to all tests that use `assert.revert`.

test/factory.js
juniset: Since we are looking at the events generated by a call to `createCounterfactualWallet()` only `BaseWalet.init()` can trigger the event.

I have nonetheless added additional checks in the test to verify that the amount is correct and that the `sender` property of the event is `address(0)`.

contracts/wallet/WalletFactory.sol
juniset: This is what I used initially but playing with CREATE2 in another project I noticed that create2 sometimes returns an address when no code is deployed. I did not investigate further but decided to use the check used by openzeppelin and other projects.

contracts/wallet/WalletFactory.sol
juniset: We've experienced many problems with overloading and the different clients used in our stack in the past (Web3J, web3.swift, web3.js and ethers.js). Maybe this has been fixed but I'd prefer to avoid overloading as much as I can.

contracts/wallet/WalletFactory.sol
elenadimitrova: Open Zeppelin don't use that though https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Create2.sol#L26 ?

test/factory.js
```diff
@@ -87,7 +87,7 @@ describe("Test Wallet Factory", function () {
         index++;
     });
 
-    describe("Create wallets with CREATE", () => {
+    describe("Create wallets with CREATE", () => { return;
```
elenadimitrova: ?

contracts/wallet/WalletFactory.sol
juniset: I guess they no longer do https://blog.openzeppelin.com/getting-the-most-out-of-create2/.

But it doesn't change the fact that I witnessed some unreliable behavior when checking solely on the returned address so I'll rather keep checking on the bytecode.

deployment/8_upgrade_wf.js
```diff
@@ -0,0 +1,89 @@
...
+        abiUploader.upload(WalletFactoryWrapper, "contracts"),
+        abiUploader.upload(BaseWalletWrapper, "contracts")
+    ]);
+
+    ////////////////////////////////////
+    // Set authorisations
+    ////////////////////////////////////
+
+    // Set the WalletFactory as a manager of ENSManager
+    await multisigExecutor.executeCall(ENSManagerWrapper, "addManager", [WalletFactoryWrapper.contractAddress]);
```
elenadimitrova: In the ENS script 7, we need to remove the `addManager` call for the old wallet factory.

PR #60, opened 2020-02-13
Upgrade etherlime

elenadimitrova:
Upgrades `etherlime` from 2.2.4 to 2.2.6 to incorporate the fix for flattening contracts https://github.com/LimeChain/etherlime/issues/302

Additionally it provides support for compiling 0.6.x code.

I've also included an .nvmrc file to set the node version to 8.12.0


elenadimitrova:
Let's please avoid updating with merge commits.

PR #64, opened 2020-02-25
Enable code coverage for contracts

elenadimitrova:
With the release of `etherlime@2.3.0` its `coverage` functionality now uses `solidity-coverage` 🎉 . It's still rough around the edges but at least we managed to get a code coverage report out by hacking some of the missing bits.

Additionally all instances of `assert.revert` have been updated to `assert.revertWith` to validate the returned error message against the expected point of failure.

Starting test coverage:
| |% Statements | % Branch  | % Functions  | % Lines
| -------------| ------------- | ------------- | ------------- | ------------- |
Before | 71.16  | 51.32 | 73.27  | 71.44  |
Now     | 74.25 | 60.59 | 77.67 | 74.43 | 

We also add threshold checks for test coverage for the current values so to ensure that we don't drop back to lower numbers. With time we'll gradually increase coverage (specifically aiming for near 100% branch coverage).

Here we also move all external contracts out into `/lib` folder (which was most of the contracts previously found in `/contracts/test`. Also we remove `DappRegistry` and `DappStorage` as obsolete and no longer used (todo: still need to remove the config entries for those).

Logged https://github.com/LimeChain/etherlime/issues/315 to deal with inconsistency in overlaod selection.


test/simpleUpgrader.js
```diff
@@ -27,11 +27,19 @@ describe("Test SimpleUpgrader", function () {
...
-            let module = await deployer.deploy(Module, {}, registry.contractAddress, false, 0);
-            await registry.registerModule(module.contractAddress, ethers.utils.formatBytes32String(name));
-            let isRegistered = await registry.isRegisteredModule(module.contractAddress);
+            let initialModule = await deployer.deploy(Module, {}, registry.contractAddress, false, 0);
+            await registry.registerModule(initialModule.contractAddress, ethers.utils.formatBytes32String(name));
+            let isRegistered;
+            // Here we adjust how we call isRegisteredModule which has 2 overlaods, one accepting a single address
+            // and a second accepting an array of addresses. Behaviour as to which overload is selected to run
+            // differs between CI and Coverage environments, adjusted for this here
+            if (process.env.CI && process.env.SOLIDITY_COVERAGE) {
```
olivdb: The first branch of this `if` needs to be executed whenever `SOLIDITY_COVERAGE=1`, instead of when `CI=1 && SOLIDITY_COVERAGE=1`, otherwise `npm run test:coverage` fails on this test when ran locally (when `CI=0`).

scripts/provision_lib_artefacts.sh
```diff
@@ -0,0 +1,27 @@
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
olivdb: Since we now use the `scripts` directory to contain any type of scripts (as opposed to only containing scripts that perform on-chain transactions, as was the case before), I would suggest to also move `deploy.sh` to this directory.

test/guardianManager.js
```diff
@@ -143,7 +140,7 @@ describe("GuardianManager", function () {
                 }
             });
 
-            it("should add many Guardians (relayed transaction)", async () => {
+            it("should add many Guardians", async () => {
```
olivdb: Why not name tests in a way that distinguishes relayed from non-relayed transactions?

elenadimitrova: It's obvious from the test itself, if a `manager.relay` call is there then it's _relayed_. When searching for relayed transactions that's the string I'll search for, rather than the name of the test. The name should be a representation of the condition being tested rather than details of its execution. This is also prone to errors as there's no check the title matches the execution.

test/simpleUpgrader.js
elenadimitrova: I'm not sure what you mean. This test passes both in CI and locally and in `test` and in `test:coverage`. How do I replicate the problem you are talking about?

scripts/provision_lib_artefacts.sh
elenadimitrova: Makes sense.

test/simpleUpgrader.js
olivdb: Running `npm run test:coverage` fails with:
```
348 passing (13m)
  1 failing

  1) Test SimpleUpgrader
       Registering modules
         should register modules in the registry:
     Error: invalid value for array
      at resolveAddresses (node_modules/etherlime-lib/node_modules/ethers/contract.js:92:19)
      at /Users/ovdbigge/argent/argcover/node_modules/etherlime-lib/node_modules/ethers/contract.js:78:29
      at Array.forEach (<anonymous>)
      at resolveAddresses (node_modules/etherlime-lib/node_modules/ethers/contract.js:70:19)
      at EtherlimeGanacheWrapper.isRegisteredModule (node_modules/etherlime-lib/node_modules/ethers/contract.js:141:16)
      at Context.it (.coverage_tests/simpleUpgrader.js:39:47)
      at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:189:7)

Error: Some of the test scenarios failed!
    at Object.runCoverage (/Users/ovdbigge/argent/argcover/node_modules/etherlime/cli-commands/etherlime-test/etherlime-coverage.js:84:9)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:189:7)
[...]
TypeError: Cannot read property 'stop' of undefined
    at Object.commandProcessor [as handler] (/Users/ovdbigge/argent/argcover/node_modules/etherlime/cli-commands/commands.js:421:21)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:189:7)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! argent-contracts@1.0.0 test:coverage: `bash ./scripts/provision_lib_artefacts.sh & SOLIDITY_COVERAGE=1 npx etherlime coverage --solcVersion 0.5.4 && istanbul check-coverage --statements 74 --branches 60 --functions 77 --lines 74`
npm ERR! Exit status 1
[...]
```

Replacing L36 with `if (process.env.SOLIDITY_COVERAGE) {` fixes the issue.

elenadimitrova: It runs okay for me locally :/

test/guardianManager.js
olivdb: > The name should be a representation of the condition being tested rather than details of its execution.

I think a test name should ideally contain no less information than "when in state S0, performing action A yields state S1" (unless obvious from context). Here the two tests (relayed and not relayed) share the same initial and expected final states but differ by the action being tested. Even though we expect those two actions to always be equivalent, they are still distinct actions performed by a user (which is why we write a different test for each). I'm not sure what you mean by 'details of its execution'.

> This is also prone to errors as there's no check the title matches the execution.

Isn't this the case with any test? How does specifying the action being performed make it more error prone?

Also, more pragmatically, etherlime test does not always return the line on which the test failed, so being able to search the test suite using a test name that is more specific saves a bit of time.

test/simpleUpgrader.js
olivdb: Weird :( You don't have a local environment variable `CI` set to `1` by any chance?

test/guardianManager.js
elenadimitrova: Perhaps this needs a larger discussion. I have started a [list of items to discuss around unit testing](https://trello.com/c/h7wl8ohs) and will setup a meeting towards the end of this sprint to discuss and agree what if anything needs changing. For now I'll revert back the test names above.

test/simpleUpgrader.js
elenadimitrova:
```
process.env.CI undefined
process.env.SOLIDITY_COVERAGE 1
```
Have you switched to node 10?

elenadimitrova: @juniset just ran coverage okay as well. Definitely something with your environment.

olivdb: Yes I'm using node 10.12.

What is the reason behind checking for `process.env.CI == 1` in the first place btw? Why isn't it in principle sufficient to check for `process.env.SOLIDITY_COVERAGE` ?

elenadimitrova: It's to work around a possible bug in either `ethers` or `etherlime` which picks up the wrong overload for `isRegisteredModule` and fails in CI, see [sample failing build](https://circleci.com/gh/argentlabs/argent-contracts/229).

PR #65, opened 2020-03-02
Fix/wf audit comments

juniset:
Addressing the suggestions following the audit of `ArgentENSManager.sol` and `WalletFactory.sol`:
- added the guardian to the salt when creating a wallet with CREATE2 to create a deterministic relationship between wallet address and management rights
- extracted the creation logic of `createWallet` and `createCounterfactualWallet` to an internal method to remove duplicated code.

### DIFF
--- a/contracts/wallet/WalletFactory.sol
+++ b/contracts/wallet/WalletFactory.sol
@@ -84,10 +84,7 @@ contract WalletFactory is Owned, Managed {
         external
         onlyManager
     {
-        validateInputs(_owner, _modules, _label);
-        Proxy proxy = new Proxy(walletImplementation);
-        address payable wallet = address(proxy);
-        configureWallet(BaseWallet(wallet), _owner, _modules, _label, address(0));
+        _createWallet(_owner, _modules, _label, address(0));
     }
 
     /**
@@ -109,10 +106,7 @@ contract WalletFactory is Owned, Managed {
         onlyManager
         guardianStorageDefined
     {
-        validateInputs(_owner, _modules, _label);
-        Proxy proxy = new Proxy(walletImplementation);
-        address payable wallet = address(proxy);
-        configureWallet(BaseWallet(wallet), _owner, _modules, _label, _guardian);
+        _createWallet(_owner, _modules, _label, _guardian);
     }
 
     /**
@@ -133,16 +127,7 @@ contract WalletFactory is Owned, Managed {
         external
         onlyManager
     {
-        validateInputs(_owner, _modules, _label);
-        bytes32 newsalt = keccak256(abi.encodePacked(_salt, _owner, _modules));
-        bytes memory code = abi.encodePacked(type(Proxy).creationCode, uint256(walletImplementation));
-        address payable wallet;
-        // solium-disable-next-line security/no-inline-assembly
-        assembly {
-            wallet := create2(0, add(code, 0x20), mload(code), newsalt)
-            if iszero(extcodesize(wallet)) { revert(0, returndatasize) }
-        }
-        configureWallet(BaseWallet(wallet), _owner, _modules, _label, address(0));
+        _createCounterfactualWallet(_owner, _modules, _label, address(0), _salt);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


contracts/wallet/WalletFactory.sol
```diff
@@ -200,6 +176,30 @@ contract WalletFactory is Owned, Managed {
...
 
+    /**
+     * @dev Gets the address of a counterfactual wallet with a first default guardian.
+     * @param _owner The account address.
+     * @param _modules The list of modules.
+     * @param _guardian The guardian address.
+     * @param _salt The salt.
+     * @return the address that the wallet will have when created using CREATE2 and the same input parameters.
+     */
+    function getAddressForCounterfactualWalletWithGuardian(
```
elenadimitrova: If we are going to be deduplicating code, we should do it here as well for `getAddressForCounterfactualWalletWithGuardian` and `getAddressForCounterfactualWallet`

juniset: Agreed and changed.

contracts/wallet/WalletFactory.sol
```diff
@@ -239,15 +233,64 @@ contract WalletFactory is Owned, Managed {
...
+    }
+
     /**
      * @dev Helper method to configure a wallet for a set of input parameters.
      * @param _wallet The target wallet
      * @param _owner The account address.
      * @param _modules The list of modules.
      * @param _label ENS label of the new wallet, e.g. franck.
-     * @param _guardian (Optional) The guardian address.
+     * @param _guardian The guardian address.
```
elenadimitrova: The guardian address is again optional here.

contracts/wallet/WalletFactory.sol
```diff
@@ -239,15 +233,64 @@ contract WalletFactory is Owned, Managed {
         //do nothing
     }
 
+    // *************** Internal Functions ********************* //
+
+    /**
+     * @dev Helper method to create a wallet for an owner account.
+     * The wallet is initialised with a list of modules, a first guardian, and an ENS..
```
elenadimitrova: Can we remove this ".." after ENS. It's leading me to think there's something else not said there.

contracts/wallet/WalletFactory.sol
```diff
@@ -239,15 +233,64 @@ contract WalletFactory is Owned, Managed {
...
+     * The wallet is created using the CREATE opcode.
+     * @param _owner The account address.
+     * @param _modules The list of modules.
+     * @param _label ENS label of the new wallet, e.g. franck.
+     * @param _guardian The guardian address.
```
elenadimitrova: The guardian is optional here. Should add it to the description.

contracts/wallet/WalletFactory.sol
```diff
@@ -269,19 +312,58 @@ contract WalletFactory is Owned, Managed {
...
         // emit event
         emit WalletCreated(address(_wallet), _owner, _guardian);
     }
 
+    /**
+     * @dev Gets the address of a counterfactual wallet.
+     * @param _owner The account address.
+     * @param _modules The list of modules.
+     * @param _salt The salt.
+     * @param _guardian The guardian address.
```
elenadimitrova: The guardian is optional here.

contracts/wallet/WalletFactory.sol
```diff
@@ -194,10 +170,28 @@ contract WalletFactory is Owned, Managed {
...
+     * @param _owner The account address.
+     * @param _modules The list of modules.
+     * @param _guardian The guardian address.
+     * @param _salt The salt.
+     * @return the address that the wallet will have when created using CREATE2 and the same input parameters.
+     */
+    function getAddressForCounterfactualWalletWithGuardian(
+        address _owner,
+        address[] calldata _modules,
+        address _guardian,
```
elenadimitrova: We should validate the guardian address is non empty here as it's not done further down the call chain

contracts/wallet/WalletFactory.sol
```diff
@@ -109,10 +106,7 @@ contract WalletFactory is Owned, Managed {
         onlyManager
         guardianStorageDefined
     {
-        validateInputs(_owner, _modules, _label);
-        Proxy proxy = new Proxy(walletImplementation);
-        address payable wallet = address(proxy);
-        configureWallet(BaseWallet(wallet), _owner, _modules, _label, _guardian);
+        _createWallet(_owner, _modules, _label, _guardian);
```
elenadimitrova: We should validate the guardian address is non empty here as it's not done further down the call chain

contracts/wallet/WalletFactory.sol
juniset: I don't agree. If the guardian is empty (address(0)) it will be ignored both by this method and when the wallet is created by calling `createCounterfactualWalletWithGuardian()` with an empty guardian. So the address will match the created wallet as expected.

juniset: Ok I actually see your point. It's misleading to call `createCounterfactualWalletWithGuardian()` and have no guardian added. I've fixed it.

contracts/wallet/WalletFactory.sol
juniset: Fixed as well.

PR #73, opened 2020-03-19
More flexible approve and call

juniset:
Update of the `approveTokenAndCallContract` method of `ApprovedTransfer` module to support the consumer of the approved tokens to be different than the called contract. This is needed to e.g. interact with the 0x protocol where the user needs to approve a specific `AssetProxy` contract before submitting an order to the `Exchange` contract.

Also addition of the `execute.sh` script that was used to execute the `register.js` and `deregister.js` scripts but was kept locally so far.

### DIFF
--- a/contracts/modules/ApprovedTransfer.sol
+++ b/contracts/modules/ApprovedTransfer.sol
@@ -79,7 +79,7 @@ contract ApprovedTransfer is BaseModule, RelayerModule, BaseTransfer {
 
     /**
     * @dev lets the owner do an ERC20 approve followed by a call to a contract.
-    * The address to approve may be different then the contract to call.
+    * The address to approve may be different than the contract to call.
     * We assume that the contract does not require ETH.
     * @param _wallet The target wallet.
     * @param _token The token to approve.
--- a/contracts/modules/ApprovedTransfer.sol
+++ b/contracts/modules/ApprovedTransfer.sol
@@ -79,18 +79,21 @@ contract ApprovedTransfer is BaseModule, RelayerModule, BaseTransfer {
 
     /**
     * @dev lets the owner do an ERC20 approve followed by a call to a contract.
-    * We assume that the contract will pull the tokens and does not require ETH.
+    * The address to approve may be different then the contract to call.
+    * We assume that the contract does not require ETH.
     * @param _wallet The target wallet.
     * @param _token The token to approve.
-    * @param _contract The address of the contract.
+    * @param _spender The address to approve.
     * @param _amount The amount of ERC20 tokens to approve.
+    * @param _contract The contract to call.
     * @param _data The encoded method data
     */
     function approveTokenAndCallContract(
         BaseWallet _wallet,
         address _token,
-        address _contract,
+        address _spender,
         uint256 _amount,
+        address _contract,
         bytes calldata _data
     )
         external
@@ -98,7 +101,7 @@ contract ApprovedTransfer is BaseModule, RelayerModule, BaseTransfer {
         onlyWhenUnlocked(_wallet)
     {
         require(!_wallet.authorised(_contract) && _contract != address(_wallet), "AT: Forbidden contract");
-        doApproveToken(_wallet, _token, _contract, _amount);
+        doApproveToken(_wallet, _token, _spender, _amount);
         doCallContract(_wallet, _contract, 0, _data);
     }
 
--- a/lib/test/TestContract.sol
+++ b/lib/test/TestContract.sol
@@ -1,5 +1,5 @@
 pragma solidity ^0.5.4;
-
+import "./TokenConsumer.sol";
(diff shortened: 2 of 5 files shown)


test/approvedTransfer.js
```diff
@@ -157,7 +157,7 @@ describe("Test Approved Transfer", function () {
         });
     });
 
-    describe("Transfer approved by smart-contract guardians", () => {
+    describe("Transfer approved by smart-contract guardians", () => { return;
```
olivdb: I think you forgot to remove a bunch of `return` statements in this file

juniset: Ooops,  indeed.

Fixed.

contracts/modules/ApprovedTransfer.sol
```diff
@@ -79,26 +79,29 @@ contract ApprovedTransfer is BaseModule, RelayerModule, BaseTransfer {
 
     /**
     * @dev lets the owner do an ERC20 approve followed by a call to a contract.
-    * We assume that the contract will pull the tokens and does not require ETH.
+    * The address to approve may be different then the contract to call.
```
elenadimitrova: Typo: then -> than

lib/test/TestContract.sol
```diff
@@ -23,4 +28,19 @@ contract TestContract {
...
+    function setStateAndPayTokenWithConsumer(uint256 _state, address _erc20, uint256 _amount) public {
+        bool success = tokenConsumer.consume(_erc20, msg.sender, address(this), _amount);
+        if (success) {
+            state = _state;
+            emit StateSet(_state, _amount);
+        }
+    }
+}
+
+contract TokenConsumer {
```
elenadimitrova: I have gone through to separate all contracts into their own files, can we do the same here? There already are a lot of Test* contracts under `lib/test/` so this can also be added there.

test/approvedTransfer.js
```diff
@@ -307,47 +307,86 @@ describe("Test Approved Transfer", function () {
...
             await addGuardians([guardian1, ...(await createSmartContractGuardians([guardian2, guardian3]))]);
             let count = (await guardianManager.guardianCount(wallet.contractAddress)).toNumber();
             assert.equal(count, 3, '3 guardians should be active');
             let before = await erc20.balanceOf(contract.contractAddress);
             // should succeed with 2 confirmations
             dataToTransfer = contract.contract.interface.functions['setStateAndPayToken'].encode([2, erc20.contractAddress, amountToApprove]);
-            let txReceipt = await manager.relay(transferModule, "approveTokenAndCallContract", [wallet.contractAddress, erc20.contractAddress, contract.contractAddress, amountToApprove, dataToTransfer], wallet, [owner, ...sortWalletByAddress([guardian1, guardian2])]);
+            let txReceipt = await manager.relay(transferModule, "approveTokenAndCallContract", [wallet.contractAddress, erc20.contractAddress, contract.contractAddress, amountToApprove, contract.contractAddress, dataToTransfer], wallet, [owner, ...sortWalletByAddress([guardian1, guardian2])]);
```
elenadimitrova: Add check the `txReceipt` is ok

test/approvedTransfer.js
```diff
@@ -307,47 +307,86 @@ describe("Test Approved Transfer", function () {
...
+        });
+
+        it('should approve token for a spender then call a contract with 3 guardians, spender != contract', async () => {
+            await addGuardians([guardian1, ...(await createSmartContractGuardians([guardian2, guardian3]))]);
+            let count = (await guardianManager.guardianCount(wallet.contractAddress)).toNumber();
+            assert.equal(count, 3, '3 guardians should be active');
+            let before = await erc20.balanceOf(contract.contractAddress); 
+            // should succeed with 2 confirmations
+            dataToTransfer = contract.contract.interface.functions['setStateAndPayTokenWithConsumer'].encode([2, erc20.contractAddress, amountToApprove]);
+            let txReceipt = await manager.relay(transferModule, "approveTokenAndCallContract", [wallet.contractAddress, erc20.contractAddress, consumer, amountToApprove, contract.contractAddress, dataToTransfer], wallet, [owner, ...sortWalletByAddress([guardian1, guardian2])]);
```
elenadimitrova: Add check the `txReceipt` is ok

lib/test/TestContract.sol
juniset: Done.

test/approvedTransfer.js
juniset: I've removed the `txReceipt` as it isn't used anywhere in the test and we already test for the functional success of the test.

test/approvedTransfer.js
juniset: Same as above.

contracts/modules/ApprovedTransfer.sol
juniset: Done.

PR #80, opened 2020-03-30
Update develop with changes from master

elenadimitrova:


elenadimitrova:
Did a manual merge instead.

PR #84, opened 2020-04-01
Enhance approveAndCall logic

elenadimitrova:
In the current implementation of `approveTokenAndCallContract`, the amount of tokens approved overwrites any existing amount of the same token and spender. Here we add logic to restore the original approved amount after the `approveTokenAndCallContract` call in both (`ApprovedTransfer` and `TransferManager`) and emit a new dedicated event with the following definition:

```
event ApprovedAndCalledContract(
        address indexed wallet,
        address indexed to,
        address spender,
        address indexed token,
        uint256 amountApproved,
        uint256 amountSpent,
        bytes data
    );
```

This PR also enables a different contract address for the approved spender and the called contract for `TransferManager` module, analogous to behaviour already implemented for `ApprovedTransfer` in #73 

Also I needed `SafeMath` in the `BaseTransfer` contract so pushing it there meant this was provisioned to both `ApprovedTransfer` and `TransferManager`.

We also decommission the `config.modules.TokenTransfer` config entry.


contracts/modules/common/BaseTransfer.sol
```diff
@@ -77,4 +85,48 @@ contract BaseTransfer is BaseModule {
...
+        uint256 _amount,
+        address _contract,
+        bytes memory _data
+    )
+        internal
+    {
+        uint256 currentAllowance = ERC20(_token).allowance(address(_wallet), _spender);
+
+        // Approve the desired amount
+        bytes memory methodData = abi.encodeWithSignature("approve(address,uint256)", _spender, _amount);
```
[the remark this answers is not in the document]
elenadimitrova: The problem with that is that the new allowance amount left can be different to the original allowance.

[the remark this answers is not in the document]
th3m477: Also we would be allowing that TX to use your existing allowance, so should set explicit amount before the call, make the call, and then reset to old allowance

contracts/modules/common/BaseTransfer.sol
```diff
@@ -32,7 +32,15 @@ contract BaseTransfer is BaseModule {
...
-
+    event ApprovedAndCalledContract(
+        address indexed wallet,
+        address indexed to,
+        address spender,
+        address indexed token,
+        uint256 amountApproved,
+        uint256 amountSpent,
+        bytes data
+    );
```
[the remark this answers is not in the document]
th3m477: I don't think uint256 values need to be indexed generally, makes no client difference here as it's a new event.

Addresses make more sense as we are more likely to do exploratory network analysis in the future, or quick filters (i.e. "show me all the transfers I made to X address"), so more of a case for `spender`, but up to you @juniset @elenadimitrova

deployment/7_upgrade_1_6.js
```diff
@@ -100,6 +101,20 @@ const deploy = async (network) => {
...
+    TransferManager,
+    {},
+    config.contracts.ModuleRegistry,
+    config.modules.TransferStorage,
+    config.modules.GuardianStorage,
+    config.contracts.TokenPriceProvider,
+    config.settings.securityPeriod || 0,
+    config.settings.securityWindow || 0,
+    config.settings.defaultLimit || "1000000000000000000",
+    ["test", "staging", "prod"].includes(network) ? config.modules.TokenTransfer : "0x0000000000000000000000000000000000000000",
```
juniset, line 114: It should be `config.modules.TransferManager` and not `config.modules.TokenTransfer`.

deployment/7_upgrade_1_6.js
```diff
@@ -12,11 +12,12 @@ const TokenPriceProvider = require("../build/TokenPriceProvider");
...
+const TransferManager = require("../build/TransferManager");
 
 const utils = require("../utils/utilities.js");
 
 const TARGET_VERSION = "1.6.0";
-const MODULES_TO_ENABLE = ["ApprovedTransfer", "RecoveryManager", "MakerV2Manager"];
+const MODULES_TO_ENABLE = ["ApprovedTransfer", "RecoveryManager", "MakerV2Manager", "TransferManager"];
 const MODULES_TO_DISABLE = ["UniswapManager"];
 
 const BACKWARD_COMPATIBILITY = 3;
```
juniset, line 23: Didn't we say we would set `BACKWARD_COMPATIBILITY` to `1` and remove the part of the script where we create a  `LegacyUpgrader`? Or are we addressing this in a separate PR?

contracts/modules/common/BaseTransfer.sol
```diff
@@ -77,4 +85,53 @@ contract BaseTransfer is BaseModule {
...
+
+        // Calculate the approved amount that was spent after the call
+        uint256 unusedAllowance = ERC20(_token).allowance(address(_wallet), _spender);
+        uint256 usedAllowance = SafeMath.sub(totalAllowance, unusedAllowance);
+        // Ensure the amount spent does not exceed the amount approved for this call
+        require(usedAllowance <= _amount, "BT: insufficient amount for call");
+
+        // Restore the original allowance amount.
+        methodData = abi.encodeWithSignature("approve(address,uint256)", _spender, existingAllowance);
+        invokeWallet(address(_wallet), _token, 0, methodData);
```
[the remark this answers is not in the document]
elenadimitrova: I intentionally omitted it as we won't pay for storage updates when the value doesn't change. So implicitly this is enforced anyway while keeping the code simpler.

deployment/7_upgrade_1_6.js
elenadimitrova: There's a separate PR for it https://github.com/argentlabs/argent-contracts/pull/91 as I wanted to double check the clients can support the upgrade through 1.4 -> 1.6

deployment/7_upgrade_1_6.js
elenadimitrova: But `config.modules.TokenTransfer` is the address of the old `config.modules.TransferManager` no?

elenadimitrova: Maybe when upgrading the TransferManager we should copy its old value here though

contracts/modules/common/BaseTransfer.sol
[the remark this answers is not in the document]
juniset: Yes, but I believe there will still be quite some gas consumed by invoking the wallet: internal call to the wallet, then delegatecall to the wallet implementation, then checking that the calling address is an authorised module, then calling the token contract which itself need to do a few check before updating the memory slot (even if this last step is free).

deployment/7_upgrade_1_6.js
juniset: It is the address of the first version of the TransferManager, yes. But here we are going from the second version (`config.modules.TransferManager`) to the third one. So the new TransferManager needs to read the state of a wallet's limit from the second version which is `config.modules.TransferManager`.

elenadimitrova: But we update the `config.modules.TransferManager` address to the new one in the same script.

contracts/modules/common/BaseTransfer.sol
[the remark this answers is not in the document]
elenadimitrova: It's around `12470` gas which is approx 1 cent worth at current prices. I can't tell whether the complexity outweighs the advantage but I'll add it.

deployment/7_upgrade_1_6.js
elenadimitrova: Okay, removed `TokenTransfer` in favour of using `TransferManager` value prior to upgrade updates.

PR #87, opened 2020-04-03
Consistent deployment gas price

juniset:
Making `gasPrice` consistent across all deployment scripts. The gas price can be set in the .env file as `DEPLOYER_GAS_PRICE`. If not present the value of 20 Gwei will be used by default.


deployment/7_upgrade_1_6.js
```diff
@@ -139,12 +140,12 @@ const deploy = async (network) => {
   // Disable KyberNetwork on the TokenPriceProvider
   // ////////////////////////////////////////////////////
 
-  if (!await TokenPriceProviderWrapper.contract.managers(deploymentWallet.address)) {
-    const TokenPriceProviderAddManagerTx = await TokenPriceProviderWrapper.contract.addManager(deploymentWallet.address);
-    await TokenPriceProviderWrapper.verboseWaitForTransaction(TokenPriceProviderAddManagerTx,
-      `Set ${deploymentWallet.address} as the manager of the TokenPriceProvider`);
-  }
-  const TokenPriceProviderSetKyberNetworkTx = await TokenPriceProviderWrapper.contract.setKyberNetwork("0x0000000000000000000000000000000000000000");
+  const TokenPriceProviderAddManagerTx = await TokenPriceProviderWrapper.contract.addManager(deploymentWallet.address, { gasPrice });
```
olivdb, line 143: I noticed that you left out my `if (!await TokenPriceProviderWrapper.contract.managers(deploymentWallet.address)) {` but I think it's ok. Not really needed.

utils/multisigexecutor.js
```diff
@@ -32,7 +33,7 @@ class MultisigExecutor {
       signature = ethers.utils.joinSignature(split);
 
       // Call "execute" on the Multisig wallet with data and signatures
-      const executeTransaction = await this._multisigWrapper.contract.execute(contractAddress, 0, data, signature, { gasLimit: 2000000 });
+      const executeTransaction = await this._multisigWrapper.contract.execute(contractAddress, 0, data, signature, this._overrides);
```
olivdb, line 36: There is a multisig transaction that failed (silently) on CI: https://app.circleci.com/pipelines/github/argentlabs/argent-contracts/451/workflows/5f21b4a6-2bfe-491f-8113-527c50ddf6f0/jobs/485

I wonder if this could be because you have reduced the gasLimit from 2M to 300k here.

deployment/7_upgrade_1_6.js
juniset: I think I dropped it by mistake when doing the rebase. I will add it again.

utils/multisigexecutor.js
juniset: Hmm, I did not see any fail transactions when running the deployment scripts, but I guess we set it to 2M for a reason. I will fix it.

utils/deploy-manager.js
```diff
@@ -7,9 +7,10 @@ const ConfiguratorLoader = require("./configurator-loader.js");
 const PrivateKeyLoader = require("./private-key-loader.js");
 const ABIUploader = require("./abi-uploader.js");
 const VersionUploader = require("./version-uploader.js");
+const utils = require("./utilities.js");
 
 const defaultConfigs = {
-  gasPrice: 20000000000, // 20 Gwei
+  gasPrice: utils.bigNumberify(process.env.DEPLOYER_GAS_PRICE || 20000000000),
```
elenadimitrova, line 13: Can this be in config rather than env vars? We have settings->deployer configuration in there already.

deployment/1_setup_test_environment.js
```diff
@@ -27,22 +28,23 @@ async function deployENSRegistry(deployer, owner, domain) {
...
 }
 
 // For development purpose
 async function deployKyber(deployer) {
+  const { gasPrice } = deployer.defaultOverrides;
   const KyberWrapper = await deployer.deploy(Kyber);
   const ERC20Wrapper = await deployer.deploy(ERC20, {}, [KyberWrapper.contractAddress], TEST_ERC20_SUPPLY, TEST_ERC20_DECIMALS);
 
-  const addToken = await KyberWrapper.contract.addToken(ERC20Wrapper.contractAddress, TEST_ERC20_RATE, TEST_ERC20_DECIMALS);
+  const addToken = await KyberWrapper.contract.addToken(ERC20Wrapper.contractAddress, TEST_ERC20_RATE, TEST_ERC20_DECIMALS, { gasPrice });
```
elenadimitrova, line 47: `Initializing UniswapFactory` step is also missing `gasPrice` here.

utils/deploy-manager.js
juniset: No because the value to use depends on the network status (typically the `fast` value on https://ethgasstation.info/) and we want to set it at the beginning of a deployment.

juniset:
> In general we should avoid using the `.contract` on etherlime contract wrappers as that points to the underlying ethers.js contract which for gasPrice handling differs in behaviour, in that it doesn't use the etherlime provider set default option. All `.contract` references in deployment scripts and otherwise can be replaced by direct calls on the wrapper, with the exception of encoding call data we use in some tests.
> 
> I am not in favour of passing `gasPrice` explicitly as it will be forgotten in some calls and we have no way of verifying it apart from trawling through the deployment test to watch for 2 GWei transactions.

I considered both approaches and decided to use the `contract` object of ethers.js because it gives us more flexibility. In particular the wrapper object of etherlime doesn't give us any control on the `gasLimit` used for a transaction and we've had failing transactions in deployment because the gas was not correctly estimated. This is the reason we are already using the contract object everywhere in the scripts.

By the way I explained these 2 options on the call last week and thought we had agreed to use the contract object instead of the wrapper. I probably should have explained the tradeoff better.

> This again raises the question of migrating to `truffle` which has a much clearer `gasPrice` handling than this.

If that's the case and we endup rewriting all the scripts for truffle anyway then maybe which option we choose (the contract or the wrapper) doesn't really matter as long as we are consistent.

deployment/7_upgrade_1_6.js
```diff
@@ -140,11 +141,13 @@ const deploy = async (network) => {
...
   if (!await TokenPriceProviderWrapper.contract.managers(deploymentWallet.address)) {
-    const TokenPriceProviderAddManagerTx = await TokenPriceProviderWrapper.contract.addManager(deploymentWallet.address);
+    const TokenPriceProviderAddManagerTx = await TokenPriceProviderWrapper.contract.addManager(deploymentWallet.address, { gasPrice });
     await TokenPriceProviderWrapper.verboseWaitForTransaction(TokenPriceProviderAddManagerTx,
       `Set ${deploymentWallet.address} as the manager of the TokenPriceProvider`);
   }
-  const TokenPriceProviderSetKyberNetworkTx = await TokenPriceProviderWrapper.contract.setKyberNetwork("0x0000000000000000000000000000000000000000");
+
+  const TokenPriceProviderSetKyberNetworkTx = await TokenPriceProviderWrapper.contract.setKyberNetwork("0x0000000000000000000000000000000000000000",
+    { gasPrice });
```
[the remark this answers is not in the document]
juniset: Yep, I forgot those 2 when rebasing my PR. Done.

deployment/1_setup_test_environment.js
juniset: Done.

PR #88, opened 2020-04-04
Adding unit tests for MakerV2

olivdb:
MakerV2Manager currently only has integration tests using Kovan MCD deployment. 
This PR adds ganache unit tests for both the loan and investment functionalities of MakerV2.

### DIFF
--- a/contracts/modules/maker/MakerV2Base.sol
+++ b/contracts/modules/maker/MakerV2Base.sol
@@ -18,7 +18,7 @@ import "../common/BaseModule.sol";
 import "../common/RelayerModule.sol";
 import "../common/OnlyOwnerModule.sol";
 import "../../../lib/utils/SafeMath.sol";
-import "../../../lib/maker/MakerV2Interfaces.sol";
+import "../../../lib/maker/MakerInterfaces.sol";
 import "../../infrastructure/MakerRegistry.sol";
 
 /**
--- a/contracts/modules/maker/MakerV2Base.sol
+++ b/contracts/modules/maker/MakerV2Base.sol
@@ -54,7 +54,7 @@ contract MakerV2Base is BaseModule, RelayerModule, OnlyOwnerModule {
     constructor(
         ModuleRegistry _registry,
         GuardianStorage _guardianStorage,
-        ScdMcdMigration _scdMcdMigration
+        ScdMcdMigrationLike _scdMcdMigration
     )
         BaseModule(_registry, _guardianStorage, NAME)
         public
--- a/contracts/modules/maker/MakerV2Base.sol
+++ b/contracts/modules/maker/MakerV2Base.sol
@@ -77,7 +77,7 @@ contract MakerV2Base is BaseModule, RelayerModule, OnlyOwnerModule {
         BaseWallet _wallet,
         uint256 _amount
     )
-        public
+        external
         onlyWalletOwner(_wallet)
         onlyWhenUnlocked(_wallet)
     {
--- a/test/integration/makerV2Manager_invest.js
+++ b/test/integration/makerV2Manager_invest.js
@@ -1,0 +1,217 @@
+// How to run this test file:
+// 1. add your private key (KOVAN_PRIV_KEY) and Infura API key (INFURA_KEY) to .env
+// --- TO TEST ON KOVAN: ----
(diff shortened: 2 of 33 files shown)


test/integration/makerV2Manager_invest.js
```diff
@@ -0,0 +1,217 @@
+// How to run this test file:
```
juniset, line 1: why do you have separated some tests in /integration?

test/makerV2Manager_invest.js
```diff
@@ -1,216 +1,115 @@
...
-    const { configurator } = deployManager;
-    deployer = deployManager.deployer;
-    testManager = new TestManager([...Array(10)].map(() => deployer), network, deployer);
-    owner = deployer.signer;
-    const { config } = configurator;
-
-    const makerRegistry = await deployer.deploy(MakerRegistry);
+    const m = await deployMaker(deployer, infrastructure);
+    [sai, dai] = [m.sai, m.dai];
+    const { migration, pot } = m;
```
juniset, line 29: why not do 
```
[sai, dai, migration, pot] = [m.sai, m.dai, m.migration, m.pot];
```
or 
```
const { sai, dai, migration, pot } = m;
```

utils/utilities.js
```diff
@@ -1,8 +1,28 @@
-const ethers = require("ethers");
 const readline = require("readline");
+const ethers = require("ethers");
+
+const { bigNumberify } = ethers.utils;
+
+const ETH_TOKEN = "0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE";
+const RAY = bigNumberify("1000000000000000000000000000"); // 10**27
```
juniset, line 7: where are these constants used? Seems like they are only used in defi_deployer.js where they are also locally defined.

contracts/modules/maker/MakerV2Base.sol
```diff
@@ -77,7 +77,7 @@ contract MakerV2Base is BaseModule, RelayerModule, OnlyOwnerModule {
         BaseWallet _wallet,
         uint256 _amount
     )
-        public
+        external
```
juniset, line 80: why did you make that change? Isn't it better for the constructor of a base contract to be `public` so it can be called in the constructor of a contract that extends it?

test/integration/makerV2Manager_invest.js
olivdb: These are the previous Kovan tests. We could just delete them but since I may have cut some corners when deploying and setting up MCD in my ganache tests, there may be some value to keep them around as an additional set of tests that touch a real world MCD deployment. Open to both.

test/makerV2Manager_invest.js
olivdb: I like to keep variables as local as they can be. `pot` and `migration` are only used in `before()` so they can be declared locally and destructured with `{}` but `sai` and `dai` are used in multiple functions so they need to be declared out of the scope of these functions and can therefore not be destructured with `{}` (you can only use variable destructuring at the point of declaration).

utils/utilities.js
olivdb: They are imported by `makerV2Manager_loan.js`, `makerV2Manager_invest.js` and `makerManager_loan.js`  with the exception of the `USD_PER_xxx` variables which are not  actually imported by any file (my bad). I think it would actually make more sense for these variables to be exported from `defi-deployer.js` since they are only used by maker test scripts. I will just leave `ETH_TOKEN` in `utilities.js` and make all test files that define `ETH_TOKEN` import it from `utilities.js` instead.

contracts/modules/maker/MakerV2Base.sol
olivdb: Yes absolutely. Not sure where this change came from and why I don't see it in my local repo. Weird.

olivdb: Oh I get it, the change is not on the constructor, it's on `swapSaiToDai()` which didn't need to be made public any more after we decided not to auto-convert SAI to DAI for debt repayment.

test/integration/makerV2Manager_invest.js
juniset: Yes, let's keep them.

PR #93, opened 2020-04-13
Add static contract analysis to CI

elenadimitrova:
Enables static analysis on the contracts using `slither` https://github.com/crytic/slither
This is run by CI in a parallel python-based container on all builds now. Some valid suggested fixes were applied the rest were ignored in the `slither.db.json`. Two new npm commands added
`security:slither` and `security:slither:triage`, the former runs slither analyzer on the contracts and the latter runs the same but in triage mode allowing us to ignore errors that are either false positives or we don't consider a threat. 

- Includes extracting all test contracts out in a `/contracts-test` folder due to overlap in contract names between dappsys and openzeppelin contracts we use in unit testing, namely `IERC20` which trips up `slither` atm.

- Removes duplication of `IUniswapExchange` and `IUniswapFactory` contract definitions which otherwise causes `slither` to error . 

Found and logged a couple of problems with the latest `slither release 0.6.11` https://github.com/crytic/slither/issues/456 and https://github.com/crytic/slither/issues/457. Those were resolved in `slither release 0.6.12`


contracts/infrastructure/ens/ArgentENSManager.sol
```diff
@@ -125,7 +125,7 @@ contract ArgentENSManager is IENSManager, Owned, Managed {
      * @param _subnode The target subnode.
      * @return true if the subnode is available.
      */
-    function isAvailable(bytes32 _subnode) public view returns (bool) {
+    function isAvailable(bytes32 _subnode) external view returns (bool) {
```
juniset, line 128: Is this change necessary? I thought the idea was to not touch the infrastructure contracts for the next release.

contracts/infrastructure/ens/ArgentENSResolver.sol
```diff
@@ -46,7 +46,7 @@ contract ArgentENSResolver is Owned, Managed, ENSResolver {
      * @param _node The node to update.
      * @param _addr The address to set.
      */
-    function setAddr(bytes32 _node, address _addr) public onlyManager {
+    function setAddr(bytes32 _node, address _addr) external onlyManager {
```
juniset, line 49: Same comment as above. Are these changes necessary for Slither or is just cleaning?

contracts/modules/RecoveryManager.sol
```diff
@@ -132,10 +129,13 @@ contract RecoveryManager is BaseModule, RelayerModule {
...
-        _wallet.setOwner(config.recovery);
-        emit RecoveryFinalized(address(_wallet), config.recovery);
-        guardianStorage.setLock(_wallet, 0);
+        address recoveryOwner = config.recovery;
         delete recoveryConfigs[address(_wallet)];
+
+        _wallet.setOwner(recoveryOwner);
+        guardianStorage.setLock(_wallet, 0);
+
+        emit RecoveryFinalized(address(_wallet), config.recovery);
```
juniset, line 138: It should be `recoveryOwner` and not `config.recovery`.

contracts/modules/RecoveryManager.sol
```diff
@@ -145,9 +145,10 @@ contract RecoveryManager is BaseModule, RelayerModule {
      */
     function cancelRecovery(BaseWallet _wallet) external onlyExecute onlyWhenRecovery(_wallet) {
         RecoveryConfig storage config = recoveryConfigs[address(_wallet)];
-        emit RecoveryCanceled(address(_wallet), config.recovery);
-        guardianStorage.setLock(_wallet, 0);
         delete recoveryConfigs[address(_wallet)];
+        guardianStorage.setLock(_wallet, 0);
+
+        emit RecoveryCanceled(address(_wallet), config.recovery);
```
juniset, line 151: Since we deleted the storage slot at position `recoveryConfigs[address(_wallet)]`, shouldn't `config.recovery` always return `0` ?

elenadimitrova: Yes you're right.

contracts/infrastructure/ens/ArgentENSManager.sol
elenadimitrova: These were raised by `slither` on the ENS contracts for [this rule](https://github.com/crytic/slither/wiki/Detector-Documentation#public-function-that-could-be-declared-as-external). They are not dangerous so have ignored them for now but good to fix in future.

contracts/modules/RecoveryManager.sol
```diff
@@ -145,9 +145,11 @@ contract RecoveryManager is BaseModule, RelayerModule {
      */
     function cancelRecovery(BaseWallet _wallet) external onlyExecute onlyWhenRecovery(_wallet) {
         RecoveryConfig storage config = recoveryConfigs[address(_wallet)];
-        emit RecoveryCanceled(address(_wallet), config.recovery);
-        guardianStorage.setLock(_wallet, 0);
+        address recoveryOwner = config.recovery;
```
juniset, line 148: Since `config` is only used to define `recoveryOwner` it is probably more efficient to do
```
address recoveryOwner = recoveryConfigs[address(_wallet)].recovery;
```

PR #97, opened 2020-04-23
Event log failures in RelayerModule execute calls

elenadimitrova:
- Log failures of sub-call in `Relayer.execute` function to ensure returned revert messages are captured. In the cases when the `refund` call has failed the message logged is `RM: refund failed`

- Rebase all modules to `RelayerModuleV2` which was renamed back to `RelayerModule`.

- Extend the `getRequiredSignatures` method on relayer to return extra information on owner signature requirements needed to encapsulate `validateSignatures` logic in the relayer only.

- Add the benchmark script to CI to improve visibility on gas costs in introduced changes.

We attempted to refactor the `OnlyOwnerModule` logic into the `RelayerModule` for a cleaner inheritance model however decided against it in the end. Here is a summary of that analysis for posterity.
The `OnlyOwnerModule` contract main purpose is to enable relayed `addModule` calls. There were two options there:
1) add logic to the `RelayerModule.execute` call directly, thus decomposing every call data bytes to read whether the current call is to `addModule` 
2) selectively add that logic in the modules that require different signatures. 
3) same as option 1 but also change the signature of `getRequiredSignatures` to take only the method signature instead of the deconstructed call data 
Comparison table of gas costs for the three options is below:

| Task                                          | Original gas spent | Gas change 1 | Diff change 1 | Gas change 2 | Diff change 2 | Gas change 3 | Diff change 3 |
|-----------------------------------------------|--------------------|--------------|---------------|--------------|---------------|--------------|---------------|
| Add a guardian (relayed)                      | 155100             | 155505       | 405           | 155124       | 24            | 155270       | 170           |
| Revoke a guardian (relayed)                   | 81409              | 81785        | 376           | 81392        | -17           | 81550        | 141           |
| Lock wallet (relayed)                         | 131944             | 132342       | 398           | 132227       | 283           | 132186       | 242           |
| Unlock wallet (relayed)                       | 81981              | 82384        | 403           | 82235        | 254           | 82211        | 230           |
| Execute recovery                              | 158301             | 158864       | 563           | 158411       | 110           | 158615       | 314           |
| Change limit (relayed)                        | 97835              | 98260        | 425           | 97778        | -57           | 98109        | 274           |
| ETH transfer, limit disabled (relayed)        | 93986              | 94409        | 423           | 93943        | -43           | 94245        | 259           |
| ETH small transfer (relayed)                  | 117083             | 117535       | 452           | 117052       | -31           | 117359       | 276           |
| ETH transfer to whitelisted account (relayed) | 91346              | 91757        | 411           | 91279        | -67           | 91586        | 240           |
| ETH large transfer approval by one guardian   | 106209             | 106673       | 464           | 106415       | 206           | 106485       | 276           |
| ETH large transfer approval by two guardians  | 119161             | 119534       | 373           | 119340       | 179           | 119362       | 201           |


test/approvedTransfer.js
```diff
@@ -262,8 +263,9 @@ describe("Approved Transfer", function () {
           const txReceipt = await manager.relay(approvedTransfer, "approveTokenAndCallContract",
             [wallet.contractAddress, erc20.contractAddress, wallet.contractAddress, amountToApprove, target.contractAddress, invalidData],
             wallet, [owner, ...sortWalletByAddress([guardian1, guardian2])]);
-          const success = parseRelayReceipt(txReceipt);
-          assert.isNotOk(success, "approveTokenAndCall should fail when target contract is invalid");
+          const { success, error } = parseRelayReceipt(txReceipt);
+          assert.isFalse(success);
+          assert.equal(error, "AT: Forbidden contract");
```
olivdb, line 268: Could we instead use the same short syntax as you used in `lockManager.js` ?
```
await assert.revertWith(manager.relay(...), "AT: Forbidden contract");
```

test/relayer.js
```diff
@@ -1,19 +1,20 @@
...
 const Wallet = require("../build/BaseWallet");
-const TestModule = require("../build/TestModule");
-const TestModuleRelayerV2 = require("../build/TestModuleRelayerV2");
+const TestModule = require("../build/TestModuleRelayer");
 const Registry = require("../build/ModuleRegistry");
 const GuardianStorage = require("../build/GuardianStorage");
 const ApprovedTransfer = require("../build/ApprovedTransfer");
-const RecoveryManager = require("../build/RecoveryManager");
+const RecoveryManager = require("../build/RecoveryManager"); // non-owner only module
+const NftTransfer = require("../build/NftTransfer"); // owner only module
```
olivdb, line 11: We have `TestOnlyOwnerModule.sol` that could be used here instead

test/simpleUpgrader.js
```diff
@@ -97,7 +97,7 @@ describe("SimpleUpgrader", function () {
       // create module V2
       const moduleV2 = await deployer.deploy(Module, {}, registry.contractAddress, false, 0);
       // register module V2
-      await registry.registerModule(moduleV2.contractAddress, formatBytes32String("V2"));
+      // await registry.registerModule(moduleV2.contractAddress, formatBytes32String("V2"));
```
olivdb, line 100: Did you intentionally leave this line commented out or did you want to delete it?

contracts/test/TestModuleRelayer.sol
```diff
@@ -4,13 +4,13 @@ import "../modules/common/BaseModule.sol";
 import "../modules/common/RelayerModule.sol";
 
 /**
- * @title TestModule
- * @dev Basic test module.
- * @author Julien Niset - <julien@argent.xyz>
+ * @title TestModuleRelayer
```
olivdb, line 7: `TestModuleRelayerV2` was named that way to distinguish it from `TestModule` which was based on `RelayerModule` (V1). Now that we have only one version of the `RelayerModule`, we could just have the present module be named `TestModule` instead of `TestModuleRelayer`.

test/relayer.js
elenadimitrova: I intentionally removed it to test behaviour with an actual OnlyOwnerModule we use. This could be a separate discussion altogether but I feel we overuse the TestContracts approach. Having a dedicated test contract is needed when it's a test case we can't achieve with the real-life contracts. Thus we keep test coverage reports up to date as well.

test/simpleUpgrader.js
elenadimitrova: This was intentionally done in [this commit](https://github.com/argentlabs/argent-contracts/pull/97/commits/92d2573ffafca0a1636a14201e0ac4233f44fbbd).

contracts/test/TestModuleRelayer.sol
elenadimitrova: Makes sense.

contracts/modules/RecoveryManager.sol
```diff
@@ -192,21 +192,25 @@ contract RecoveryManager is BaseModule, RelayerModule {
...
+    function getRequiredSignatures(BaseWallet _wallet, bytes memory _data) internal view returns (uint256, OwnerSignature) {
         bytes4 methodId = functionPrefix(_data);
         if (methodId == EXECUTE_RECOVERY_PREFIX) {
-            return SafeMath.ceil(guardianStorage.guardianCount(_wallet), 2);
+            uint numberOfSignaturesRequired = SafeMath.ceil(guardianStorage.guardianCount(_wallet), 2);
+            return (numberOfSignaturesRequired, OwnerSignature.Disallowed);
         }
         if (methodId == FINALIZE_RECOVERY_PREFIX) {
-            return 0;
+            return (0, OwnerSignature.Disallowed);
```
juniset, line 184: This is a detail, but shouldn't it be cheaper to return the Enum value corresponding to 0, i.e. `OwnerSignature.Required`?

contracts/modules/common/RelayerModule.sol
```diff
@@ -105,7 +105,8 @@ contract RelayerModule is BaseModule {
         bytes32 signHash = getSignHash(address(this), address(_wallet), 0, _data, _nonce, _gasPrice, _gasLimit);
         require(checkAndUpdateUniqueness(_wallet, _nonce, signHash), "RM: Duplicate request");
         require(verifyData(address(_wallet), _data), "RM: Target of _data != _wallet");
-        uint256 requiredSignatures = getRequiredSignatures(_wallet, _data);
+        uint256 requiredSignatures;
+        (requiredSignatures,) = getRequiredSignatures(_wallet, _data);
```
juniset: This can be written as `(uint256 requiredSignatures,) = getRequiredSignatures(_wallet, _data);` which I personally find more readable.

contracts/modules/common/RelayerModule.sol
```diff
@@ -106,9 +89,11 @@ contract RelayerModule is BaseModule {
         require(checkAndUpdateUniqueness(_wallet, _nonce, signHash), "RM: Duplicate request");
         require(verifyData(address(_wallet), _data), "RM: Target of _data != _wallet");
         uint256 requiredSignatures;
-        (requiredSignatures,) = getRequiredSignatures(_wallet, _data);
+        OwnerSignature ownerSignatureRequirement;
+        (requiredSignatures, ownerSignatureRequirement) = getRequiredSignatures(_wallet, _data);
```
juniset, line 93: This can be written as ```(uint256 requiredSignatures, OwnerSignature  ownerSignatureRequirement) = getRequiredSignatures(_wallet, _data);``` which I personally find more readable.

contracts/modules/common/RelayerModule.sol
```diff
@@ -52,23 +60,9 @@ contract RelayerModule is BaseModule {
     * @param _data The data of the relayed transaction.
     * @return The number of required signatures.
     */
-    function getRequiredSignatures(BaseWallet _wallet, bytes memory _data) internal view returns (uint256);
```
juniset, line 55: The @return description should be updated to mention the OwnerSignature flag.

contracts/modules/RecoveryManager.sol
elenadimitrova: I tested this with the following tests but gas cost of each didn't change
`should let anyone finalize the recovery procedure after the recovery period (267ms, 57481 gas)`
`should not let anyone finalize the recovery procedure before the end of the recovery period (207ms, 56306 gas)`
I suppose the memory for the return type is reserved and same whether or not they are zero values.

contracts/modules/common/RelayerModule.sol
```diff
@@ -93,18 +87,21 @@ contract RelayerModule is BaseModule {
...
-                }
-            }
+        require(verifyData(address(_wallet), _data), "RM: Target of _data != _wallet");
+        (uint256 requiredSignatures, OwnerSignature ownerSignatureRequirement) = getRequiredSignatures(_wallet, _data);
+        require(requiredSignatures * 65 == _signatures.length, "RM: Wrong number of signatures");
+        require(requiredSignatures == 0 || validateSignatures(_wallet, signHash, _signatures, ownerSignatureRequirement),
+         "RM: Invalid signatures");
+        // The correctness of the refund is checked on the next line using an `if` instead of a `require`
+        // in order to prevent a failing refund from being replayable in the future.
+        bytes memory returnData;
```
juniset, line 97: Shouldn't we provide an error string to `returnData` when the following IF is not met? Potentially an ELSE with something like "cannot refund" otherwise we may have an error string for all scenarios but that one.

contracts/modules/common/RelayerModule.sol
```diff
@@ -99,6 +99,9 @@ contract RelayerModule is BaseModule {
             // solium-disable-next-line security/no-call-value
             (success, returnData) = address(this).call(_data);
             refund(_wallet, startGas - gasleft(), _gasPrice, _gasLimit, requiredSignatures, msg.sender);
+        } else {
+            success = false;
```
juniset, line 103: I believe this assignment is not necessary as the default value of `success` is already `false`.

PR #100, opened 2020-04-30
Consistent SafeMath library use

elenadimitrova:
We previously relied on a mix of Maker and OpenZeppelin SafeMath library implementations that we merged into our own `SafeMath` contract. This PR refactors those usages to reference imported external contracts directly rather than our copy of them. We've kept the `ceil` function implementation in own `ArgentSafeMath` library contract as that function is not available in the other two implementations we use. 

Here we mandate the following usage:
Generic SafeMath functions use `openzeppelin-solidity/contracts/math/SafeMath.sol`
Maker logic relying on RAY and WAD math use `DSMath`
`ceil` logic to be used from `modules/common/ArgentSafeMath.sol` 

Additionally `CarefulMath` used by Compound is kept as a separate contract as defined in original repo and imported as-is in `/lib`.


contracts/modules/common/ArgentSafeMath.sol
```diff
@@ -0,0 +1,42 @@
+/* The MIT License (MIT)
+
+Copyright (c) 2016 Smart Contract Solutions, Inc.
```
olivdb, line 3: 2020, Argent Labs?

contracts/legacy/LegacyMakerManager.sol
```diff
@@ -554,11 +555,13 @@ contract LegacyMakerManager is Loan, BaseModule, RelayerModule, OnlyOwnerModule
...
-            .rmul(_makerCdp.mat())               // x 1.5
-            .rmul(1010000000000000000000000000) // x (1+1%) cushion
-            .rdiv(_makerCdp.tag());              // ÷ ~170 USD/PETH
+        _minCollateral = rdiv(
+            rmul(
+                rmul(
+                    rmul(daiDebt(_cup, _makerCdp), _makerCdp.vox().par()), // DAI debt x ~1 USD/DAI
+                    _makerCdp.mat()),           // x 1.5
+                1010000000000000000000000000),  // x (1+1%) cushion
+            _makerCdp.tag());                  // ÷ ~170 USD/PETH
```
olivdb, line 564: What's the reason for avoiding the dot notation here and in `MakerV2Loan`? I find it less readable.

contracts/modules/common/ArgentSafeMath.sol
elenadimitrova: Sorry, yes this was a leftover from the old `SafeMath` lib we had.

contracts/legacy/LegacyMakerManager.sol
elenadimitrova: Reason being is that while openzeppelin's `SafeMath` is a library, Maker's `DSMath` is a contract. Additionally the entire Maker solution uses this notation.

PR #107, opened 2020-05-06
Generate contract api definitions

elenadimitrova:
cc @petar-argent


elenadimitrova:
Handed over to @tms1337 to incorporate in the dashboard.

PR #108, opened 2020-05-06
Remove obsolete Compound imports

elenadimitrova:
Removes `Invest` and `Loan` contract dependencies from the Compound module. 

Here we also move the `/base` contracts folder under `/infrastructure` folder as they're only relevant and used there.


contracts/modules/CompoundManager.sol
```diff
@@ -87,184 +75,6 @@ contract CompoundManager is BaseModule, RelayerModule, OnlyOwnerModule {
...
-        require(error == 0, "Compound: failed to get account liquidity");
-        if (liquidity > 0) {
-            return (1, liquidity);
-        }
-        if (shortfall > 0) {
-            return (2, shortfall);
-        }
-        return (0,0);
-    }
-
```
juniset, line 254: Why are you removing all the Loan logic from CompoundManager? While it is not implemented in the clients yet it should still be on the roadmap. Since this code has been audited and we are still planning to use it at some point I don't see why we should remove it from the contract now.

juniset:
The `Owned` and `Managed` imports need to be updated in the ENS contracts (ArgentENsManager and ArgentENS Resolver).

PR #95, opened 2020-04-16
Remove Kyber price syncing from TokenPriceProvider

elenadimitrova:
does what it says on the tin

PR #109, opened 2020-05-06
Revert "Remove Kyber price syncing from TokenPriceProvider"

elenadimitrova:
Reverts argentlabs/argent-contracts#95


deployment/7_upgrade_1_6.js
```diff
@@ -157,6 +157,21 @@ const deploy = async (network) => {
...
+    const TokenPriceProviderAddManagerTx = await TokenPriceProviderWrapper.contract.addManager(deploymentWallet.address, { gasPrice });
+    await TokenPriceProviderWrapper.verboseWaitForTransaction(TokenPriceProviderAddManagerTx,
+      `Set ${deploymentWallet.address} as the manager of the TokenPriceProvider`);
+  }
+
+  const TokenPriceProviderSetKyberNetworkTx = await TokenPriceProviderWrapper.contract.setKyberNetwork("0x0000000000000000000000000000000000000000",
+    { gasPrice });
+  await TokenPriceProviderWrapper.verboseWaitForTransaction(TokenPriceProviderSetKyberNetworkTx,
+    "Disable the KyberNetwork on the TokenPriceProvider");
+
```
juniset, line 174: This was executed on the previous deployment and is no longer required. The address of kyber network is already set to `address(0)`.

Also, the backend addresses are already managers and the multisig is already the owner of the contract. Basically the next section (Change the owner of TokenPriceProvider) is no longer required either.

deployment/7_upgrade_1_6.js
```diff
@@ -157,35 +154,6 @@ const deploy = async (network) => {
       [wrapper.contractAddress, utils.asciiToBytes32(wrapper._contract.contractName)]);
   }
 
-  // ////////////////////////////////////////////////////
```
juniset, line 160: we should not upload the ABI of the TokenPriceProvider (L144)

PR #112, opened 2020-05-13
Code coverage improvements

elenadimitrova:
Includes the Maker modules in code coverage. Also adds missing tests for storage contracts.


test/guardianManager.js
```diff
@@ -324,6 +324,21 @@ describe("GuardianManager", function () {
       count = (await guardianStorage.guardianCount(wallet.contractAddress)).toNumber();
       assert.equal(count, 2, "there should be 2 guardians again");
     });
+
+    it("should be able to remove a guardian that is not last in the list", async () => {
```
olivdb, line 328: This says "_not_ last in the list" but the guardian revoked is `guardians[2]` which is the last in the list.

test/makerV2Manager_invest.js
```diff
@@ -21,22 +24,33 @@ describe("MakerV2 DSR", function () {
...
   let sai;
   let dai;
+  let jug;
+  let vat;
+  let gov;
 
   before(async () => {
     const m = await deployMaker(deployer, infrastructure);
-    [sai, dai] = [m.sai, m.dai];
+    [sai, dai, jug, vat, gov] = [m.sai, m.dai, m.jug, m.vat, m.gov];
```
olivdb, line 33: `jug`, `vat and `gov` are not used anywhere outside this block so they could instead be destructured on the next line along with `migration` and `pot`.

test/guardianManager.js
elenadimitrova: Yes indeed I am testing this `if` condition logic https://github.com/argentlabs/argent-contracts/blob/develop/contracts/modules/storage/GuardianStorage.sol#L71 . Renamed.

PR #114, opened 2020-05-18
Upgrade modules and wallet contracts to solidity 0.6.10

elenadimitrova:
Upgrades contracts from `/wallet` and `/modules` to Solidity 0.6.10 while `/infrastructure` contracts remain on 0.5.4 except for the `WalletFactory` and `TokenPriceProvider` which are also upgraded to 0.6.10 (for that purpose I've had to move them out of /infrastructure to be compiled with the rest of the upgraded contracts until https://github.com/LimeChain/etherlime/issues/325 is fixed. 
Overlapping areas of logic were abstracted into interfaces with wider range compilation pragma i.e. `pragma solidity >=0.5.4 <0.7.0;` due to the fact there are no breaking changes to interfaces between the two major releases.

Includes refactoring that pulls apart the dependencies between storage contracts, `BaseWallet` and `Module` in order to allow storage contracts to remain on solc 0.5.4. Interactions between these layers is done via the `IModule`, `IWallet`, `IGuardianStorage` and `ITransferStorage` interfaces.

Also upgraded the [openzeppelin-contracts library to 3.0.1](https://github.com/OpenZeppelin/openzeppelin-contracts/releases/tag/v3.0.1)

The following changes in gas consumption for benchmark tests is observed between releases:
| Task [1 ETH = 500 USD]                        | Gas 0.5.4 | Gas 0.6.9 | Diff   |
|-----------------------------------------------|-----------|-----------|--------|
| Create a wallet without ENS (1 module)        | 486247    | 479970    | -6277  |
| Create a wallet without ENS (2 module)        | 515802    | 508848    | -6954  |
| Create a wallet without ENS (3 module)        | 545376    | 537746    | -7630  |
| Create a wallet without ENS (all modules)     | 832433    | 819941    | -12492 |
| Create a wallet with ENS (all modules)        | 832517    | 820025    | -12492 |
| Add first guardian (direct)                   | 119451    | 119462    | 11     |
| Request add second guardian (direct)          | 74359     | 73473     | -886   |
| Confirm add second guardian (direct)          | 92939     | 93452     | 513    |
| Request revoke guardian (direct)              | 60275     | 59840     | -435   |
| Confirm revoke guardian (direct)              | 89758     | 89565     | -193   |
| Add a guardian (relayed)                      | 154175    | 153712    | -463   |
| Revoke a guardian (relayed)                   | 80909     | 80115     | -794   |
| Lock wallet (direct)                          | 86299     | 86118     | -181   |
| Unlock wallet (direct)                        | 51322     | 51028     | -294   |
| Lock wallet (relayed)                         | 130832    | 130210    | -622   |
| Unlock wallet (relayed)                       | 80867     | 80137     | -730   |
| Execute recovery                              | 156766    | 155862    | -904   |
| Change limit (direct)                         | 61931     | 61438     | -493   |
| Change limit (relayed)                        | 97260     | 96337     | -923   |
| ETH transfer, limit disabled (direct)         | 57422     | 56340     | -1082  |
| ETH transfer, limit disabled (relayed)        | 93189     | 91689     | -1500  |
| ETH small transfer (direct)                   | 66884     | 65748     | -1136  |
| ETH small transfer (relayed)                  | 116025    | 114453    | -1572  |
| ETH transfer to whitelisted account (direct)  | 55073     | 54057     | -1016  |
| ETH transfer to whitelisted account (relayed) | 90864     | 89401     | -1463  |
| ETH large transfer to untrusted account       | 80270     | 79116     | -1154  |
| ETH large transfer approval by one guardian   | 105440    | 104010    | -1430  |
| ETH large transfer approval by two guardians  | 118092    | 116390    | -1702  |

Additionally we've removed the multiple inheritance in modules, e.g. `contract NftTransfer is BaseModule, RelayerModule, OnlyOwnerModule` -> `contract NftTransfer is OnlyOwnerModule` as the former causes the following compile errors caused by `BaseModule`, `RelayerModule`, `OnlyOwnerModule` all implementing versions of `addModule` and `checkAndUpdateUniqueness` methods.

```
/argent-contracts/contracts/modules/NftTransfer.sol:26:1: TypeError: Derived contract must override function "addModule". Two or more base classes define function with same name and parameter types.
contract NftTransfer is BaseModule, RelayerModule, OnlyOwnerModule {
^ (Relevant source part starts here and spans across multiple lines).
/Users/Elena/Source/argent-contracts/contracts/modules/common/BaseModule.sol:97:5: Definition in "BaseModule":
    function addModule(BaseWallet _wallet, Module _module) external virtual override strictOnlyWalletOwner(_wallet) {
    ^ (Relevant source part starts here and spans across multiple lines).
/Users/Elena/Source/argent-contracts/contracts/modules/common/OnlyOwnerModule.sol:48:5: Definition in "OnlyOwnerModule":
    function addModule(BaseWallet _wallet, Module _module) external override virtual onlyWalletOwner(_wallet) {
    ^ (Relevant source part starts here and spans across multiple lines).
,/Users/Elena/Source/argent-contracts/contracts/modules/NftTransfer.sol:26:1: TypeError: Derived contract must override function "checkAndUpdateUniqueness". Two or more base classes define function with same name and parameter types.
contract NftTransfer is BaseModule, RelayerModule, OnlyOwnerModule {
^ (Relevant source part starts here and spans across multiple lines).
/Users/Elena/Source/argent-contracts/contracts/modules/common/OnlyOwnerModule.sol:56:5: Definition in "OnlyOwnerModule":
    function checkAndUpdateUniqueness(BaseWallet _wallet, uint256 _nonce, bytes32 /* _signHash */) internal override returns (bool) {
    ^ (Relevant source part starts here and spans across multiple lines).
/Users/Elena/Source/argent-contracts/contracts/modules/common/RelayerModule.sol:157:5: Definition in "RelayerModule":
    function checkAndUpdateUniqueness(BaseWallet _wallet, uint256 _nonce, bytes32 _signHash) internal virtual returns (bool) {
    ^ (Relevant source part starts here and spans across multiple lines).
```

Note that due to https://github.com/crytic/slither/issues/405 we've had to disable `slither` analysis on the wallet contracts. Logged this in trello to remind us to reenable it once `slither` supports analysis of the new 0.6. syntax.


package.json
```diff
@@ -57,10 +61,10 @@
     "inquirer": "^7.0.0",
     "istanbul": "^0.4.5",
     "node-fetch": "^2.3.0",
-    "openzeppelin-solidity": "2.3.0",
+    "openzeppelin-solidity": "3.0.1",
```
olivdb, line 64: Though both `openzeppelin-solidity` and `@openzeppelin/contracts` npm repositories are still being updated, `@openzeppelin/contracts` seems too be the preferred repo now (it's the one mentioned on their Github page).

contracts-test/LegacyBaseWallet.sol
```diff
@@ -24,7 +24,7 @@ pragma solidity ^0.5.4;
 
     /**
     * @dev Utility method to recover any ERC20 token that was sent to the
-    * module by mistake. 
+    * module by mistake.
```
olivdb, line 27: Not related to this PR but this should say "contract" not "module"

contracts-test/TestERC20.sol
```diff
@@ -1,32 +1,24 @@
...
-    uint8 public decimals;
-
+contract TestERC20 is ERC20("ArgentToken", "AGT") {
     constructor (address[] memory _initialAccounts, uint _supply, uint8 _decimals) public {
-        decimals = _decimals;
-        for(uint i = 0; i < _initialAccounts.length; i++) {
-            _mint(_initialAccounts[i], _supply * 10**uint(_decimals));
+        super._setupDecimals(_decimals);
+        for (uint i = 0; i < _initialAccounts.length; i++) {
+            super._mint(_initialAccounts[i], _supply * 10**uint(_decimals));
```
olivdb, line 13: Is `super.` required here? You didn't use it in `TestERC721.mint()`

olivdb:
Perhaps we should prefix all contracts in `contracts/legacy` with `LegacyXXX`. Otherwise wouldn't the JSON artifacts in `build/` produced by `npm run compile:legacy` clash with those produced by `npm run compile:module` ?

package.json
elenadimitrova: I had no idea about the new one. Will update.

contracts/wallet/Proxy.sol
```diff
@@ -31,21 +32,20 @@ contract Proxy {
...
+            let result := delegatecall(gas(), target, 0, calldatasize(), 0, 0)
+            returndatacopy(0, 0, returndatasize())
+            switch result
+            case 0 {revert(0, returndatasize())}
+            default {return (0, returndatasize())}
         }
     }
+
+    receive() external payable {
+        emit Received(msg.value, msg.sender, msg.data);
```
olivdb, line 49: Probably inconsequential but maybe worth notifying the clients that whereas before `Received` could not have been emitted with both `msg.value == 0` and `msg.data == 0`, it is now possible.

contracts/wallet/BaseWallet.sol
```diff
@@ -117,42 +119,43 @@ contract BaseWallet {
...
+                returndatacopy(0, 0, returndatasize())
+                switch result
+                case 0 {revert(0, returndatasize())}
+                default {return (0, returndatasize())}
             }
         }
     }
+
+    receive() external payable {
+        emit Received(msg.value, msg.sender, msg.data);
```
olivdb, line 159: I think we can remove this event emission and leave `receive()` empty. If I'm not mistaken `receive()` here will only ever be called in our tests.

contracts/wallet/Proxy.sol
elenadimitrova: I couldn't work out what's right for this event. [Here](https://github.com/argentlabs/argent-contracts/blob/fd1222b422cc2a1ad1c9e9518fe4f14717e9ffcf/contracts/wallet/BaseWallet.sol#L143) it is emitted for 0 values while in the Proxy it was not.

olivdb: Yes the previous implementation was confusing. I think it would have made more sense to check `if (msg.data.length == 0) { emit ... }` in the Proxy instead of `if (msg.data.length == 0 && msg.value > 0) { emit ... }`. This is equivalent to your new implementation, which I think is good.

juniset:
Shouldn't we move the /legacy folder out of /contracts (e.g. in contracts-legacy)? I believe we should only have in /contracts what would be part of a security audit.

elenadimitrova:
@olivdb @juniset regarding your comments on the `legacy` contracts, you're both right and I've pushed a change to both bring out the contracts into own folder at /contracts folder level as well as compile them in their own build-legacy folder to separate build artefacts we use in tests.

olivdb:
I would remove the `Legacy` prefix from `LegacyTransferManager` and `LegacyMakerManager`. I would also move remove the `Legacy` prefix from `LegacyBaseWallet` and move the file to `contracts-legacy`. But since there is a more recent "legacy" `BaseWallet` in there, I think we should place the contracts in `contracts-legacy` in subdirectories according to the last release in which they appeared. That would mean that all the legacy contracts would be placed under `contract-legacy/v1.6.0` except `BaseWallet.sol` which should probably go in `contract-legacy/v1.3.0`

contracts/wallet/BaseWallet.sol
elenadimitrova: I've made `Receive` only emit on non-zero values https://github.com/argentlabs/argent-contracts/pull/114/commits/12c79ec27dba5802f5665672cc11d0786b186f28

contracts/modules/TransferManager.sol
```diff
@@ -130,7 +130,6 @@ contract TransferManager is OnlyOwnerModule, BaseTransfer, LimitManager {
         }
     }
 
-    // TODO: We should inherit the OnlyOwnerModule implementation
```
olivdb: You probably tried the same things but I would have expected that we could have made `TransferManager.addModule()` call `OnlyOwnerModule.addModule(_wallet, _module);` instead of duplicating its code. Or alternatively, make `OnlyOwnerModule` be the last contract inherited by `TransferManager` (instead of `LimitManager`) and use `super.addModule(_wallet, _module);` in `TransferManager.addModule()`. But none of these solutions seemed to work for me. Is that a solidity bug or am I missing something?

Also, if we're going to duplicate `OnlyOwnerModule`'s code, let's change "BM: module is not registered" to "TM: module is not registered".

contracts/wallet/BaseWallet.sol
```diff
@@ -117,42 +120,45 @@ contract BaseWallet {
...
+                let result := staticcall(gas(), module, 0, calldatasize(), 0, 0)
+                returndatacopy(0, 0, returndatasize())
+                switch result
+                case 0 {revert(0, returndatasize())}
+                default {return (0, returndatasize())}
             }
         }
     }
+
+    receive() external payable {
```
olivdb, line 159: This function is not needed and can be removed entirely. It will never be reached from the Proxy (since calls to the Proxy with `msg.data == 0` are not forwarded by the Proxy).

juniset:
- The tests files that use legacy contracts (e.g. baseWallet.js) must be updated as they are still looking for legacy contracts in `/build` .

- The deployment script `5_deploy_modules.js` does not use the newly deployed GuardianStorage and TransferStorage contracts to instantiate some of the modules.

contracts/modules/TransferManager.sol
elenadimitrova: It's because `addModule` is `external`, see Chris's explanation in this discussion https://github.com/ethereum/solidity/issues/3881#issuecomment-579456777
We have the option of removing `addModule` from the `IModule` interface, make it public and then use `super`. Atm it's not used off that interface so I could go down that route. What do you think?

juniset: I personally think it is important to keep the `addModule` method in the `IModule` interface as it will protect us from future mistake.

Guaranteeing that every module can add another module combined with the fact that a wallet needs at least one module ensures the wallet can never be in a blocked state (in the sense of not being upgradable).

elenadimitrova: Yet another way is to switch `IModule` from an `interface` to `abstract contract` which allows public functions with no implementation at the same time as mandating all functions to be implemented.

juniset: Yes that works for me.

olivdb: We could keep `IModule` as-is but make `addModule` in `BaseModule` and `OnlyOwnerModule` be `public` instead of `external`.

elenadimitrova: That's also an option. I don't feel strongly for either of the two. I'll just go with your version as it causes less changes to current state. Let me know if that's okay

contracts/modules/common/BaseModule.sol
```diff
@@ -80,6 +80,17 @@ contract BaseModule is IModule {
         _;
     }
 
+    /**
+    * @dev Utility method enbaling anyone to recover ERC20 token sent to the
```
olivdb, line 84: enbaling typo :)

contracts/wallet/BaseWallet.sol
elenadimitrova: It's actually used in tests as we're not creating wallets that are Proxies but just new-ing up a `BaseWallet` instance. It's arguable whether this is correct and we shouldn't be testing with a proper Proxy-based wallet.

contracts/modules/common/BaseModule.sol
```diff
@@ -80,6 +80,17 @@ contract BaseModule is IModule {
...
+    * module by mistake and transfer them to the Module Registry.
+    * @param _token The token to recover.
+    */
+    function recoverToken(address _token) external override {
+        uint total = ERC20(_token).balanceOf(address(this));
+        bool success = ERC20(_token).transfer(address(registry), total);
```
olivdb, line 90: Maybe our 2.0 upgrade is a good opportunity to also enable recovery of non-compliant ERC20 tokens. See https://medium.com/coinmonks/missing-return-value-bug-at-least-130-tokens-affected-d67bf08521ca 
We could for example use OpenZeppelin's `SafeERC20.safeTransfer` or something simpler.

elenadimitrova: Valid point. Can you add it to trello? Trying to keep the upgrade PR focussed.

contracts/wallet/BaseWallet.sol
elenadimitrova: It helps fix the coverage problem though as I can remove the logic of `receive` which allows coverage to pass.

olivdb: Yes but removing that method won't break the tests as they currently stand. The `fallback()` of BaseWallet will be executed instead, so I believe we can remove `receive()` regardless of whether we decide to rewrite the tests to use `Proxy` or not

contracts/modules/common/BaseModule.sol
olivdb: Done

contracts/wallet/BaseWallet.sol
elenadimitrova: I does break the tests as fallback won't execute when there's no calldata received. That was the whole point of splitting the fallback function.

olivdb: As per the doc: 
```
A payable fallback function is also executed for plain Ether transfers, if no receive Ether function is present. [...]
```
https://solidity.readthedocs.io/en/v0.6.10/contracts.html#fallback-function

I also verified in Remix that the following contract is able to receive ether via empty calldata:
```
pragma solidity ^0.6.10;
contract FallbackTest {
    fallback() external payable {}
}
```

contracts/wallet/BaseWallet.sol
```diff
@@ -117,41 +120,38 @@ contract BaseWallet {
      * @param _value The value of the transaction.
      * @param _data The data of the transaction.
      */
-    function invoke(address _target, uint _value, bytes calldata _data) external moduleOnly returns (bytes memory _result) {
+    function invoke(address _target, uint _value, bytes calldata _data) external virtual moduleOnly returns (bytes memory _result) {
```
juniset, line 123: Why is `invoke()` a virtual method?

contracts/wallet/Proxy.sol
```diff
@@ -31,21 +32,22 @@ contract Proxy {
...
+            returndatacopy(0, 0, returndatasize())
+            switch result
+            case 0 {revert(0, returndatasize())}
+            default {return (0, returndatasize())}
+        }
+    }
 
-        if (msg.data.length == 0 && msg.value > 0) {
+    receive() external payable {
+        if (msg.value > 0) {
```
juniset, line 49: I think we can safely remove this `if` . The idea was to not log 0 ETH transfer, but there is no real reason to do that and we are basically making an unnecessary check on every transfer.

contracts/wallet/BaseWallet.sol
elenadimitrova: Okay it's not failing because plain ether is not accepted but because the call to `fallback` costs more than 21000 gas. https://app.circleci.com/pipelines/github/argentlabs/argent-contracts/707/workflows/26010b94-04c4-4a0c-ad62-c9cbe0e39afe/jobs/897
I can't think of any way to completely remove the `receive` function off `BaseWallet` really but feel free to push any changes you think are possible.

contracts/wallet/BaseWallet.sol
elenadimitrova: Because we are overriding it in `FakeWallet` test contract.

contracts/wallet/Proxy.sol
elenadimitrova: Ref https://github.com/argentlabs/argent-contracts/pull/114#discussion_r434496224

contracts/wallet/BaseWallet.sol
juniset: I see, but I don't really like that we are suggesting that the method can be overriden. I don't find it clean as our intention is for BaseWallet to be a final class.

We could have FakeWallet implement IWallet and duplicate the code of BaseWallet.

contracts/wallet/BaseWallet.sol
olivdb: I see. I don't love the idea of having a `receive()` function in `BaseWallet` that is only there for tests. On the other hand having to use a Proxy + BaseWallet in all our tests feels a bit heavy. Maybe using a `BaseWalletTest` in our tests, that inherits from `BaseWallet` and only adds an empty `receive()` to `BaseWallet` could be a middle ground? If not, let's just keep the empty `receive()`.

olivdb: Thinking about this more, testing with a Proxy + BaseWallet is probably the right move. We could maybe add a `createWallet()` method to `test-manager.js` that does the deployment of both contracts + wrapping of the proxy with the BaseWallet abi.

contracts/wallet/BaseWallet.sol
elenadimitrova: I tried using the Doppelganger mocking framework but found a problem instantiating it and logged  https://github.com/EthWorks/Doppelganger/issues/7
Ended up with code duplication until a proper solution is found.

olivdb:
Why doesn't `contracts-legacy/v1.3.0` contain only `BaseWallet.sol`? All the other files in there seem identical to those in `contracts-legacy/v1.5.0`

olivdb:
I think `contracts-legacy/v1.5.0` should be renamed `contracts-legacy/v1.6.0` as all the contracts in there are in their versions last used in `1.6`.

elenadimitrova:
>Why doesn't contracts-legacy/v1.3.0 contain only BaseWallet.sol and OldTestModule.sol? All the other files in there seem identical to those in contracts-legacy/v1.5.0 and are not used in tests.

Because otherwise we'll have cross references between legacy folders, in this case 1.3.0 importing contracts from 1.5.0. I think we should keep legacy copies atomic to a given release, otherwise it will likely get very tangled up.

elenadimitrova:
>I think contracts-legacy/v1.5.0 should be renamed contracts-legacy/v1.6.0 as all the contracts in there are in their versions last used in 1.6.
`TransferManager` is from 1.5 but I should probably change it to the 1.6 version as we are supporting up to one version back in upgrades. `MakerManager` hasn't changed so I can rename the folder and it'll be correct for all.

PR #119, opened 2020-06-11
SimpleUpgrader ensure a module is registered before upgrading

elenadimitrova:
Before we could perform an upgrade via SimpleUpgrader to a module that is not registered in the ModuleRegistry. Here we add a check to ensure that can't happen.

We also fix a small typo in a revert message in `WalletFactory` here.

### DIFF
--- a/contracts/modules/SimpleUpgrader.sol
+++ b/contracts/modules/SimpleUpgrader.sol
@@ -53,6 +53,7 @@ contract SimpleUpgrader is BaseModule {
         uint256 i = 0;
         //add new modules
         for (; i < toEnable.length; i++) {
+            require(registry.isRegisteredModule(toEnable[i]), "SU: module is not registered");
             BaseWallet(_wallet).authoriseModule(toEnable[i], true);
         }
         //remove old modules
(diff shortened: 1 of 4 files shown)


contracts/modules/SimpleUpgrader.sol
```diff
@@ -53,6 +53,7 @@ contract SimpleUpgrader is BaseModule {
         uint256 i = 0;
         //add new modules
         for (; i < toEnable.length; i++) {
+            require(registry.isRegisteredModule(toEnable[i]), "SU: module is not registered");
```
olivdb, line 56: Is this PR needed because:
1) we want the code to strictly enforce the constraint that "a module must be registered prior to being authorised by a wallet"
or 2) we want to catch the scenario where we forget to register modules that are being added via a registered upgrader ?

If 2), then we could do the check above in the constructor of `SimpleUpgrader` to save some gas.

juniset, line 56: I think we should use the `function isRegisteredModule(address[] calldata _modules) external view returns (bool);` method to check all the modules in one internal call and save some gas.

PR #120, opened 2020-06-17
Refactor TokenPriceProvider

elenadimitrova:
Refactor `TokenPriceProvider` into separate storage (`TokenPriceStorage`) and logic parts to allow future changes in logic without necessarily requiring to maintain two sets of prices in storage.  

The new design relies on prices set to be that of `price per token * 10^(36- token decimals)`. This allows us to eliminate the reliance on `decimals` property of tokens which is an optional ERC20 property that can't be relied on. e.g. [DGD token](https://etherscan.io/address/0xe0b7927c4af23765cb51314a0e0521a9645f0e2a#code). Additionally we've agreed to remove the`DGD` token from the token price provider. 

Also added `bn-chai` to help compare big numbers. https://www.chaijs.com/plugins/bn-chai/


contracts/modules/common/TokenPriceProvider.sol
```diff
@@ -16,33 +16,41 @@
...
+
+        try IERC20Extended(_token).decimals() returns (uint8 _decimals) {
+            decimals = _decimals;
+        }
+        catch (bytes memory /*lowLevelData*/) {
+        }
+
         uint256 price = cachedPrices[_token];
-        return price.mul(_amount).div(10**decimals);
+        uint256 etherValue = decimals == 0 ? price.mul(_amount) : price.mul(_amount).div(10**uint(decimals));
```
olivdb, line 43: Maybe drop the condition? 
```
uint256 etherValue = price.mul(_amount).div(10**uint(decimals));
```
should work for all types of tokens.

contracts/modules/common/TokenPriceProvider.sol
```diff
@@ -16,33 +16,41 @@
...
      * @return the ether value (in wei) of _amount tokens with contract _token
      */
     function getEtherValue(uint256 _amount, address _token) external view returns (uint256) {
-        uint256 decimals = ERC20(_token).decimals();
+        uint8 decimals;
+
+        try IERC20Extended(_token).decimals() returns (uint8 _decimals) {
+            decimals = _decimals;
+        }
+        catch (bytes memory /*lowLevelData*/) {
```
olivdb, line 39: Could be simplified to just `catch {}`

contracts/modules/common/TokenPriceProvider.sol
juniset, line 43: I'm a bit concerned about assuming `decimals == 0` when the `decimals` method is not present.

Looking for example at the DGD token (https://etherscan.io/token/0xE0B7927c4aF23765Cb51314A0E0521A9645F0E2A) I see that etherscan has the decimals set to 9 and this is the value we use in our token price provider. 

This means that calling `getEtherValue` will return a value 10**9 time bigger then what it should be. It is ok in terms of security (i.e. we are 10**9 times more drastic on the daily limit) but it is not consistent.

Maybe we could have a decimals mapping in our price provider for the few contracts that don't have the decimals method and use that value when the call to decimals fails?

elenadimitrova: Probably just saving some gas but maybe clarity is worth more.

elenadimitrova: To be sure I wrote a script to pull all tokens listed in the provider with no `decimals` property and there's only the DGD one. The value of 9 for its decimals we see in etherscan is really just a free text they've added and it's probably dangerous to pretend that token contract has 9 decimals when it doesn't because the property can be used elsewhere in our contracts. I think the price in the provider should be set to actual price per token: 192297647 as opposed to adding 9 zeros to this which is what we appear to have done.

lib/other/IERC20Extended.sol
```diff
@@ -0,0 +1,14 @@
+pragma solidity >=0.5.4 <0.7.0;
+
+/**
+ * Extended ERC20 contract interface to include the decimals property.
+ */
+interface IERC20Extended {
```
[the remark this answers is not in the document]
elenadimitrova: We need a bigger discussion around the ERC20 contracts we use so I logged [this](https://trello.com/c/agxNtzfQ/138-review-erc20-contracts-used). But to answer your question, OZ's `ERC20` is an implementation of their `IERC20` interface (which only declares the mandatory properties, i.e. no `decimals`). I would prefer to use an interface, rather than a contract because of the much flexible integration that provides. 

I'm going to go back to the ERC20 interface we have used elsewhere from /lib/other/ERC20.sol until we agree on a way forward.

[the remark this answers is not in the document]
elenadimitrova: The only thing to consider in this original ERC20 we have is the decimals property being a uint256 rather than a uint8 as per the standard. Should we add an assert for the decimals value fitting in a uint8 e.g. `require(_decimals < 256, "TPP: token decimals overflow");`?

contracts/modules/common/TokenPriceProvider.sol
olivdb: Taking a step back, it would be simpler and cleaner not to have to deal with decimals at all. The reason we are using decimals in the TPP is because, in a previous version of the TPP, we fetched the rates from Kyber which, unfortunately, returned a price per "whole token" instead of returning a price per "token wei". Now that we no longer use Kyber, we could simply have `cachedPrices[_token]` be the rate such that `cachedPrices[_token].div(10**18)` is the number of weis per "token wei". In this case `getEtherValue()` would simply become:
```
function getEtherValue(uint256 _amount, address _token) external view returns (uint256) {
        uint256 price = cachedPrices[_token];
        return price.mul(_amount).div(10**18);
    }
``` 
This would require some coordination with backend and clients though. But maybe worth investing this time to simplify our lives in the long term.

elenadimitrova: The problem is that it reduces accuracy for price per token wei and for 253 (out of 381) tokens that price even becomes zero.

olivdb: Copying here the examples discussed with Elena here for future reference:

Token= 0x595832f8fc6bf59c85c527fec3740a1b7a361269.
1 POWR = 0.000389118500000000 ETH.
Decimals = 6.
Currently we store cachedPrice = 389118500000000 = 0.000389118500000000*10^18  for that token.
With my proposal we would store cachedPrice = 0.000389118500000000 * 10^(36-6) = 389118500000000000000000000. And the getEtherValue would give:
 `getEtherValue(amount) = amount * cachedPrice / 10^18` (= 389118500 * amount)
whereas now it is calculated as:
`getEtherValue(amount) = amount * (0.000389118500000000*10^18) / 10^6` (= 389118500 * amount), which is equivalent.

Token= 0x05f4a42e251f2d52b8ed15e9fedaacfcef1fad27.
1 ZIL = 0.000079764579653685 ETH.
Decimals = 12.
Currently we store cachedPrice = 79764579653685 = 0.000079764579653685*10^18  for that token.
With my proposal we would store cachedPrice = 0.000079764579653685 * 10^(36-12) = 79764579653685000000. And the getEtherValue would give:
 `getEtherValue(amount) = amount * cachedPrice / 10^18` (~= 79.764579653685 * amount)
whereas now it is calculated as:
`getEtherValue(amount) = amount * (0.000079764579653685*10^18) / 10^12` (~= 79.764579653685 * amount), which is equivalent.

So essentially, the cachedPrices would just become whatever they are now multiplied by 10^(18-decimals) and getEtherValue would become `getEtherValue(amount) = amount * cachedPrice / 10^18` to take this change into account.

test/approvedTransfer.js
```diff
@@ -60,8 +58,11 @@ describe("Approved Transfer", function () {
     wallet = deployer.wrapDeployedContract(BaseWallet, proxy.contractAddress);
 
     await wallet.init(owner.address, [approvedTransfer.contractAddress, guardianManager.contractAddress]);
-    erc20 = await deployer.deploy(ERC20, {}, [infrastructure.address, wallet.contractAddress], 10000000, DECIMALS); // TOKN contract with 10M tokens (5M TOKN for wallet and 5M TOKN for account[0])
-    await priceProvider.setPrice(erc20.contractAddress, TOKEN_RATE);
+
+    const decimals = 12; // number of decimal for TOKN contract
+    const tokenRate = new BN(10).pow(new BN(37)).muln(51); // 1 TOKN = 0.00051 ETH = 0.00051*10^18 ETH wei => * 10^(36-decimals) = 0.00051*10^18 * 10^24 = 51 * 10^37
+    erc20 = await deployer.deploy(ERC20, {}, [infrastructure.address, wallet.contractAddress], 10000000, decimals); // TOKN contract with 10M tokens (5M TOKN for wallet and 5M TOKN for account[0])
+    await tokenPriceStorage.setPrice(erc20.contractAddress, tokenRate.toString());
```
juniset, line 65: Do we still need the token price contract in this test file? It is not used by the ApprovedTransfer modules so it can probably be removed completely.

contracts/infrastructure/storage/TokenPriceStorage.sol
```diff
@@ -0,0 +1,46 @@
...
+ * @title TokenPriceStorage
+ * @notice Contract storing the token prices.
+ * @dev Note that prices stored here = price per token * 10^(36-token decimals)
+ * The contract only defines basic setters and getters with no logic.
+ * Only managers of this contract can modify its state.
+ */
+contract TokenPriceStorage is ITokenPriceStorage, Storage, Managed {
+    mapping(address => uint256) public cachedPrices;
+
+    function getTokenPrice(address _token) external override view returns (uint256 _price) {
```
juniset, line 33: Can we add a method to get the prices for a list of tokens? This could be needed in the future to convert from one token to another.

contracts/modules/TransferManager.sol
```diff
@@ -177,6 +177,20 @@ contract TransferManager is OnlyOwnerModule, BaseTransfer {
         }
     }
 
+    /**
+    * @notice Get the ether value equavalent to the token amount.
```
olivdb, line 181: Typo: equavalent

PR #121, opened 2020-06-18
Paraswap Integration for TokenExchangerV2

olivdb:
This PR introduces a new version of the TokenExchanger module that relies on Paraswap instead of Kyber. This brings the following benefits:

- ERC20 to ERC20 trades
- Reduced gas cost of every trade
- Improved price and liquidity, specially on larger trades.
- Routing to multiple liquidity provider (0x, Uniswap, Kyber, Curve, etc)


contracts/modules/TokenExchanger.sol
```diff
@@ -33,139 +35,238 @@ contract TokenExchanger is OnlyOwnerModule {
 
     // Mock token address for ETH
     address constant internal ETH_TOKEN_ADDRESS = 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE;
+    // Signatures of Paraswap's trade methods
+    bytes4 constant internal MULTISWAP = 0xcbd1603e;
```
juniset, line 39: Maybe useful to have the computation of the signatures in the comment (`bytes4(keccak256(.....)`) so we know exactly what is called.

contracts/modules/TokenExchanger.sol
```diff
@@ -33,139 +35,238 @@ contract TokenExchanger is OnlyOwnerModule {
...
+     * @param _srcAmount The exact amount of source tokens to sell.
+     * @param _minDestAmount The minimum amount of destination tokens required for the trade.
+     * @param _expectedDestAmount The expected amount of destination tokens (used only in ParaSwap's events).
+     * @param _path Sequence of sets of weighted ParaSwap routes. Each route specifies an exchange to use to convert a given (exact) amount of a given
+     * source token into a given (minimum) amount of a given destination token. The path is a sequence of sets of weighted routes where the destination
+     * token of a set of weighted routes matches the source token of the next set of weighted routes in the path.
+     * @param _mintPrice gasPrice (in wei) at the time the gas tokens were minte by ParaSwap. 0 means gas token will not be used by ParaSwap
      */
-    function trade(
+    function multiSwap(
```
juniset, line 88: Why not name this method `sell` since the other one is called `buy`? It will probably avoid some confusion.

contracts/modules/TokenExchanger.sol
```diff
@@ -33,139 +35,238 @@ contract TokenExchanger is OnlyOwnerModule {
...
         }
+    }
 
-        if (fee > 0) {
-            invokeWallet(_wallet, feeCollector, fee, "");
+    function approveToken(address _wallet, address _token, uint _amount) internal returns (uint256 _existingAllowance) {
+        // TODO: Use a "safe approve" logic similar to the one implemented below in other modules
+        if (_token != ETH_TOKEN_ADDRESS) {
+            _existingAllowance = ERC20(_token).allowance(_wallet, paraswapProxy);
+            if (_existingAllowance < uint256(-1)) {
```
juniset, line 186: Why is this condition necessary?

contracts/modules/TokenExchanger.sol
```diff
@@ -33,139 +35,238 @@ contract TokenExchanger is OnlyOwnerModule {
...
         public
     {
-        kyber = _kyber;
-        feeCollector = _feeCollector;
-        feeRatio = _feeRatio;
+        paraswapSwapper = _paraswap;
+        paraswapProxy = IAugustusSwapper(_paraswap).getTokenTransferProxy();
+        referrer = _referrer;
+
+        for (uint i = 0; i < _authorisedExchanges.length; i++) {
```
juniset, line 70: I think using a black list instead of a white list would give us more flexibility. When Paraswap adds a new DEX we will want to benefit from it without having to redeploy a new module. That's one of the main benefit of using an aggregator. We should only redeploy a new module when the added DEX does not meet our security requirements.

contracts/modules/TokenExchanger.sol
```diff
@@ -33,139 +35,238 @@ contract TokenExchanger is OnlyOwnerModule {
...
+     */
+    function buy(
+        address _wallet,
+        address _srcToken,
         address _destToken,
-        uint256 _maxDestAmount,
-        uint256 _minConversionRate
+        uint256 _maxSrcAmount,
+        uint256 _destAmount,
+        uint256 _expectedDestAmount,
```
juniset, line 140: Is `_expectedDestAmount` returned by the API and is it different than `_destAmount` in a buy? Maybe `_destAmount` is not needed?

contracts/modules/TokenExchanger.sol
olivdb: Yes, I hesitated but in the end decided to use naming consistent with Paraswap's contract (which uses `buy` and `multiSwap`). Open to both.

contracts/modules/TokenExchanger.sol
olivdb: If the user has given an unlimited allowance to Paraswap (via WalletConnect), we should not change the allowance. That's not only to save gas in this particular transaction, but also because many tokens (DAI, USDT, etc) optimise `transferFrom` when the allowance is `uint(-1)` so as to save gas and I think we should not break this optimisation when the user opts for it.

olivdb: Also, without this condition `SafeMath.add(_existingAllowance, _amount)` would revert when  `_existingAllowance == uint(-1)`.

contracts/modules/TokenExchanger.sol
olivdb: This is a typo: it should be `_expectedSrcAmount`. It's given by the API and is only used by Paraswap's `Bought` event.

contracts/modules/TokenExchanger.sol
olivdb: Done

contracts/modules/TokenExchanger.sol
olivdb: Fixed

contracts/modules/TokenExchanger.sol
juniset: I would go for `buy` and `sell` irrespective of their names in Paraswap. In the end it will make it clearer for us what each method does (i.e. simpler for clients, support, etc).

contracts/modules/TokenExchanger.sol
```diff
@@ -33,139 +35,240 @@ contract TokenExchanger is OnlyOwnerModule {
...
+     */
+    function buy(
+        address _wallet,
+        address _srcToken,
         address _destToken,
-        uint256 _maxDestAmount,
-        uint256 _minConversionRate
+        uint256 _maxSrcAmount,
+        uint256 _destAmount,
+        uint256 _expectedSrcAmount,
```
juniset, line 142: This is a detail but I would expect the `_expectedSrcAmount` to be right after the `_maxSrcAmount` in the parameters of the method so you have the expected and the worst case amounts next to each other. I think it makes the inputs more readable.

contracts/modules/TokenExchanger.sol
juniset: For the record we have decided to keep the (fixed) whitelist as it is the only option that does not place any security/trust assumption on both Argent and Paraswap.

contracts/modules/TokenExchanger.sol
olivdb: Agreed and fixed

PR #122, opened 2020-06-22
Separate limit storage

juniset:
We are extracting the storage of the state of the daily limit of wallets from the `TransferManager` to a separate `LimitStorage` contract. This is done to enable other modules to check and update the daily limit.

The LimitManager no longer needs to be an abstract module and is converted to the helper LimitUtils library.

Infrastructure contracts that need to be compiled with solc 0.5.x are separated to their own `infrastructure_0.5` folder.


contracts/modules/common/LimitManager.sol
```diff
@@ -113,115 +81,62 @@ abstract contract LimitManager is BaseModule {
...
+            lStorage.setDailySpent(_wallet, safe128(alreadySpent.add(_amount)), safe64(periodEnd));
+            return true;
         }
+        return false;
     }
 
     /**
-    * @dev Helper method to check if a transfer is within the limit.
-    * If yes the daily unspent for the current period is updated.
+    * @dev Checks if a transfer is within the limit. If yes the daily spent is not updated.
```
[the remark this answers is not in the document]
juniset: It means that we are just checking if the transfer is within the daily limit but we are not updating the `dailySpent`. It only reads the storage but does not modify it. 

This is different than `checkAndUpdateDailySpent` which does the same but updates the `dailySpent` when the transfer is under the limit.

Note that this method is only used once in the `verifyRefund` logic which I'm planning on removing in the follow up PR.

scripts/provision_lib_artefacts.sh
```diff
@@ -27,7 +27,7 @@ cp build/DSValue.json .coverage_artifacts/DSValue.json
 cp build/CryptoKittyTest.json .coverage_artifacts/CryptoKittyTest.json
 cp build/TestRegistry.json .coverage_artifacts/TestRegistry.json
 cp build/GemJoin.json .coverage_artifacts/GemJoin.json
-cp build/OldTestModule.json .coverage_artifacts/OldTestModule.json
+cp build-legacy/v1.3.0//OldTestModule.json .coverage_artifacts/OldTestModule.json
```
elenadimitrova, line 30: This line can be removed altogether.

contracts/modules/common/LimitManager.sol
```diff
@@ -230,12 +145,22 @@ abstract contract LimitManager is BaseModule {
...
+    function safe128(uint256 _num) internal pure returns (uint128) {
+        require(_num < 2**128, "LM: more then 128 bits");
+        return uint128(_num);
+    }
+
+    function safe64(uint256 _num) internal pure returns (uint64) {
+        require(_num < 2**64, "LM: more then 64 bits");
+        return uint64(_num);
+    }
+
```
elenadimitrova, line 165: This should probably live in the `ArgentSafeMath` library.

contracts/modules/common/LimitManager.sol
```diff
@@ -113,115 +81,62 @@ abstract contract LimitManager is BaseModule {
...
-            if (expense.periodEnd < now) {
-                // solium-disable-next-line security/no-block-members
-                expense.periodEnd = uint64(now + 24 hours);
-                expense.alreadySpent = uint128(_amount);
-            } else {
-                expense.alreadySpent += uint128(_amount);
-            }
-        }
+    function getLimit(address _wallet) internal view returns (uint256, uint256, uint256) {
+        return lStorage.getLimit(_wallet);
```
elenadimitrova, line 130: I found the combination of this `getLimit` and `currentLimit` functions confusing. It appears that `getLimit` does not give me back my wallet's actual limit if the securityPeriod has passed but for that I need to call `currentLimit`. I suggest we only have one getter for limit which gives us the currently active wallet limit. If we made the `limits` mapping public  whoever is interested in the actual values can query that.

contracts/modules/common/LimitManager.sol
```diff
@@ -91,13 +61,11 @@ abstract contract LimitManager is BaseModule {
      * @param _securityPeriod The security period.
      */
     function changeLimit(address _wallet, uint256 _newLimit, uint256 _securityPeriod) internal {
```
elenadimitrova, line 63: Should we require that the new limit is > 0? Otherwise that effectively means transfers and token approvals are disabled and also no refunds from the TransferManager can happen. Might even be sensible to have a minimum required here?

contracts/modules/common/LimitManager.sol
```diff
@@ -17,72 +17,42 @@
...
 import "./BaseModule.sol";
+import "../../infrastructure/storage/ILimitStorage.sol"; 
 
 /**
  * @title LimitManager
  * @dev Module to manage a daily spending limit
- * @author Julien Niset - <julien@argent.im>
+ * @author Julien Niset - <julien@argent.xyz>
  */
 abstract contract LimitManager is BaseModule {
```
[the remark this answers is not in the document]
juniset: I've removed the inheritance from `BaseModule` but did not make it a library for simplicity.

contracts/infrastructure/storage/LimitStorage.sol
```diff
@@ -0,0 +1,104 @@
...
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.5.4;
```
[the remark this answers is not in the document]
juniset: I've separated all infrastructure contracts that need to be compiled with solc 0.5 into their own `infrastructure_0.5` folder.

contracts/modules/common/LimitManager.sol
juniset: The problem is that `ArgentSafeMath` is under `/module` while these methods are only used in `/infrastructure/storage`. 

If we want to do some reshuffling we should do it in a separate PR so I'll these 2 methods there for now.

contracts/modules/common/LimitManager.sol
```diff
@@ -91,13 +61,11 @@ abstract contract LimitManager is BaseModule {
      * @param _securityPeriod The security period.
      */
     function changeLimit(address _wallet, uint256 _newLimit, uint256 _securityPeriod) internal {
-        Limit storage limit = limits[_wallet].limit;
+        (uint256 current, uint256 pending, uint256 changeAfter) = getLimit(_wallet);
```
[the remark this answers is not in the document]
juniset: To return structs from non-internal methods we need the `ABIEncoderV2` enabled which is still flagged as `experimental`.

contracts/modules/common/LimitManager.sol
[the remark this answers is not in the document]
elenadimitrova: The comment currently means it only doesn't update the daily spent if it's within the limit. Maybe reword just to:

>Checks if a transfer is within the limit.

contracts/modules/common/LimitManager.sol
juniset: No. Setting the limit to 0 turns the Wallet into a normal multisig where all transfers need to be approved multiple keys (the guardians). I find that use case interesting and I know some of users are using their wallet that way.

contracts/modules/common/LimitManager.sol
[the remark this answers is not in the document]
juniset: In the end I renamed LimitManager to LimitUtils and converted it into a library.

contracts/modules/common/LimitManager.sol
juniset: I've left them in the new `LimitUtils` library for now as they are only used in the context of the daily limit.

contracts/modules/common/LimitManager.sol
juniset: I've removed the `getLimit` method from the `LimitUtils` library.

contracts/modules/common/LimitManager.sol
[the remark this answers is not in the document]
olivdb: FYI the doc now says
```
 [ABIEncoderV2] is considered non-experimental as of Solidity 0.6.0. You still have to explicitly activate it using pragma experimental ABIEncoderV2; - we kept the same pragma, even though it is not considered experimental anymore.
```

contracts/modules/common/LimitUtils.sol
```diff
@@ -0,0 +1,195 @@
...
+
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.6.10;
+
+import "@openzeppelin/contracts/math/SafeMath.sol";
+import "../../infrastructure/storage/ILimitStorage.sol";
+
+/**
+ * @title LimitManager
+ * @dev Helper library to manage the daily limit and interface with a IILimitStorage contract.
```
olivdb, line 24: typo `IILimitStorage`

contracts/modules/TransferManager.sol
```diff
@@ -52,12 +52,16 @@ contract TransferManager is OnlyOwnerModule, BaseTransfer, LimitManager {
...
     uint256 public securityWindow;
+    // The default limit
+    uint128 public defaultLimit;
     // The Token storage
     ITransferStorage public transferStorage;
     // The Token price provider
     TokenPriceProvider public priceProvider;
     // The previous limit manager needed to migrate the limits
-    LimitManager public oldLimitManager;
+    TransferManager public oldLimitManager;
```
olivdb, line 62: For clarity, since we are doing away with the concept of a LimitManager, I would rename that variable `oldTransferManager`

contracts/modules/TransferManager.sol
```diff
@@ -372,7 +378,64 @@ contract TransferManager is OnlyOwnerModule, BaseTransfer, LimitManager {
...
+            uint256 alreadySpent,
+            uint256 periodEnd
+        ) = limitStorage.getLimitAndDailySpent(_wallet);
+        uint256 currentLimit = LimitUtils.currentLimit(current, pending, changeAfter);
+        // solium-disable-next-line security/no-block-members
+        if (now > periodEnd) {
+            // solium-disable-next-line security/no-block-members
+            return (currentLimit, uint64(now.add(24 hours)));
+        } else if (alreadySpent < currentLimit) {
+            return (currentLimit.sub(alreadySpent),uint64(periodEnd));
```
olivdb, line 435: Would need a space after the `,` to appease my OCD 😬

contracts/modules/common/LimitUtils.sol
```diff
@@ -0,0 +1,195 @@
...
+    {
+        // solium-disable-next-line security/no-block-members
+        if (_changeAfter > 0 && _changeAfter < now) {
+            return _pending;
+        }
+        return _current;
+    }
+
+    function safe128(uint256 _num) internal pure returns (uint128) {
+        require(_num < 2**128, "LM: more then 128 bits");
```
olivdb, line 186: LM -> LU

contracts/infrastructure/storage/LimitStorage.sol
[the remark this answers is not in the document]
olivdb: What is `infrastructure_0.5` used for? Also, since you didn't specify a different `--buildDirectory` for `compile:infrastructure_0.5`, I think the artifacts that this command produces will just be overwritten by the ones produced by `compile:infrastructure_0.6`.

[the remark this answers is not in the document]
juniset: This is just to separate the infrastructure contracts that can be compiled with solc 0.6 and the ones that are currently in prod and require 0.5. There is no overlapping between the 2 folders.

contracts/modules/TransferManager.sol
```diff
@@ -29,7 +29,7 @@ import "../../lib/other/ERC20.sol";
  * This module is the V2 of TokenTransfer.
  * @author Julien Niset - <julien@argent.xyz>
  */
-contract TransferManager is OnlyOwnerModule, BaseTransfer, LimitManager {
+contract TransferManager is OnlyOwnerModule, BaseTransfer {
```
elenadimitrova, line 32: ❤️

contracts/modules/common/LimitManager.sol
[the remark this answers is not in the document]
juniset: I've enabled the `ABIEncoderV2` and introduced structs.

PR #128, opened 2020-07-07
Revisited upgrade logic when wallet under recovery (locked)

elenadimitrova:
Disabled the ability to add a module (and therefore upgrade) when a wallet is locked. A wallet can be locked either explicitly by a guardian or during a wallet recovery. 

Also remove obsoleted `securityWindow` and `securityPeriod` properties from `RecoveryManager`.


contracts/modules/RecoveryManager.sol
```diff
@@ -87,18 +83,14 @@ contract RecoveryManager is RelayerModule {
...
-        uint256 _lockPeriod,
-        uint256 _securityPeriod,
-        uint256 _securityWindow
+        uint256 _lockPeriod
     )
         BaseModule(_registry, _guardianStorage, NAME)
         public
     {
-        require(_lockPeriod >= _recoveryPeriod && _recoveryPeriod >= _securityPeriod + _securityWindow, "RM: insecure security periods");
+        require(_lockPeriod >= _recoveryPeriod, "RM: insecure security periods");
```
juniset, line 91: It makes sense to remove these 2 parameters from the constructor since they are never used and there is no guarantee that they correspond to the values used in the other modules. But we should add a comment on the contract to make sure we have validated that condition before deployment, something along the line of "For the wallet to be secure we must have `_recoveryPeriod >= _securityPeriod + _securityWindow` where `_securityPeriod` and `_securityWindow` are the security parameters of adding/removing guardians and confirming large transfers.".

elenadimitrova: When I was analysing the various security windows I thought about potentially consolidating these into the `BaseModule` in order to ensure consistency. The only disadvantage I could think of is the ability to change these independently on modules which I'm not sure we'll ever want. What do you think?

```
GuardianManager
securityWindow 43200
securityPeriod 86400

TransferManager
securityWindow 43200
securityPeriod 86400

RecoveryManager
recoveryPeriod 129600
lockPeriod       432000

LockManager
lockPeriod 432000
```

juniset: There has been several requests from users to increase these security periods so we might change them at some point.  One of these things that we discuss periodically... While it is not concrete enough to influence our design, I still think I would prefer to not have these hardcoded in every module and keep some flexibility.

For now I think a clear comment in the contract is sufficient.

elenadimitrova: done

contracts/modules/common/IModule.sol
```diff
@@ -26,20 +26,19 @@ pragma solidity >=0.5.4 <0.7.0;
...
     /**
-     * @dev Inits a module for a wallet by e.g. setting some wallet specific parameters in storage.
+     * @notice Inits a module for a wallet by e.g. setting some wallet specific parameters in storage.
      * @param _wallet The wallet.
      */
     function init(address _wallet) external;
 
     /**
-     * @dev Adds a module to a wallet.
+     * @notice Adds a module to a wallet. Cannot execute when walelt is locked (or under recovery)
```
juniset, line 41: typo in wallet

PR #130, opened 2020-07-09
Improve the required signatures logic in relayer

elenadimitrova:
Add an entry in the `OwnerSignature` enum to make the intention clearer that there are cases where no mandated signatures are required.

Also here squeezed in a new npm command for performing a clean compile.

### DIFF
--- a/contracts/modules/common/RelayerModule.sol
+++ b/contracts/modules/common/RelayerModule.sol
@@ -91,6 +91,10 @@ abstract contract RelayerModule is BaseModule {
         require(checkAndUpdateUniqueness(_wallet, _nonce, signHash), "RM: Duplicate request");
         require(verifyData(_wallet, _data), "RM: Target of _data != _wallet");
         (uint256 requiredSignatures, OwnerSignature ownerSignatureRequirement) = getRequiredSignatures(_wallet, _data);
+        if (requiredSignatures == 0) {
+            require(ownerSignatureRequirement == OwnerSignature.Anyone, "RM: Wrong number of required signatures");
+        }
+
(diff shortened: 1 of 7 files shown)


contracts/modules/common/RelayerModule.sol
```diff
@@ -90,6 +91,10 @@ abstract contract RelayerModule is BaseModule {
         require(checkAndUpdateUniqueness(_wallet, _nonce, signHash), "RM: Duplicate request");
         require(verifyData(_wallet, _data), "RM: Target of _data != _wallet");
         (uint256 requiredSignatures, OwnerSignature ownerSignatureRequirement) = getRequiredSignatures(_wallet, _data);
+        if (requiredSignatures == 0) {
+            require(ownerSignatureRequirement == OwnerSignature.Anyone, "RM: Wrong number of required signatures");
+        }
+
```
juniset, line 97: I think it would be a little cleaner to have that condition in a single line require:
```
require(requiredSignatures > 0 || ownerSignatureRequirement == OwnerSignature.Anyone, "RM: Wrong number of required signatures");
```

package.json
```diff
@@ -17,6 +17,7 @@
     "compile:wallet": "npx etherlime compile --workingDirectory contracts/wallet --runs=999",
     "compile:test": "npx etherlime compile --workingDirectory contracts-test --runs=999",
     "compile": "npm run compile:infrastructure && npm run compile:modules && npm run compile:wallet",
+    "cc": "rm -rf build && npm run compile:lib && npm run compile && npm run compile:legacy && npm run compile:test",
```
juniset, line 20: This will make our life so much easier :-)

We should probably add `rm -rf build-legacy` as well.

PR #132, opened 2020-07-10
Wallets can only be created with a guardian

elenadimitrova:
Remove factory methods that allowed an empty guardian in wallet creation. Additionally we've removed the ability to change the `GuardianStorage` and assigning it in the constructor.

The `createCounterfactualWallet` method now returns the address of the wallet.


contracts/wallet/WalletFactory.sol
```diff
@@ -44,95 +44,49 @@ contract WalletFactory is Owned, Managed {
...
-     * The wallet is initialised with a list of modules, a first guardian, and an ENS..
+     * The wallet is initialised with a list of modules, a first guardian, and an ENS.
      * The wallet is created using the CREATE opcode.
      * @param _owner The account address.
      * @param _modules The list of modules.
      * @param _label ENS label of the new wallet, e.g. franck.
      * @param _guardian The guardian address.
      */
-    function createWalletWithGuardian(
+    function createWallet(
```
juniset, line 75: Do we want to change the name of this method? I think it doesn't hurt to be explicit about our intentions but it also means less code change on the backend.

contracts/wallet/WalletFactory.sol
```diff
@@ -143,7 +97,7 @@ contract WalletFactory is Owned, Managed {
      * @param _guardian The guardian address.
      * @param _salt The salt.
      */
-    function createCounterfactualWalletWithGuardian(
+    function createCounterfactualWallet(
```
juniset, line 100: Same comment as above: do we want to change the name?

contracts/wallet/WalletFactory.sol
```diff
@@ -308,10 +198,9 @@ contract WalletFactory is Owned, Managed {
         }
         // initialise the wallet with the owner and the extended modules
         _wallet.init(_owner, extendedModules);
-        // add guardian if needed
-        if (_guardian != address(0)) {
-            IGuardianStorage(guardianStorage).addGuardian(address(_wallet), _guardian);
-        }
+        // add guardian
```
juniset, line 201: The `(Optional)` word should be removed from the `_guardian` param in the method description.

contracts/wallet/WalletFactory.sol
```diff
@@ -44,95 +44,49 @@ contract WalletFactory is Owned, Managed {
...
-        _;
-    }
-
     // *************** Constructor ********************** //
 
     /**
      * @dev Default constructor.
      */
-    constructor(address _moduleRegistry, address _walletImplementation, address _ensManager) public {
+    constructor(address _moduleRegistry, address _walletImplementation, address _ensManager, address _guardianStorage) public {
```
juniset, line 54: Now I remember why the guardianStorage was provided with a setter: it is because in the deployment scripts we deploy the wallet factory before we deploy the modules and their storage. If the guardian storage is passed in the constructor of the factory we will need to update the scripts and the order in which contracts are deployed.

elenadimitrova: That's fine I think. Normally I'd expect storage contracts to be deployed first so I think I prefer that option. 
I've fixed the deployment scripts which I hadn't updated with the new constructor though when I was looking at this.

contracts/wallet/WalletFactory.sol
elenadimitrova: I prefer the shorter, cleaner name. Otherwise it'll be more correct to call it `createWalletWithModulesLabelAndGuardian`. I have not yet understood the complexities of the code changes in the back but I don't think simplicity and clarity should be sacrificed for the sake of renaming two methods.

contracts/wallet/WalletFactory.sol
```diff
@@ -196,7 +136,10 @@ contract WalletFactory is Owned, Managed {
         returns (address _wallet)
     {
         require(_guardian != (address(0)), "WF: guardian cannot be null");
```
juniset, line 138: Maybe we should add a check to make sure the `_owner` is not `address(0)` and that the length of `_modules` is not 0.

juniset, line 138: To be consistent we should add a check that `_owner != address(0)` and that `_modules.length > 0` as we know we cannot create wallets without these 2 conditions.

elenadimitrova: Since we validate these in the create calls, I didn't think it's necessary but we can indeed do the same validation here via `_validateInputs(_owner, _modules, _label, _guardian);`

juniset: The problem is that `_validateInputs` also checks that the modules are registered which I feel should not be checked in the ` getAddressForCounterfactualWallet` method.

PR #133, opened 2020-07-14
Allow whitelisted tokens to be called in TransferManager

elenadimitrova:
Adds a `isWhitelisted()` check inside `TransferManager.authoriseContractCall()`.

### DIFF
--- a/contracts/modules/TransferManager.sol
+++ b/contracts/modules/TransferManager.sol
@@ -248,7 +248,7 @@ contract TransferManager is OnlyOwnerModule, BaseTransfer {
     {
         if (!isWhitelisted(_wallet, _contract)) {
             // Make sure we don't call a supported ERC20 that's not whitelisted
-            notCoveredByDailyLimit(_wallet, _contract);
+            require(!coveredByDailyLimit(_wallet, _contract), "TM: Forbidden contract");
 
             require(LimitUtils.checkAndUpdateDailySpent(limitStorage, _wallet, _value), "TM: Call contract above daily limit");
         }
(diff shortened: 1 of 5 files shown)


contracts/modules/TransferManager.sol
```diff
@@ -532,11 +530,10 @@ contract TransferManager is OnlyOwnerModule, BaseTransfer {
     * @param _wallet The target wallet.
     * @param _contract The address of the contract.
      */
-    function authoriseContractCall(address _wallet, address _contract) internal view {
+    function isNotManagedTokenOrLimitDisabled(address _wallet, address _contract) internal view {
```
olivdb, line 533: I'd maybe rename `isNotManagedTokenOrLimitDisabled` to `isNotKnownTokenOrLimitDisabled` since our `tokenPriceStorage` does not really "manage" tokens.

contracts/modules/common/BaseTransfer.sol
```diff
@@ -43,9 +43,18 @@ abstract contract BaseTransfer is BaseModule {
         bytes data
     );
     // *************** Internal Functions ********************* //
+    /**
+    * @notice Make sure a contract call is not trying to call a module, the wallet itself, or a supported ERC20.
+    * @param _wallet The target wallet.
+    * @param _contract The address of the contract.
+     */
+    modifier authorisedContractCall(address _wallet, address _contract) {
```
olivdb, line 51: Maybe prefix the modifier with `only` to be consistent with our other modifiers?

contracts/modules/TransferManager.sol
juniset: To go one step further I think the name of this method should reflect that we are checking if the target contract is protected by the daily limit (not just known). So I would suggest something like `isNotCoveredByLimitOrLimitDisabled` returning a boolean and coupled with a require to make the code more readable.

elenadimitrova: Maybe `isNotCoveredByLimit` also covers `..LimitDisabled` part of the name and we can keep it shorter to `notCoveredByDailyLimit`?

juniset: Yes, with the `require(notCoveredByDailyLimit()...'  in the external methods like it :-)

PR #134, opened 2020-07-15
Adding success checks to CompoundManager

olivdb:
CToken methods such as `mint(uint256)` do not revert when failing but return a non-zero error code instead. 
This PR adds checks to ensure that the expected token balance changes have occurred after a Compound method is executed.

### DIFF
--- a/contracts/modules/CompoundManager.sol
+++ b/contracts/modules/CompoundManager.sol
@@ -393,9 +393,9 @@ contract CompoundManager is OnlyOwnerModule {
     function borrow(address _wallet, address _token, address _cToken, uint256 _amount) internal {
         require(_cToken != address(0), "CM: No market for target token");
         require(_amount > 0, "CM: amount cannot be 0");
-        uint256 initialTokenAmount = _token == ETH_TOKEN_ADDRESS ? _wallet.balance : ERC20(_token).balanceOf(_wallet);
+        uint256 initialTokenAmount = _token == ETH_TOKEN ? _wallet.balance : ERC20(_token).balanceOf(_wallet);
         invokeWallet(_wallet, _cToken, 0, abi.encodeWithSignature("borrow(uint256)", _amount));
-        uint256 finalTokenAmount = _token == ETH_TOKEN_ADDRESS ? _wallet.balance : ERC20(_token).balanceOf(_wallet);
+        uint256 finalTokenAmount = _token == ETH_TOKEN ? _wallet.balance : ERC20(_token).balanceOf(_wallet);
         require(finalTokenAmount > initialTokenAmount, "CM: borrow failed to increase token balance");
     }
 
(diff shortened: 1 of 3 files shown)


contracts/modules/CompoundManager.sol
```diff
@@ -370,21 +376,27 @@ contract CompoundManager is OnlyOwnerModule {
...
      * @dev Borrows underlying tokens from a cToken contract.
      * @param _wallet The target wallet.
+     * @param _token The token contract.
      * @param _cToken The cToken contract.
      * @param _amount The amount of underlying tokens to borrow.
      */
-    function borrow(address _wallet, address _cToken, uint256 _amount) internal {
-        require(_cToken != address(0), "Compound: No market for target token");
-        require(_amount > 0, "Compound: amount cannot be 0");
+    function borrow(address _wallet, address _token, address _cToken, uint256 _amount) internal {
```
juniset, line 393: Why do we need to pass the `_token`? We could get it from the `_cToken` contract and not have to add another parameter to the method. Or do the opposite and pass the token and get the cToken from the registry.

olivdb: We could but why pay the gas cost of an external call + extra read when we already have the data available? It's just an internal function.

contracts/modules/CompoundManager.sol
```diff
@@ -358,9 +360,13 @@ contract CompoundManager is OnlyOwnerModule {
      * @param _amount The amount of cToken to redeem.
      */
     function redeem(address _wallet, address _cToken, uint256 _amount) internal {
-        require(_cToken != address(0), "Compound: No market for target token");
-        require(_amount > 0, "Compound: amount cannot be 0");
+        // The following commented `require()` is not necessary as `ICToken(cToken).balanceOf(_wallet)` in `removeInvestment()` would have reverted if `_cToken == address(0)`
+        // It is however left as a comment as a reminder to include it if `removeInvestment()` is changed to use amounts instead of fractions.
+        // require(_cToken != address(0), "CM: No market for target token");
```
juniset, line 365: I think I would prefer to keep the require. Otherwise we are making the assumption that `removeInvestment` will not change and that no new method will be added that calls `redeem`. Also, are we sure that `ICToken(cToken).balanceOf(_wallet)` reverts if `cToken = address(0)`. If I remember correctly calling a method on a EOA just returns `0x`.

Anyway, I think the code would be cleaner and more robust with this simple check.

contracts/modules/CompoundManager.sol
juniset: Of course, I did not notice it was an internal method.

contracts/modules/CompoundManager.sol
olivdb: Yes I verified that `ICToken(cToken).balanceOf(_wallet)` reverts if `cToken` is an EOA. I think you are thinking of low level calls that succeed even when the target is an EOA (`cToken.call(abi.encodeWithSignature("balanceOf"))` would succeed and return `0x` for an EOA).

I removed that `require` because I realised that there was no way to hit it in coverage. I don't think this `require` is so critical that we want to reduce our coverage to include it (it's just there to save a bit of gas if the inputs are incorrect) so if you really want to keep it I'd suggest to move it to before the `balanceOf` call in `removeInvestment` so that it can be checked in our tests/coverage.

juniset: If it was security related I would have wanted to keep it there, but since it's only purpose is to save some gas I'm happy to have it commented out.

PR #137, opened 2020-07-16
Switch Solidity linter to solhint

elenadimitrova:
Replaces `ethlint` with `solhint` as the solium library doesn't seem to be maintained atm and is not supporting 0.6 syntax.

The `precommit` logic has also been updated to use `lint-staged` package instead of a script. This required the upgrade of node as it needs at least `node v10.21`. 

Also move WalletFactory back to `infrastructure` contracts. This was moved out in #114 due to these being compiled with 0.5 but were since split to allow for two bundles of infrastructure contracts targeting 0.5 and 0.6.


contracts/modules/TokenExchanger.sol
```diff
@@ -82,9 +81,9 @@ contract TokenExchanger is OnlyOwnerModule {
      * @param _srcAmount The exact amount of source tokens to sell.
      * @param _minDestAmount The minimum amount of destination tokens required for the trade.
      * @param _expectedDestAmount The expected amount of destination tokens (used only in ParaSwap's Swapped event).
-     * @param _path Sequence of sets of weighted ParaSwap routes. Each route specifies an exchange to use to convert a given (exact) amount of a given
-     * source token into a given (minimum) amount of a given destination token. The path is a sequence of sets of weighted routes where the destination
-     * token of a set of weighted routes matches the source token of the next set of weighted routes in the path.
+     * @param _path Sequence of sets of weighted ParaSwap routes. Each route specifies an exchange to use to convert a given (exact) amount of
+     * a given source token into a given (minimum) amount of a given destination token. The path is a sequence of sets of weighted routes where
```
olivdb, line 85: Could this be breaking NatSpec? In any case, I find it a bit less readable than having each param description on a single line. Since solhint doesn't seem to be able to distinguish code length from comment length, should we maybe wrap NatSpec blocks that contain long lines within `/* solhint-disable */ ...   /* solhint-enable */` ?

contracts/modules/common/Utils.sol
```diff
@@ -35,13 +35,13 @@ library Utils {
...
         // for v we load 32 bytes ending with v (the first 31 come from s) then apply a mask
-        // solium-disable-next-line security/no-inline-assembly
+
         assembly {
             r := mload(add(_signatures, add(0x20,mul(0x41,_index))))
             s := mload(add(_signatures, add(0x40,mul(0x41,_index))))
             v := and(mload(add(_signatures, add(0x41,mul(0x41,_index)))), 0xff)
         }
-        require(v == 27 || v == 28); // solium-disable-line error-reason
+        require(v == 27 || v == 28);
```
olivdb, line 44: I would add `"reason-string": ["warn", {"maxLength":32}]` to the ruleset and add `solhint-disable-next-line reason-string` for that particular line as a reminder to revisit whether we have a good reason not to include a reason here :)

contracts/modules/common/Utils.sol
```diff
@@ -50,7 +50,7 @@ library Utils {
     */
     function functionPrefix(bytes memory _data) internal pure returns (bytes4 prefix) {
         require(_data.length >= 4, "RM: Invalid functionPrefix");
-        // solium-disable-next-line security/no-inline-assembly
+
```
olivdb, line 53: I'm a bit unsure about removing all the assembly warnings from the code. In some cases, they may serve as an explicit reminder to check whether some new Solidity syntax has come out that removes the need for assembly. (This is such a case by the way: we could replace the following assembly code with the new array slicing syntax - something for a future PR).

elenadimitrova: I've switched the `no-inline-assembly` rule to an error and explicitly ignored the existing 3 instances.

contracts/modules/common/Utils.sol
elenadimitrova: I'd even make it an error

contracts/modules/TokenExchanger.sol
elenadimitrova: I don't see why the line length shouldn't be valid for comments as well and I would want to avoid disabling solhint unless we really have to. Natspec does support multiline comments so shouldn't be a problem https://solidity.readthedocs.io/en/latest/natspec-format.html#documentation-example

.solhint.json
```diff
@@ -0,0 +1,14 @@
...
+  "plugins": [],
+    "rules": {
+      "compiler-version": ["off"],
+      "no-inline-assembly": ["error"],
+      "not-rely-on-time": ["off"],
+      "avoid-low-level-calls": ["off"],
+      "max-line-length": ["error", 150],
+      "func-param-name-mixedcase": "error",
+      "modifier-name-mixedcase": "error",
+      "reason-string": ["error", { "maxLength":60 }]
```
olivdb, line 12: I think a maxLength of 32 would be more useful because, if I remember correctly, reasons with 32 characters or less are stored as a single bytes32. So it'd bee nice to be notified when a reason needs to be shortened to 32 characters.

contracts/modules/common/Utils.sol
olivdb: There are a few more usages of `assembly` (in BaseModule, GuardianUtils, Utils).
Also, I think you forgot to include the `solhint-disable-next-line reason-string` for the `require(v == 27 || v == 28);` in Utils.

.solhint.json
elenadimitrova: Okay I've demoted it to a warning and switched the `maxLenth`

contracts/modules/common/Utils.sol
elenadimitrova: I've switched missing or too long reason strings to warnings now as per https://github.com/argentlabs/argent-contracts/pull/137#discussion_r457459171

elenadimitrova: There was also an issue running solhint in subfolders more than 1 level deep, the ** wildcard isn't working for some reason, I tried a couple of versions of `node` to check if it's that but without success. To fix it I've added a second run of solhint explicitly for the second level of folders. Obviously this won't work if we ever have more than 2 layers of folders but I can't see that happening.

PR #138, opened 2020-07-16
Change and reset daily limit with guardians.

juniset:
We add 2 new methods to the `ApprovedTransfer` to simplify the management of the daily limit:
- `changeLimit` to change the current value of the daily limit immediately when approved by guardians
- `resetLimit` to reset the daily consumption to 0 immediately when approved by guardians.

We also make the following changes:
- Every action on the `ApprovedTransfer` module resets the daily consumption to 0 (calls `resetLimit` internally).
- On the `TransferManager` decreasing the daily limit becomes immediate, while increasing the limit remains pending for the duration of the security period.

### DIFF
--- a/contracts/modules/ApprovedTransfer.sol
+++ b/contracts/modules/ApprovedTransfer.sol
@@ -15,21 +15,34 @@
 
 // SPDX-License-Identifier: GPL-3.0-only
 pragma solidity ^0.6.10;
+pragma experimental ABIEncoderV2;
 
 import "./common/Utils.sol";
+import "./common/LimitUtils.sol";
 import "./common/BaseTransfer.sol";
+import "../infrastructure/storage/ILimitStorage.sol";
 
 /**
  * @title ApprovedTransfer
- * @dev Module to transfer tokens (ETH or ERC20) with the approval of guardians.
- * @author Julien Niset - <julien@argent.im>
+ * @dev Module to transfer tokens (ETH or ERC20) or call third-party contracts with the approval of guardians.
+ * @author Julien Niset - <julien@argent.xyz>
  */
 contract ApprovedTransfer is BaseTransfer {
 
     bytes32 constant NAME = "ApprovedTransfer";
 
-    constructor(IModuleRegistry _registry, IGuardianStorage _guardianStorage) BaseModule(_registry, _guardianStorage, NAME) public {
+    // The limit storage
+    ILimitStorage public limitStorage;
 
+    constructor(
+        IModuleRegistry _registry,
+        IGuardianStorage _guardianStorage,
+        ILimitStorage _limitStorage
+    )
+        BaseModule(_registry, _guardianStorage, NAME)
+        public
+    {
+        limitStorage = _limitStorage;
     }
 
     /**
@@ -52,6 +65,7 @@ contract ApprovedTransfer is BaseTransfer {
         onlyWhenUnlocked(_wallet)
     {
         doTransfer(_wallet, _token, _to, _amount, _data);
+        doResetLimit(_wallet);
     }
 
     /**
@@ -73,6 +87,7 @@ contract ApprovedTransfer is BaseTransfer {
         onlyAuthorisedContractCall(_wallet, _contract)
     {
         doCallContract(_wallet, _contract, _value, _data);
+        doResetLimit(_wallet);
     }
 
     /**
@@ -100,6 +115,27 @@ contract ApprovedTransfer is BaseTransfer {
         onlyAuthorisedContractCall(_wallet, _contract)
     {
         doApproveTokenAndCallContract(_wallet, _token, _spender, _amount, _contract, _data);
+        doResetLimit(_wallet);
+    }
+
+    /**
+     * @dev Changes the daily limit. The change is immediate.
+     * @param _wallet The target wallet.
+     * @param _newLimit The new limit.
+     */
+    function changeLimit(address _wallet, uint256 _newLimit) external onlyWalletModule(_wallet) onlyWhenUnlocked(_wallet) {
+        limitStorage.setLimit(_wallet, ILimitStorage.Limit(LimitUtils.safe128(_newLimit), uint128(0), uint64(0)));
+        doResetLimit(_wallet);
+        // solium-disable-next-line security/no-block-members
+        emit LimitChanged(_wallet, _newLimit, uint64(now));
+    }
+
+    /**
+    * @dev Resets the daily consumtion, i.e. DailySpent.alreadySpent = 0 and DailySpent.periodEnd = now.
+    * @param _wallet The target wallet.
+    */
+    function resetLimit(address _wallet) external onlyWalletModule(_wallet) onlyWhenUnlocked(_wallet) {
+        doResetLimit(_wallet);
     }
 
     /**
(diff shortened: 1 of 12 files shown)


contracts/modules/common/LimitUtils.sol
```diff
@@ -56,14 +56,15 @@ library LimitUtils {
...
         ILimitStorage.Limit memory limit = _lStorage.getLimit(_wallet);
-        // solium-disable-next-line security/no-block-members
         uint256 currentLimit = currentLimit(limit);
-        ILimitStorage.Limit memory newLimit = ILimitStorage.Limit(
-            safe128(currentLimit),
-            safe128(_targetLimit),
+        ILimitStorage.Limit memory newLimit;
+        if (_targetLimit <= currentLimit) {
+            // solium-disable-next-line security/no-block-members
+            newLimit = ILimitStorage.Limit(safe128(_targetLimit), uint128(0), safe64(now));
```
elenadimitrova, line 63: It is not better to set the `Limit.changeAfter` property to `0` instead of `now`? Both work logic wise but it seemed more natural when the `Limit.pending` is `0` for `Limit.changeAfter` to be `0` as well.
Same as this logic https://github.com/argentlabs/argent-contracts/pull/138/files#diff-252645bba53001a15e101c0e5845da2eR127

contracts/modules/ApprovedTransfer.sol
```diff
@@ -113,4 +149,13 @@ contract ApprovedTransfer is BaseTransfer {
         uint numberOfSignatures = 1 + Utils.ceil(guardianStorage.guardianCount(_wallet), 2);
         return (numberOfSignatures, OwnerSignature.Required);
     }
+
+    /**
+    * @dev Helper to Reset the daily consumption.
+    * @param _wallet The target wallet.
+    */
+    function doResetLimit(address _wallet) internal {
```
elenadimitrova, line 157: This is probably better placed in the `LimitUtils` ?

contracts/modules/ApprovedTransfer.sol
```diff
@@ -100,6 +115,27 @@ contract ApprovedTransfer is BaseTransfer {
...
+        doResetLimit(_wallet);
+        // solium-disable-next-line security/no-block-members
+        emit LimitChanged(_wallet, _newLimit, uint64(now));
+    }
+
+    /**
+    * @dev Resets the daily consumtion, i.e. DailySpent.alreadySpent = 0 and DailySpent.periodEnd = now.
+    * @param _wallet The target wallet.
+    */
+    function resetLimit(address _wallet) external onlyWalletModule(_wallet) onlyWhenUnlocked(_wallet) {
```
elenadimitrova, line 137: Please rename as it's confusing the "daily spent" with "daily limit" which are two different values in limit storage. Maybe call it resetDailySpent?

test/approvedTransfer.js
```diff
@@ -36,30 +38,44 @@ describe("Approved Transfer", function () {
...
     relayerModule = await deployer.deploy(RelayerModule, {},
       registry.contractAddress,
       guardianStorage.contractAddress,
-      ethers.constants.AddressZero,
+      limitStorage.contractAddress,
       ethers.constants.AddressZero);
     manager.setRelayerModule(relayerModule);
     walletImplementation = await deployer.deploy(BaseWallet);
+
+    limitModule = await deployer.deploy(TestLimitModule, {}, registry.contractAddress, guardianStorage.contractAddress, limitStorage.contractAddress);
```
elenadimitrova, line 65: We should always try to test using the actual contracts and not sure we need this `TestLimitModule` implementation. Where it's used here it's for `getDailySpent` and `getLimit` which we can use `LimitStorage` for and also setting the Wallet's limit and daily spent, which we can get in the wallet's initialisation. Can we remove the `TestLimitModule` altogether?

contracts/modules/common/BaseTransfer.sol
```diff
@@ -40,6 +40,8 @@ abstract contract BaseTransfer is BaseModule {
         uint256 amountSpent,
         bytes data
     );
+    event LimitChanged(address indexed wallet, uint indexed newLimit, uint64 indexed startAfter);
```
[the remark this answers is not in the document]
juniset: Done.

contracts/modules/common/LimitUtils.sol
juniset: Agreed. I hesitated between both options but the one you suggest is more consistent.

juniset: So actually the logic in https://github.com/argentlabs/argent-contracts/pull/138/files#diff-252645bba53001a15e101c0e5845da2eR127 is not consistent. I will update it so that when the limit is changed immediately the `changeAfter` parameter is always set to `now`.

juniset: Note that if we set `limit.changeAfter` to `now` then we must also set `limit.pending` to `_targetLimit` if we want `LimitUtils.currentLimit` to return the correct value.

contracts/modules/ApprovedTransfer.sol
juniset: Done

contracts/modules/ApprovedTransfer.sol
juniset: Done.

test/approvedTransfer.js
juniset: I agree on the principle but here the `TestLimitModule` is mostly introduced as a mean to initialise the limit and the dailySpent to non-zero values which is non trivial. The other alternative solution I see is adding an instance of the `TransferManager` and making a small transfer to update the dailySpent... but I find that option more complicated and equally unclean (testing the ApprovedTransfer becomes dependant of the correct behaviour of the TransferManager).

Unless I'm missing something? You mention wallet initialisation. How could we use that to set the limit and dailySpent to non-zero values?

contracts/modules/common/LimitUtils.sol
olivdb: > Note that if we set limit.changeAfter to now then we must also set limit.pending to _targetLimit if we want LimitUtils.currentLimit to return the correct value.

I don't think that's the case: the `if` in `currentLimit()` will return `false` since `now < now` is `false`. So it seems we could set `limit.pending` to `0`

contracts/modules/ApprovedTransfer.sol
```diff
@@ -100,6 +115,31 @@ contract ApprovedTransfer is BaseTransfer {
...
+     * @dev Changes the daily limit. The change is immediate.
+     * @param _wallet The target wallet.
+     * @param _newLimit The new limit.
+     */
+    function changeLimit(address _wallet, uint256 _newLimit) external onlyWalletModule(_wallet) onlyWhenUnlocked(_wallet) {
+        uint128 targetLimit = LimitUtils.safe128(_newLimit);
+        // solium-disable-next-line security/no-block-members
+        ILimitStorage.Limit memory newLimit = ILimitStorage.Limit(targetLimit, targetLimit, LimitUtils.safe64(now));
+        // solium-disable-next-line security/no-block-members
+        ILimitStorage.DailySpent memory resetDailySpent = ILimitStorage.DailySpent(uint128(0), LimitUtils.safe64(now));
```
olivdb, line 131: I think `DailySpent.periodEnd` could be set to 0 (both here and in `LimitUtils.resetDailySpent()`) since, unlike `Limit.changeAfter`, it is not used in events. That would make the treatment of this param slightly different to the way `Limit.changeAfter` is set but considering the number of times `resetDailySpent()` can be called, I think the gas savings would be worth this small inconsistency IMO.

juniset: Changed.

PR #140, opened 2020-07-20
Adding approveWethAndCallContract

olivdb:
Adding methods `approveWethAndCallContract` to `TransferManager` and `ApprovedTransfer` which:
- wrap ETH into WETH
- approve WETH
- call a contract

### DIFF
--- a/contracts/modules/common/BaseTransfer.sol
+++ b/contracts/modules/common/BaseTransfer.sol
@@ -26,6 +26,9 @@ import "./LimitUtils.sol";
  */
 abstract contract BaseTransfer is BaseModule {
 
+    // The address of the WETH token
+    address public wethToken;
+
     // *************** Events *************************** //
 
     event Transfer(address indexed wallet, address indexed token, uint256 indexed amount, address to, bytes data);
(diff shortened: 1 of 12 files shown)


utils/config/ganache.json
```diff
@@ -4,23 +4,34 @@
...
     "securityWindow": 240,
     "feeRatio": 15,
     "defaultLimit": "1000000000000000000"
   },
-  "CryptoKitties": { "contract": "0x0000000000000000000000000000000000000000" },
+  "CryptoKitties": {
+    "contract": "0x0000000000000000000000000000000000000000"
+  },
   "defi": {
+    "weth": "0x0000000000000000000000000000000000000000",
```
elenadimitrova, line 34: Should be in the `config-schema.json` as well

contracts/modules/common/BaseTransfer.sol
```diff
@@ -150,4 +161,31 @@ abstract contract BaseTransfer is BaseModule {
...
+    function doApproveWethAndCallContract(
+        address _wallet,
+        address _spender,
+        uint256 _amount,
+        address _contract,
+        bytes memory _data
+    )
+        internal
+    {
+        uint256 wethBalance = ERC20(wethToken).balanceOf(_wallet);
```
elenadimitrova, line 183: I guess having this function abstracted here has its benefits but the downside is having to check the token balance for the wallet twice, firstly here then again in `doApproveTokenAndCallContract`. I think I slightly lean towards your current option though.

olivdb: Yes, I think getting rid of `doApproveWethAndCallContract` and having the ETH wrapping logic added to `doApproveTokenAndCallContract` would be reasonable but I kind of like to keep this "special case" separate.

PR #146, opened 2020-07-29
Мigrate to truffle

elenadimitrova:
Here we migrate the solution from `etherlime` to the `truffle` framework. Truffle is used for compilation (using docker), testing, deployment and its extensions framework additionally implements code coverage, gas cost testing and etherscan verification.

**Notable changes include**
- Using `bn.js` consistently for comparing and working with big numbers
- Network `ganache` is now named `development`, note that configuration file has changed from `ganache.json` to `development.json`
- The `deployer` role is now minimal and it's no longer required to deploy new instances of a contract or create references to existing contracts. e.g. `deployer.deploy(GuardianStorage)` becomes `GuardianStorage.new()` and `deployer.wrapDeployedContract(ENS, newConfig.ENS.ensRegistry)` becomes `ENS.at(newConfig.ENS.ensRegistry)`.
- `verboseWaitForTransaction` is no longer required as the transaction receipt is returned directly for each contract call. 
- `.contractAddress` becomes `.address`
- Contracts loading from outside the `build folder` is done via `@truffle\contract` until https://github.com/trufflesuite/truffle/issues/3436 is resolved

**Plugins used**
- Gas reports using `eth-gas-reporter` and CodeChecks
https://github.com/argentlabs/argent-contracts/pull/146/checks?check_run_id=1341683785

These reports differ to the [benchmark script results](https://circleci.com/api/v1.1/project/github/argentlabs/argent-contracts/1666/output/105/0?file=true&allocation-id=5fa3ca8cd3519a35493dcf80-0-build%2F4F68937A) in that gas consumption per method is given which doesn't take into consideration chained calls though Proxy and/or Relayer contracts (meta transactions). We could accept these instances as relayed transactions take on average the same amount of extra gas ~45,000 while the Proxy calls are relevant only for the BaseWallet calls which are minimal.
 
- Code coverage running via a plugin developed by @cgewecke in https://github.com/argentlabs/argent-contracts/pull/177 ref https://github.com/sc-forks/solidity-coverage/issues/559

- Etherscan verification using `truffle-plugin-verify` ref https://github.com/rkalis/truffle-plugin-verify/issues/46
This is done explicitly via the `./scripts/verify.js` script which verifies in EtherScan all contracts from the configuration file's "contracts" and "modules" sections for the test, staging and production networks.

**blockers**
- Async provider factory in truffle config https://github.com/trufflesuite/truffle/issues/1054
- Improve performance https://github.com/trufflesuite/truffle/issues/3509#issuecomment-723571229

**nice-to-haves**
- Upgrade to latest `truffle` https://github.com/trufflesuite/ganache-cli/issues/702
- Enable getting the `chainId` in `utilities.getChainId()` after https://github.com/trufflesuite/truffle/issues/2688 is resolved


package.json
```diff
@@ -6,29 +6,28 @@
...
...57e8d94dd7e1a29055646bafde7010f38a900f55bbd7647880faa6ee8, 100000000000000000000000000000000000000\" --account=\"0x2030b463177db2da82908ef90fa55ddfcef56e8183caf60db464bc398e736e6f, 100000000000000000000000000000000000000\" --account=\"0x62ecd49c4ccb41a70ad46532aed63cf815de15864bc415c87d507afd6a5e8da2, 100000000000000000000000000000000000000\" --account=\"0xf473040b1a83739a9c7cc1f5719fab0f5bf178f83314d98557c58aae1910e03a, 100000000000000000000000000000000000000\" --account=\"0x823d590ed2cb5e8493bb0efc834771c1cde36f9fc49b9fe3620ebd0754ad6ea2, 100000000000000000000000000000000000000\" --account=\"0xd6d710943471e4c37ceb787857e7a2b41ca57f9cb4307ee9a9b21436a8e709c3, 100000000000000000000000000000000000000\" --account=\"0x187bb12e927c1652377405f81d93ce948a593f7d66cfba383ee761858b05921a, 100000000000000000000000000000000000000\" --account=\"0xf41486fdb04505e7966c8720a353ed92ce0d6830f8a5e915fbde735106a06d25, 100000000000000000000000000000000000000\" --account=\"0x6ca40ba4cca775643398385022264c0c414da1abd21d08d9e7136796a520a543, 100000000000000000000000000000000000000\" --account=\"0xfac0bc9325ad342033afe956e83f0bf8f1e863c1c3e956bc75d66961fe4cd186, 100000000000000000000000000000000000000\"",
```
olivdb, line 21: Wouldn't it be easier to run ganache-cli in deterministic mode (flag -d) and use these accounts instead of importing (and exporting) etherlime accounts?

utils/utilities.js
```diff
@@ -138,12 +175,102 @@ module.exports = {
...
+  },
+
+  async getNonceForRelay() {
+    const block = await web3.eth.getBlockNumber();
+    const timestamp = new Date().getTime();
+    return `0x${ethers.utils.hexZeroPad(ethers.utils.hexlify(block), 16)
+      .slice(2)}${ethers.utils.hexZeroPad(ethers.utils.hexlify(timestamp), 16).slice(2)}`;
+  },
+
+  async assertRevert(promise, revertMessage) {
```
olivdb, line 253: Is there a benefit in using our own function here instead of the one from `truffle-assertions` ?(https://github.com/rkalis/truffle-assertions/blob/master/index.js#L146)

package.json
elenadimitrova: good point. updated

utils/utilities.js
elenadimitrova: I hadn't heard of `truffle-assertions`. Works well to replace the utils method. Updated.

elenadimitrova: Updated the `assertRevert` usage with `truffle-assertions.reverts`. I experimented with `eventEmitted` as well but that still isn't able to decode nested events (from our RelayerManager) but this missing feature seems to be tracked here https://github.com/rkalis/truffle-assertions/issues/6

deployment/6_register_modules.js
```diff
@@ -1,115 +1,109 @@
...
+  const TransferManagerWrapper = await TransferManager.new(config.modules.TransferManager);
+  const TokenExchangerWrapper = await TokenExchanger.new(config.modules.TokenExchanger);
+  const NftTransferWrapper = await NftTransfer.new(config.modules.NftTransfer);
+  const CompoundManagerWrapper = await CompoundManager.new(config.modules.CompoundManager);
+  const MakerV2ManagerWrapper = await MakerV2Manager.new(config.modules.MakerV2Manager);
+  const RelayerManagerWrapper = await RelayerManager.new(config.modules.RelayerManager);
+  const VersionManagerWrapper = await VersionManager.new(config.modules.VersionManager);
+
+  const ModuleRegistryWrapper = await ModuleRegistry.new(config.contracts.ModuleRegistry);
+  const MultiSigWrapper = await MultiSig.new(config.contracts.MultiSigWallet);
```
juniset, line 48: Shouldn't it be `ContractWrapper = await Contract.at(config...)` instead of using `ContractWrapper = await Contract.new(config...)`?

elenadimitrova: Yes, thanks, updated.

juniset (approved):

truffle-config.base.js
```diff
@@ -20,18 +20,16 @@
...
 
-async function getKeys() {
-  const accounts = await web3.eth.getAccounts();
-  const deploymentAccount = accounts[0];
-  const manager = new DeployManager(deploymentAccount);
-  await manager.setup();
-  const { pkey, infuraKey } = manager;
-  return (pkey, infuraKey);
+function getKeys() {
+  // NOTE: While https://github.com/trufflesuite/truffle/issues/1054 is implemented we are using a temporary fix
```
olivdb, line 29: FYI we used to have a custom `AWSWalletProvider` to solve that issue https://github.com/argentlabs/solidity/blob/dev/truffle-awswallet-provider/index.js

elenadimitrova: I'll add a card to investigate that if the truffle fix is delayed

PR #149, opened 2020-08-11
Addressing Trail of Bits' audit

olivdb:
Addresses the first part of issues and improvements found in the Trail of Bits audit:

1 and 6. Add an `isTradable` flag to tokens in the `TokenPriceStorage` contract and only allow trades with those "verified" tokens. `TokenPriceProvider` is updated to store the following for each token:
```
        uint184 cachedPrice;
        uint64 updatedAt;
        bool isTradable;
```
A particular token price can be changed by a manager but not more frequently than every `minPriceUpdatePeriod` which is changeable by the multisig (owner).
A token can be set to non-tradable at any time by the multisig or a manager
A token can be set to tradable at any time by the multisig

Confirmed with the backend team that 1) the backend web3 lib supports uint184 and 2) all current prices would comfortably fit in a uint184 (our biggest price uses 92 bits)

.3. Removing auto conversion of ETH into DAI when repaying Maker loan
Instead of adding the ETH-DAI purchase to the daily limit we will simply remove that functionality and force users to have the required amount of DAI to repay a loan.

.4. Add `chainID` opcode in the signature schema of meta transactions

.9. Allows the removal of a pending address whitelist request.

.10. Changes for loop counter from `uint8` to `uint256` to limit an issue for wallets with more than 510 guardians.

.11. natspec documentation fixes

.12. Relayed transactions to `ApprovedTransfer` module throw for wallets with no guardians

.13. Whitelisting contract instead of proxy (spender)

.14. isERC721 replaced with a check that the contract is not a known ERC20

.16. Refund transfers made to fail if transfer method returns false

.19. Failed transfers made to revert the subcall

.25. Added `DailyLimitMigrated`, `DailyLimitDisabled` and `LoanAcquired` events. 
todo:

### DIFF
--- a/contracts/infrastructure/storage/TokenPriceStorage.sol
+++ b/contracts/infrastructure/storage/TokenPriceStorage.sol
@@ -23,7 +23,7 @@ import "../base/Managed.sol";
 /**
  * @title TokenPriceStorage
  * @notice Contract storing the token prices.
- * @notice Note that prices stored here = price per token * 10^(36-token decimals)
+ * @notice Note that prices stored here = price per token * 10^(18-token decimals)
  * The contract only defines basic setters and getters with no logic.
  * Only managers of this contract can modify its state.
  */
--- a/contracts/infrastructure/storage/TokenPriceStorage.sol
+++ b/contracts/infrastructure/storage/TokenPriceStorage.sol
@@ -50,13 +50,13 @@ contract TokenPriceStorage is ITokenPriceStorage, Storage, Managed {
     }
     function getPriceForTokenList(address[] calldata _tokens) external view returns (uint184[] memory _prices) {
         _prices = new uint184[](_tokens.length);
-        for (uint i = 0; i < _tokens.length; i++) {
+        for (uint256 i = 0; i < _tokens.length; i++) {
             _prices[i] = tokenInfo[_tokens[i]].cachedPrice;
         }
     }
     function getTradableForTokenList(address[] calldata _tokens) external view returns (bool[] memory _tradable) {
         _tradable = new bool[](_tokens.length);
-        for (uint i = 0; i < _tokens.length; i++) {
+        for (uint256 i = 0; i < _tokens.length; i++) {
             _tradable[i] = tokenInfo[_tokens[i]].isTradable;
         }
     }
@@ -67,7 +67,8 @@ contract TokenPriceStorage is ITokenPriceStorage, Storage, Managed {
         minPriceUpdatePeriod = _newPeriod;
     }
     function setPriceForTokenList(address[] calldata _tokens, uint184[] calldata _prices) external onlyManager {
-        for (uint16 i = 0; i < _tokens.length; i++) {
+        require(_tokens.length == _prices.length, "TPS: Array length mismatch");
+        for (uint i = 0; i < _tokens.length; i++) {
             uint64 updatedAt = tokenInfo[_tokens[i]].updatedAt;
             require(updatedAt == 0 || block.timestamp >= updatedAt + minPriceUpdatePeriod, "TPS: Price updated too early");
             tokenInfo[_tokens[i]].cachedPrice = _prices[i];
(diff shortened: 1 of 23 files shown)


contracts/infrastructure/storage/TokenPriceStorage.sol
```diff
@@ -23,32 +23,61 @@ import "../base/Managed.sol";
...
     }
 
-    function setPriceForTokenList(address[] calldata _tokens, uint256[] calldata _prices) external override onlyManager {
+    // Setters
+    
+    function setMinPriceUpdatePeriod(uint256 _newPeriod) external onlyOwner {
+        minPriceUpdatePeriod = _newPeriod;
+    }
+    function setPriceForTokenList(address[] calldata _tokens, uint184[] calldata _prices) external onlyManager {
         for (uint16 i = 0; i < _tokens.length; i++) {
```
juniset, line 70: We should probably check that `_tokens.length == _prices.length` and use an `uint256` for the loop.

contracts/infrastructure/storage/TokenPriceStorage.sol
```diff
@@ -23,32 +23,61 @@ import "../base/Managed.sol";
...
+            require(updatedAt == 0 || block.timestamp >= updatedAt + minPriceUpdatePeriod, "TPS: Price updated too early");
+            tokenInfo[_tokens[i]].cachedPrice = _prices[i];
+            tokenInfo[_tokens[i]].updatedAt = uint64(block.timestamp);
         }
     }
-
-    function setPrice(address _token, uint256 _price) external override onlyManager {
-        cachedPrices[_token] = _price;
+    function setTradableForTokenList(address[] calldata _tokens, bool[] calldata _tradable) external {
+        for (uint16 i = 0; i < _tokens.length; i++) {
```
juniset, line 78: Same comment as above: we must check that the arrays have equal length and probably use a `uint256` for the loop.

Also, why are we preventing `managers` to make tokens tradable? I think it is ok to let managers make tokens tradable, and it will make it much easier operationally.

olivdb: If a manager could set a token to `tradable` then there would be a possibility for a compromised manager to immediately add a fake token and set `tradable` to `true` to bypass the daily limit and drain any wallet it controls. 

Since we decided to introduce `minPriceUpdatePeriod` to mitigate the power of a manager, I think it makes sense to also restrict a manager's ability to add tradable tokens.

contracts/modules/common/BaseTransfer.sol
```diff
@@ -77,7 +77,11 @@ abstract contract BaseTransfer is BaseModule {
             invokeWallet(_wallet, _to, _value, EMPTY_BYTES);
         } else {
             bytes memory methodData = abi.encodeWithSignature("transfer(address,uint256)", _to, _value);
-            invokeWallet(_wallet, _token, 0, methodData);
+            bytes memory transferSuccessBytes = invokeWallet(_wallet, _token, 0, methodData);
+            // Check transfer is successful, when `transfer` returns a success bool result
+            if (transferSuccessBytes.length > 0) {
+                require(abi.decode(transferSuccessBytes, (bool)), "RM: Refund transfer failed");
```
olivdb, line 83: You forgot to change the error reason here

contracts/infrastructure/storage/TokenPriceStorage.sol
juniset: I personally think that the added (operational) complexity is not worth the risk. Having to go through the multisig every-time we want to add a new token will be a pain, while the risk of someone compromising a wallet AND compromising our backend is minimal.

contracts/modules/common/Utils.sol
```diff
@@ -42,7 +42,10 @@ library Utils {
             v := and(mload(add(_signatures, add(0x41,mul(0x41,_index)))), 0xff)
         }
         require(v == 27 || v == 28);
-        return ecrecover(_signedHash, v, r, s);
+
+        address recoveredAddress = ecrecover(_signedHash, v, r, s);
+        require(recoveredAddress != address(0), "Utils: ecrecover returned 0");
```
juniset, line 47: We should also include the check agains signature maleability.

contracts/infrastructure/storage/TokenPriceStorage.sol
juniset: Ok ignore my last comment, you are right.

contracts/modules/common/Utils.sol
elenadimitrova: I didn't include it as it's not in the recommendation due to the fact we don't allow signature re-use.

juniset (approved):

PR #151, opened 2020-08-24
New architecture using Features (L2 modules) calling Version manager (L1 module)

olivdb:
We introduce a new architecture to reduce the gas cost of a wallet creation and upgrade:
- Each wallet will only have one `VersionManager` module authorised.
- All previous modules (`RelayerManager`, `TransferManager`, etc) are now called "Features".
- The Argent multisig can add a new version to the VersionManager which consists of a bundle of Features.
- When a wallet is created, the `VersionManager` is added as the sole module for that wallet and a version number is selected for the wallet. The corresponding bundle of Features is thereby authorised for the wallet.
- Features only invoke the wallet (or change its owner) indirectly via the `VersionManager`. This is because the `BaseWallet` contract remains unchanged and this contract only allows a module to invoke the wallet (or change its owner).
- Likewise, write access by Features to storage contracts is done indirectly via the `VersionManager`. This is because a wallet only knows about modules (and not features) so storage contracts can ask a wallet if a calling module is authorised but could not do the same for a feature.
- The lock-related logic contained within `GuardianStorage` has been extracted into a dedicated `LockStorage`. This was required because of the indirect access by Features to storage contracts described above. This means that the lock-related logic still contained within `GuardianStorage` should be ignored.

Other change:
- The exchanges whitelisted to be used with the `TokenExchanger` are no longer specified in the constructor of the `TokenExchanger` but are instead read from a new `DexRegistry` controller by the multisig. This change was made so that when a new DEX gets supported by Paraswap, the DEX can be supported by the `TokenExchanger` without requiring a new upgrade of that feature. This new `DexRegistry` together with Paraswap's own whitelist, constitute two lines of defence to prevent exploiting the `TokenExchanger` with a malicious trade.

Gas cost comparison before and after this refactoring 
| Task                                          | Gas  before      | Gas  after     | Diff   |
|-----------------------------------------------|-----------|-----------|--------|
| Create a wallet without ENS (1 module)        | 311640    | -         | -      |
| Create a wallet without ENS (2 module)        | 340513    | -         | -      |
| Create a wallet without ENS (3 module)        | 369386    | -         | -      |
| Create a wallet without ENS (all modules)     | 689086    | 432999    | -256087 |
| Create a wallet with ENS (all modules)        | 930312    | 681534    | -248778 |
│ Upgrade all modules on a wallet               |    518050       | 85924     |   -432126     | 
| Request add second guardian (direct)          | 77806     | 72908     |  4898   |
| Confirm add second guardian (direct)          | 92581     | 99358     | 6777   |
| Lock wallet (direct)                          | 90401     |  93105    | 2704 |
| Unlock wallet (direct)                        | 55288     |  57423    | 2135 |
| Lock wallet (relayed)                         | 145031    | 146576    | 1545 |
| Unlock wallet (relayed)                       | 94947     |  96491   |  1544 |
| Execute recovery                              | 170080    | 176744   |6664|
| Change limit (direct)                         | 78235     |  85296   | 7061 |
| Change limit (relayed)                        | 122621    |  129565   | 6944  |
| ETH transfer, limit disabled (direct)         | 67178     |  74835   | 7657 |
| ETH transfer, limit disabled (relayed)        | 111753     |  119771   | 8018 |
| ETH small transfer (direct)                   | 80831     |   93838  | 13007 |
| ETH small transfer (relayed)                  | 140139    |  153384  | 13245 |
| ETH transfer to whitelisted account (direct)  | 59679     |   66573  | 6894 |
| ETH transfer to whitelisted account (relayed) | 104258     |  111497   | 7239 |
| ETH large transfer to untrusted account       | 95559     |  108804   | 13245 |
| ETH large transfer approval by one guardian   | 128842    | 142157   | 13315 |
| ETH large transfer approval by two guardians  | 141479    | 154765    | 13286 |


contracts/modules/common/BaseFeature.sol
```diff
@@ -0,0 +1,178 @@
...
+        registry = _registry;
+        lockStorage = _lockStorage;
+        versionManager = _versionManager;
+        emit FeatureCreated(_name);
+    }
+
+    /**
+    * @inheritdoc IFeature
+    */
+    function recoverToken(address _token) external override {
```
juniset, line 113: Tokens should be transferred to the `VersionManager` which would remove the dependency of the Features on the `ModuleRegistry`. We can then remove it from the constructor of all the features.

juniset: We will need to add 
```
    function recoverToken(address _token) external onlyOwner {
        uint total = ERC20(_token).balanceOf(address(this));
        ERC20(_token).transfer(msg.sender, total);
    }
```
to the `VersionManager`.

contracts/modules/common/IModule.sol
```diff
@@ -24,39 +24,9 @@ pragma solidity >=0.5.4 <0.7.0;
...
      * @param _wallet The wallet.
      */
     function init(address _wallet) external;
-
-    /**
-     * @notice Adds a module to a wallet. Cannot execute when wallet is locked (or under recovery)
-     * @param _wallet The target wallet.
-     * @param _module The modules to authorise.
-     */
-    function addModule(address _wallet, address _module) external;
```
juniset, line 52: I think we should keep that method as is it critical to have it if we want to update the `VersionManager`.

contracts/modules/VersionManager.sol
```diff
@@ -0,0 +1,250 @@
...
+    function init(address _wallet) public override(IModule, BaseFeature) onlyWallet(_wallet) {
+        doUpgradeWallet(_wallet, featuresToInit[lastVersion]);
+    }
+
+    /**
+     * @inheritdoc IFeature
+     */
+    function getRequiredSignatures(address _wallet, bytes calldata _data) external view override returns (uint256, OwnerSignature) {
+        bytes4 methodId = Utils.functionPrefix(_data);
+        if (methodId == UPGRADE_WALLET_PREFIX || methodId == ADD_MODULE_PREFIX) {
```
juniset, line 88: Maybe more explicit to write:
```
bytes4 methodId = Utils.functionPrefix(_data);
require(methodId == UPGRADE_WALLET_PREFIX || methodId == ADD_MODULE_PREFIX, "VM: Unauthorised method);
return (1, OwnerSignature.Required);
```

contracts/modules/common/BaseFeature.sol
olivdb: Done

contracts/modules/VersionManager.sol
olivdb: Agreed and done.

contracts/modules/VersionManager.sol
```diff
@@ -0,0 +1,250 @@
...
+        internal 
+    {
+        uint256 fromVersion = walletVersions[_wallet];
+        uint256 toVersion = lastVersion;
+        require(fromVersion < toVersion, "VM: Already on last version");
+        walletVersions[_wallet] = toVersion;
+
+        // Setup static call redirection
+        bytes4[] storage sigs = staticCallSignatures[toVersion];
+        for(uint256 i = 0; i < sigs.length; i++) {
```
juniset, line 243: Maybe we could save some gas by introducing a catchall static call for wallets with 1 module (i.e. new wallets) and provided that we slightly update the fallback method of `BaseWallet` : 
```
contract BaseWallet {

    fallback() external payable {
        if(msg.value > 0) {
            emit Received(msg.value, msg.sender, msg.data);
        } else {
            address module = enabled[msg.sig];
            if (module == address(0)) {
                module = enabled[CATCHALL]; // e.g. CATCHALL = 0xffffffff
            }
            require(authorised[module], "BW: must be an authorised module for static call");

            assembly {
                calldatacopy(0, 0, calldatasize())
                let result := staticcall(gas(), module, 0, calldatasize(), 0, 0)
                returndatacopy(0, 0, returndatasize())
                switch result
                case 0 {revert(0, returndatasize())}
                default {return (0, returndatasize())}
            }
        }
    }

}
```

juniset: Actually this wouldn't work. 

I'm keeping the comment so we can think more about it.

contracts/infrastructure/storage/Storage.sol
```diff
@@ -17,19 +17,31 @@
...
+    }
+
+    /**
+     * @notice Throws if the caller is not an authorised feature.
+     */
+    modifier onlyFeature(address _wallet) {
+        require(
+            IFeature(msg.sender).isFeatureAuthorisedInVersionManager(_wallet, msg.sender), 
+            "S: must be an authorized feature to call this method"
+        );
```
juniset, line 44: This is not ok since we do the check with msg.sender itself. Anybody can modify the storage by creating a custom contract with a `isFeatureAuthorisedInVersionManager` method that always returns `true` .

juniset: Unfortunately I think the only solution is to only let modules modify storages and have features proxy through the `VersionManager`. This will make the model cleaner by treating all the storages (old and new) coherently, but it will add an extra `call` (~2000 gas) to every write.

contracts/modules/VersionManager.sol
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    /**
+     * @notice Upgrade a wallet to the latest version.
+     * @dev It's cheaper to pass features to init as calldata than reading them from storage
+     * @param _wallet the wallet to upgrrade
+     * @param _featuresToInit the subset of features that need to be initialized for the walleet
+     */
+    function upgradeWallet(address _wallet, address[] calldata _featuresToInit) external onlyWalletOwnerOrFeature(_wallet) onlyWhenUnlocked(_wallet) {
+        doUpgradeWallet(_wallet, _featuresToInit);
```
juniset, line 167: I understand that we pass `_featuresToInit` in calldata to avoid a storage read and reduce the gas cost but I think it is error prone. If you pass e.g. an incomplete list of features to init then the wallet will be in the target version but in an inconsistent state (i.e. no all features initialised). It is much safer to load the list of features from `featuresToInit` since we already check in `doUpgradeWallet` which ones need to be initialised. Given that a SLOAD is only 200 gas/word I don't think the cost saving is worth the fragility.

olivdb: An SLOAD costs 800 gas since Istanbul and is likely to increase even more (see https://notes.ethereum.org/@vbuterin/BkrNbeAfD). A middle ground solution could be to check that the calldata array is the same size as the storage array?

contracts/modules/common/BaseFeature.sol
```diff
@@ -0,0 +1,173 @@
...
+        require(versionManager.isFeatureAuthorised(_wallet, msg.sender), "BF: must be a wallet feature");
+        _;
+    }
+
+    /**
+     * @notice Throws if the sender is not the owner of the target wallet or the feature itself.
+     */
+    modifier onlyWalletOwnerOrFeature(address _wallet) {
+        // Wrapping in an internal method reduces deployment cost by avoiding duplication of inlined code
+        verifyOwnerOrAuthorisedFeature(_wallet, msg.sender);
```
juniset, line 91: This is just an idea but the `Feature` part of this modifier is only there to let the `RelayerManager` call the other features. Other than that we don't want features to call each other and we know it is a security risk to have such a general check while we only want to address a particular use case.

Now that we store features in the singleton `VersionManager` we could add information about which feature can call the other features. This would make the wallet more secure at no extra gas cost, address TOB issue #23 without creating immutable dependencies between features.

contracts/modules/VersionManager.sol
olivdb: Or we could store a hash of the `featuresToInit` array in `addVersion` and compare this hash to the hash of the calldata `_featuresToInit` in `upgradeWallet()` to at least compress the cost to a single SLOAD regardless of the number of features.

olivdb: Anyway for now I'll just remove that calldata parameter to keep things simple.

contracts/modules/VersionManager.sol
```diff
@@ -234,6 +239,41 @@ contract VersionManager is IVersionManager, IModule, BaseFeature, Owned {
...
+        address _wallet,
+        ILimitStorage.Limit memory _limit,
+        ILimitStorage.DailySpent memory _dailySpent
+    ) 
+        external
+        override
+        onlyFeature(_wallet) 
+    {
+        limitStorage.setLimitAndDailySpent(_wallet, _limit, _dailySpent);
+    }
```
juniset, line 276: Now that we are routing all the writes to the different storages in the `VersionManager` I think we should add a generic `invokeStorage` method that will check that the target is legit. Otherwise we cannot add new storages in the future without doing a (costly) `VersionManager` upgrade in every wallet.

Something like what you had before:
```
function invokeStorage(address _wallet, address _storage, bytes _data) external onlyFeature(_wallet) {
   require(isStorage[_storage], "VM: invalid storage invoked");
   _storage.call(_data);
}
```

contracts/modules/common/BaseFeature.sol
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    /**
+     * @notice Helper method to invoke a wallet.
+     * @param _wallet The target wallet.
+     * @param _to The target address for the transaction.
+     * @param _value The value of the transaction.
+     * @param _data The data of the transaction.
+     */
+    function checkAuthorisedFeatureAndInvokeWallet(address _wallet, address _to, uint256 _value, bytes memory _data)
```
juniset, line 166: Why did we change the name from `invokeWallet` to `checkAuthorisedFeatureAndInvokeWallet`? It adds a lot of renaming in all the features and the method is just invoking the wallet.

contracts/modules/TransferManager.sol
```diff
@@ -249,11 +240,11 @@ contract TransferManager is OnlyOwnerModule, BaseTransfer {
         bytes calldata _data
     )
         external
-        onlyWalletOwnerOrModule(_wallet)
+        onlyWalletOwnerOrFeature(_wallet)
         onlyWhenUnlocked(_wallet)
         onlyAuthorisedContractCall(_wallet, _contract)
```
juniset, line 245: We need to update this modifier to check that we are not calling another Feature.

contracts/modules/common/BaseFeature.sol
olivdb: To avoid a name clash with  `VersionManager.invokeWallet` which is also a feature. I could rename `BaseFeature.checkAuthorisedFeatureAndInvokeWallet` back to `BaseFeature.invokeWallet` and rename `VersionManager.invokeWallet` to something else (e.g. `VersionManager.checkAuthorisedFeatureAndInvokeWallet` unless you can think of a better name)

contracts/modules/VersionManager.sol
```diff
@@ -0,0 +1,305 @@
...
+
+    event VersionAdded(uint256 _version);
+    event WalletUpgraded(address _wallet, uint256 _version);
+
+    // The Guardian Storage
+    IGuardianStorage private guardianStorage;
+    // The Transfer Storage
+    ITransferStorage private transferStorage;
+    // The Limit Storage
+    ILimitStorage private limitStorage;
```
juniset, line 65: No longer needed to keep the address of these 3 storages locally. They are only needed in the `isStorage` mapping.

contracts/modules/VersionManager.sol
```diff
@@ -0,0 +1,305 @@
...
+        ITransferStorage _transferStorage,
+        ILimitStorage _limitStorage
+    )
+        BaseFeature(_lockStorage, IVersionManager(address(this)), NAME)
+        public
+    {
+        registry = _registry;
+        guardianStorage = _guardianStorage;
+        transferStorage = _transferStorage;
+        limitStorage = _limitStorage;
```
juniset, line 91: Same as above, L89 to 91 can be deleted.

contracts/modules/common/BaseTransfer.sol
```diff
@@ -55,12 +55,17 @@ abstract contract BaseTransfer is BaseFeature {
...
-    * @notice Make sure a contract call is not trying to call a module, the wallet itself, or a supported ERC20.
+    * @notice Make sure a contract call is not trying to call a module, a feature, or the wallet itself.
     * @param _wallet The target wallet.
     * @param _contract The address of the contract.
      */
     modifier onlyAuthorisedContractCall(address _wallet, address _contract) {
-        require(_contract != _wallet && !IWallet(_wallet).authorised(_contract), "BT: Forbidden contract");
+        require(
+            _contract != _wallet && 
+            !IWallet(_wallet).authorised(_contract) &&
```
juniset, line 65: This call is obselete if we assume that the `VersionManager` is the only module since it is also a feature and will be included in `isFeatureAuthorised` . But I don't know if that assumption is worth saving ~3000 gas for a contract call. I'll let you decide.

contracts/infrastructure/DexRegistry.sol
```diff
@@ -0,0 +1,69 @@
...
+
+
+    /**
+     * @notice Add/Remove a DEX adapter to/from the whitelist.
+     * @param _dexes array of DEX adapters to add to (or remove from) the whitelist
+     * @param _authorised array where each entry is true to add the corresponding DEX to the whitelist, false to remove it
+     */
+    function setAuthorised(address[] calldata _dexes, bool[] calldata _authorised) external onlyOwner {
+        for(uint256 i = 0; i < _dexes.length; i++) {
+            require(isAuthorised[_dexes[i]] == !_authorised[i], "DR: DEX already (un)authorised");
```
juniset, line 44: Is this needed? Will be simpler operationally if it doesn't revert. We only care about the whitelist being in the correct state after the call, no?

contracts/modules/common/BaseTransfer.sol
olivdb: I agree and will remove the check. Though keep in mind that `SimpleUpgrader.init()` can be called then. That's not a problem as the call will fail (since the `SimpleUpgrader` will be unauthorised so the attempt to authorise `toEnable[i]` will revert).

contracts/infrastructure/DexRegistry.sol
olivdb: I added it so that the emitted events are correct and to catch incorrect inputs but I can make it so that already (un)authorised dexes are just skipped

contracts/infrastructure/storage/Storage.sol
```diff
@@ -17,19 +17,20 @@
 pragma solidity >=0.5.4 <0.7.0;
 
 import "../../wallet/IWallet.sol";
+import "../../modules/common/IFeature.sol";
```
juniset, line 20: This import is no longer needed.

juniset:
We should update the description of the PR to reflect that all storage contracts can only be written via the `VersionManager`.

contracts/modules/VersionManager.sol
```diff
@@ -0,0 +1,296 @@
...
+    mapping(address => uint256) public walletVersions; // [wallet] => [version]
+    // Features per version
+    mapping(address => mapping(uint256 => bool)) public isFeatureInVersion; // [feature][version] => bool
+    // Features requiring initialization for a wallet
+    mapping(uint256 => address[]) public featuresToInit; // [version] => [features]
+
+    // Supported static call signatures
+    mapping(uint256 => bytes4[]) public staticCallSignatures; // [version] => [sigs]
+    // Features executing static calls
+    mapping(uint256 => mapping(bytes4 => address)) public staticCallExecutors; // [version][sig] => [feature]
```
juniset, line 52: Given the current storage structure it is impossible to get the list of features from a version. If I remember correctly this was requested by the clients. We should check with @th3m477 if it is still the case.

contracts/infrastructure/storage/ILockStorage.sol
```diff
@@ -0,0 +1,27 @@
...
+pragma solidity >=0.5.4 <0.7.0;
+
+interface ILockStorage {
+    function isLocked(address _wallet) external view returns (bool);
+
+    function getLock(address _wallet) external view returns (uint256);
+
+    function getLocker(address _wallet) external view returns (address);
+
+    function setLock(address _wallet, address _locker, uint256 _releaseAfter) external;
```
elenadimitrova, line 26: If the above will not be used on the GuardianStorage, can we remove them from its interface? I think all modules only go via the interface rather than the implementation contract.

elenadimitrova (commented):
I understand this refactoring is likely going to make it in the codebase but for the record I have to say I disagree with the complexity it adds with adding a fifth `call` to the call chain. Although this saves ~234,000 gas from wallet creation, it adds on average ~10,000 gas per transaction which in not too long will level off this saving.

contracts/modules/VersionManager.sol
```diff
@@ -0,0 +1,296 @@
...
+        uint total = ERC20(_token).balanceOf(address(this));
+        _token.call(abi.encodeWithSelector(ERC20(_token).transfer.selector, msg.sender, total));
+    }
+
+    /**
+     * @notice Lets the owner add a new version, i.e. a new bundle of features
+     * @param _features the list of features included in the new version
+     * @param _featuresToInit the subset of features that need to be initialized for a wallet
+     */
+    function addVersion(address[] calldata _features, address[] calldata _featuresToInit) external onlyOwner {
```
juniset, line 114: I am still a bit worried about this `_featuresToInit` parameter. I understand that it saves some gas on creation/upgrade (~2000 gas per feature) but it is error prone and could create/update wallets to an inconsistent state if we forget to set it correctly. 

Maybe we should update the deployment script to better automate the generation of the `_featuresToInit` array by e.g. adding features to the array while they are created.

contracts/modules/VersionManager.sol
olivdb: I added the list of features in a version to the `VersionAdded` event: `event VersionAdded(uint256 _version, address[] _features)`. Will check with clients if that's acceptable.

contracts/infrastructure_0.5/storage/TransferStorage.sol
```diff
@@ -36,17 +36,17 @@ contract TransferStorage is ITransferStorage, Storage {
      * @notice Lets an authorised module add or remove an account from the whitelist of a wallet.
      * @param _wallet The target wallet.
      * @param _target The account to add/remove.
-     * @param _value True for addition, false for revokation.
+     * @param _whitelistAfter The epoch time at which an account starts to be whitelisted, or zero if the account is not whitelisted
      */
-    function setWhitelist(address _wallet, address _target, uint256 _value) external onlyModule(_wallet) {
-        whitelist[_wallet][_target] = _value;
+    function setWhitelist(address _wallet, address _target, uint256 _whitelistAfter) external onlyModule(_wallet) {
+        whitelist[_wallet][_target] = _whitelistAfter;
```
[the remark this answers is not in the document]
olivdb: That was just a fix of the Natspec (which was incorrect) and a renaming of `_value` -> `_whitelistAfter` to be consistent with its usage in `TransferManager`. I'll revert the variable renaming and keep only the Natspec fix.

deployment/999_benchmark.js
```diff
@@ -155,6 +156,166 @@ class Benchmark {
...
+    await registry.registerModule(newLockManagerWrapper.contractAddress, formatBytes32String("newLockManagerWrapper"));
+    await registry.registerModule(newRecoveryManagerWrapper.contractAddress, formatBytes32String("newRecoveryManagerWrapper"));
+    await registry.registerModule(newApprovedTransferWrapper.contractAddress, formatBytes32String("newApprovedTransferWrapper"));
+    await registry.registerModule(newTransferManagerWrapper.contractAddress, formatBytes32String("newTransferManagerWrapper"));
+    await registry.registerModule(newTokenExchangerWrapper.contractAddress, formatBytes32String("newTokenExchangerWrapper"));
+    await registry.registerModule(newNftTransferWrapper.contractAddress, formatBytes32String("newNftTransferWrapper"));
+    await registry.registerModule(newCompoundManagerWrapper.contractAddress, formatBytes32String("newCompoundManagerWrapper"));
+    await registry.registerModule(newMakerV2ManagerWrapper.contractAddress, formatBytes32String("newMakerV2ManagerWrapper"));
+    await registry.registerModule(newRelayerModuleWrapper.contractAddress, formatBytes32String("newRelayerModuleWrapper"));
```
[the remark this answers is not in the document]
olivdb: This was removed in a later commit

contracts/infrastructure/storage/Storage.sol
```diff
@@ -17,19 +17,20 @@
...
+ * @author Julien Niset - <julien@argent.xyz>, Olivier VDB - <olivier@argent.xyz>
  */
 contract Storage {
 
     /**
      * @notice Throws if the caller is not an authorised module.
      */
     modifier onlyModule(address _wallet) {
-        require(IWallet(_wallet).authorised(msg.sender), "TS: must be an authorized module to call this method");
+        require(IWallet(_wallet).authorised(msg.sender), "S: must be an authorized module to call this method");
```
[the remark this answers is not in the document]
olivdb: We are deploying a new `LockStorage` and `LimitStorage` that inherit from `Storage`. But `GuardianStorage` and `TransferStorage`, which also inherit from `Storage`, are not redeployed. Fixing this typo may not be worth the complexity of having our storage contracts inherit two slightly different `Storage` contracts, so I'll revert the fix.

contracts/infrastructure/storage/ILockStorage.sol
olivdb: Unfortunately, for some reason, removing these methods from `IGuardianStorage` changes the deployed bytecode of `GuardianStorage`, which we probably don't want to do since `GuardianStorage` is not redeployed.

contracts/modules/VersionManager.sol
olivdb: For the record, according to benchmarks, using `_featuresToInit` saves:
- 7344 gas per new feature during a wallet upgrade
- 6627 gas per feature during a wallet creation

olivdb: To be a bit less error-prone, we could change the deployment script to list the features that do NOT require init instead of the those the DO require init.

```
  // Add Features to Version Manager
  const newFeatures = [
    GuardianManagerWrapper.contractAddress,
    LockManagerWrapper.contractAddress,
    RecoveryManagerWrapper.contractAddress,
    ApprovedTransferWrapper.contractAddress,
    TransferManagerWrapper.contractAddress,
    TokenExchangerWrapper.contractAddress,
    NftTransferWrapper.contractAddress,
    CompoundManagerWrapper.contractAddress,
    MakerV2ManagerWrapper.contractAddress,
    RelayerManagerWrapper.contractAddress,
  ];
  const newFeaturesWithNoInit = [ // all features except the TransferManager
    GuardianManagerWrapper.contractAddress,
    LockManagerWrapper.contractAddress,
    RecoveryManagerWrapper.contractAddress,
    ApprovedTransferWrapper.contractAddress,
    TokenExchangerWrapper.contractAddress,
    NftTransferWrapper.contractAddress,
    CompoundManagerWrapper.contractAddress,
    MakerV2ManagerWrapper.contractAddress,
    RelayerManagerWrapper.contractAddress,
  ];
  const newFeatureToInit = newFeatures.filter((f) => !newFeaturesWithNoInit.includes(f));
  const VersionManagerAddVersionTx = await VersionManagerWrapper.contract.addVersion(newFeatures, newFeatureToInit, { gasPrice });
  await VersionManagerWrapper.verboseWaitForTransaction(VersionManagerAddVersionTx, "Adding New Version");

```

juniset: I'm surprised by the cost of calling init (~7k). Does this include the `TransferManager` or is this number valid for `init` methods only emitting an event? I would have expected the cost to be ~3k (1 call + 1 event).

Which reminds me that clients where using the `FeatureInitialised` (or `ModuleInitialised`) event to detect the features of a wallet. We need to make sure that it is no longer the case and that they are now using the `VersionAdded` event instead.

olivdb: I had checked with the clients that `ModuleInitialised` was in fact not used.

contracts/modules/VersionManager.sol
```diff
@@ -188,16 +202,17 @@ contract VersionManager is IVersionManager, IModule, BaseFeature, Owned {
...
      */
     function init(address _wallet) public override(IModule, BaseFeature) onlyWallet(_wallet) {
-        doUpgradeWallet(_wallet);
+        doUpgradeWallet(_wallet, newWalletVersion);
     }
 
     /**
-     * @notice Upgrade a wallet to the latest version.
+     * @notice Upgrade a wallet to a new version.
      * @dev It's cheaper to pass features to init as calldata than reading them from storage
```
juniset, line 210: The comment is no longer correct.

contracts/modules/VersionManager.sol
olivdb: It's the cost per feature for features with an `init` method that only emits an event. Also surprised by the 6-7k cost to be honeest. Roughly speaking that should be the cost of 3 reads ("feature address" + "is feature already initialized" + "is caller the version manager" = 3*800) + 1 call (2000) + 1 event (535, which we can remove btw), so I would have expected more like 5k. The rest seems to be overhead due to the for loop.

contracts/modules/VersionManager.sol
```diff
@@ -188,16 +202,17 @@ contract VersionManager is IVersionManager, IModule, BaseFeature, Owned {
...
-     * @param _wallet the wallet to upgrrade
+     * @param _wallet the wallet to upgrade
+     * @param _version the new version
      */
-    function upgradeWallet(address _wallet) external onlyWalletOwnerOrFeature(_wallet) onlyWhenUnlocked(_wallet) {
-        doUpgradeWallet(_wallet);
+    function upgradeWallet(address _wallet, uint256 _version) external onlyWalletOwnerOrFeature(_wallet) onlyWhenUnlocked(_wallet) {
```
juniset, line 214: What is the reason to let the user choose the target version? Isn't there now the risk that a client upgrades to a version that contains a bug and we cannot prevent it?

contracts/modules/VersionManager.sol
th3m477: Each app will have a version number associated with it, and as we can get version number from wallet we will assume everything else matches (no need to check specific modules).

How do I get the version number, given a wallet address? I can look at the public getter on `walletVersions`; does that mean VersionManager will not change?

contracts/modules/VersionManager.sol
```diff
@@ -0,0 +1,313 @@
...
+
+    bytes32 constant NAME = "VersionManager";
+
+    bytes4 constant internal ADD_MODULE_PREFIX = bytes4(keccak256("addModule(address,address)"));
+    bytes4 constant internal UPGRADE_WALLET_PREFIX = bytes4(keccak256("upgradeWallet(address,uint256)"));
+
+    // Last bundle version
+    uint256 public lastVersion;
+    // Version to be used for new wallet creation
+    uint256 public newWalletVersion = 1;
```
th3m477, line 43: At wallet creation we will always use latest version (feature set)? So with create2 we will no longer include the list of modules (or even a version number) when generating the counterfactual addr?

Makes sense from cost reduction perspective.

contracts/modules/VersionManager.sol
olivdb: So we should be able to save at least 800+535=1335 gas by making the default init method totally empty.

th3m477 (approved):

contracts/modules/VersionManager.sol
olivdb: Yes, that's the downside and the reason why I had prevented it to begin with. But the clients requested this in case the client is out of sync with the latest version (e.g. user wallet is on V5, client only supports V7 and on chain the last version is V10 => need to update user to V7). I am hesitating to add a way for the multisig to blacklist a past version but not sure I want to overcomplicate the VersionManager.

contracts/modules/VersionManager.sol
olivdb: At wallet creation time, the version used to initialise the wallet is `newWalletVersion` which can be changed by the multisig. This is so that, in the scenario discussed in the common chat with Sergio (where the client does not yet support the latest on-chain version), the new wallet can be created with a version compatible with the client.

The downside (compared to 2.0) is that this new 2.1 architecture introduces some trust: the user has to trust that a new faulty bundle will not be released by the time the wallet is created. An alternative could be to create all wallets on V1 and require the user to do the upgrade from V1 to the version supported by the client, but it's going to be more costly.

contracts/modules/VersionManager.sol
olivdb: Gas savings provided by _featuresToInit, when the default `init` does nothing at all:

4635 gas per new feature during a wallet upgrade
5464 gas per feature during a wallet creation

contracts/modules/VersionManager.sol
juniset: Agree on not complicating the VersionManager.

contracts/modules/VersionManager.sol
juniset: Let's leave it as it is then. It seems that it is worth the gas saving.

contracts/modules/VersionManager.sol
th3m477: @olivdb you're changing this yes? I think the guys will have discussed it with you this morning so that wallets are created with a specific version, and not `newWalletVersion` ?

olivdb: Yes, it's changed as of 47b95a3.

juniset (approved):

PR #152, opened 2020-09-07
Fix/g0 audit

juniset:
Addressing the issues raised in G0 audit:
- updating `GuardianManager.isGuardian` to only check if the target address is a guardian (but not an EOA authorised to sign for a guardian)
- adding `GuardianManager.isGuardianOrGuardianSigner` to have a public method to check is a guardian or a signer for a guardian.
- renaming `TokenPriceStorage` into `TokenPriceRegistry` and removing the dependency to `Storage`.
- resetting daily spending counter after a call to `ApprovedTransfer.approveWethAndCallContract`.

### DIFF
--- a/contracts/modules/GuardianManager.sol
+++ b/contracts/modules/GuardianManager.sol
@@ -202,7 +202,7 @@ contract GuardianManager is BaseFeature {
     * @param _guardian the address to test
     * @return _isGuardian `true` if the address is a guardian for the wallet otherwise `false`.
     */
-    function isGuardianOrGuardianSigner(address _wallet, address _guardian) public view returns (bool _isGuardian) {
+    function isGuardianOrGuardianSigner(address _wallet, address _guardian) external view returns (bool _isGuardian) {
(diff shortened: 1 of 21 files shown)


contracts/modules/GuardianManager.sol
```diff
@@ -193,7 +193,17 @@ contract GuardianManager is BaseFeature {
...
+        _isGuardian = guardianStorage.isGuardian(_wallet, _guardian);
+    }
+
+    /**
+    * @notice Checks if an address is a guardian or an account authorised to sign on behalf of a smart-contract guardian.
+    * @param _wallet The target wallet.
+    * @param _guardian the address to test
+    * @return _isGuardian `true` if the address is a guardian for the wallet otherwise `false`.
+    */
+    function isGuardianOrGuardianSigner(address _wallet, address _guardian) public view returns (bool _isGuardian) {
```
olivdb, line 205: This could be `external` to emphasise that it's only used by clients

olivdb (approved):

PR #155, opened 2020-09-14
VersionManager: fixes and additional documentation

olivdb:
Addressing Trail of Bits comments and adding more documentation for the new VersionManager-based architecture.


juniset (approved):
Only missing the description of the DexRegistry in the spec.

contracts/modules/UpgraderToVersionManager.sol
```diff
@@ -0,0 +1,82 @@
...
+pragma solidity ^0.6.12;
+
+import "./common/IModule.sol";
+import "./common/IVersionManager.sol";
+import "../infrastructure/IModuleRegistry.sol";
+import "../infrastructure/storage/ILockStorage.sol";
+import "../wallet/IWallet.sol";
+
+/**
+ * @title SimpleUpgrader
```
juniset, line 26: Title should be `UpgraderToVersionManager`.

contracts/modules/VersionManager.sol
```diff
@@ -169,7 +169,12 @@ contract VersionManager is IVersionManager, IModule, BaseFeature, Owned {
      */
     function isFeatureAuthorised(address _wallet, address _feature) public view override returns (bool) {
         uint256 version = walletVersions[_wallet];
-        return isFeatureInVersion[_feature][version] || _feature == address(this) || (version == 0 && _feature == walletFactory);
+        return isFeatureInVersion[_feature][version] ||
+            // The VersionManager is the only feature that isn't stored in isFeatureInVersion
+            _feature == address(this) ||
+            // Temporary give the factory or an upgrader module the same permissions as features
+            // when a wallet is not yet initialized (version == 0)
+            (version == 0 && (_feature == walletFactory || IWallet(_wallet).authorised(_feature)));
```
juniset, line 177: I'm not sure I like this as we are adding exceptions to  `isFeatureAuthorised` which is used in many places while they are only needed when `upgradeWallet` is called. It makes reasoning about security more complicated.

I think it would be cleaner to directly add these conditions to `upgradeWallet`.

```
function isFeatureAuthorised(address _wallet, address _feature) public view override returns (bool) {
     return isFeatureInVersion[_feature][version] || _feature == address(this);
}

function upgradeWallet(address _wallet, uint256 _toVersion) external override onlyWhenUnlocked(_wallet) {
        uint256 fromVersion = walletVersions[_wallet];
        require(isFeatureAuthorised(_wallet, msg.sender) || fromVersion == 0 && ((_feature == walletFactory || IWallet(_wallet).authorised(msg.sender)), "VM: caller cannot call upgrade");
        require(_toVersion > 0 && _toVersion <= lastVersion, "VM: Invalid _toVersion");
        require(fromVersion < _toVersion, "VM: Already on new version");
        walletVersions[_wallet] = _toVersion;
```

contracts/modules/VersionManager.sol
```diff
@@ -270,10 +263,10 @@ contract VersionManager is IVersionManager, IModule, BaseFeature, Owned {
         bytes memory _data
     ) 
         external 
-        onlyFeature(_wallet)
         override
         returns (bytes memory _res) 
     {
+        require(isSenderFeatureOrWalletFactory(_wallet), "VM: sender may not invoke wallet");
```
juniset, line 269: Why do we give the right to the `WalletFactory` to call the wallet?

juniset (approved):

contracts/modules/VersionManager.sol
```diff
@@ -208,17 +219,25 @@ contract VersionManager is IVersionManager, IModule, BaseFeature, Owned {
...
             isOwner(_wallet, msg.sender), // Upgrade triggered directly by the owner (from version 1 to version 2,3,4,...)
             "VM: sender may not upgrade wallet"
         );
-        require(_toVersion > 0 && _toVersion <= lastVersion, "VM: Invalid _toVersion");
         uint256 fromVersion = walletVersions[_wallet];
-        require(fromVersion < _toVersion, "VM: Already on new version");
-        walletVersions[_wallet] = _toVersion;
+        uint256 minVersion_ = minVersion;
+        uint256 toVersion;
+        if(toVersion < minVersion_ && _isWalletInitializing(_wallet)) {
```
juniset, line 230: It should be `_toVersion > minVersion`.

contracts/modules/VersionManager.sol
```diff
@@ -113,6 +115,15 @@ contract VersionManager is IVersionManager, IModule, BaseFeature, Owned {
         walletFactory = _factory;
     }
 
+    /**
+     * @notice Lets the owner change the minimum allowed version
+     * @param _minVersion the minimum allowed version
+     */
+    function setMinVersion(uint256 _minVersion) external onlyOwner {
+        require(_minVersion > 0 && _minVersion <= lastVersion, "VM: invalid _minVersion");
```
juniset, line 123: Shouldn't it be `_minVersion > minVersion` to prevent us from potentially reverting to an incorrect state?

contracts/modules/VersionManager.sol
```diff
@@ -254,10 +279,14 @@ contract VersionManager is IVersionManager, IModule, BaseFeature, Owned {
         bytes memory _data
     ) 
         external 
-        onlyFeature(_wallet)
         override
         returns (bytes memory _res) 
     {
+        require(
+            _isFeatureAuthorisedForWallet(_wallet, msg.sender) || // Wallet invoked by a feature
+            _isWalletInitializing(_wallet), // Wallet invoked by the WalletFactory
```
juniset, line 287: This is no longer needed since we have removed the ENS logic from the `WalletFactory`.

juniset (changes requested):

contracts/modules/VersionManager.sol
olivdb: I believe it's correct: if the version that the WalletFactory is trying to use is lower than the minimal required version, the version is automatically bumped to that minimal required version.

juniset: Sorry, I meant `_toVersion < minVersion` :-) You are using `toVersion` which hasn't been initialised at that point.

contracts/modules/VersionManager.sol
olivdb: Potentially but I thought of leaving the owner the flexibility of "re-whitelisting" a previously blacklisted version in case this was ever needed.

juniset (approved):

PR #156, opened 2020-09-23
contract and script to deploy the ArgentWalletDetector

juniset:
Simple contract to detect if a given address corresponds to an Argent wallet. 
The contract exposes a single `isArgentWallet(address)` method that returns true if the codehash matches one of the deployed Proxy and if the target implementation matches one of the deployed BaseWallet. Only the owner of the contract can add code hash and implementations.

### DIFF
--- a/contracts/infrastructure/ArgentWalletDetector.sol
+++ b/contracts/infrastructure/ArgentWalletDetector.sol
@@ -92,6 +92,7 @@ contract ArgentWalletDetector is Owned {
     */
     function addCodeAndImplementationFromWallet(address _argentWallet) external onlyOwner {
         bytes32 codeHash;   
+		// solhint-disable-next-line no-inline-assembly
     	assembly { codeHash := extcodehash(_argentWallet) }
         addCode(codeHash);
         address implementation = IWallet(_argentWallet).implementation(); 
@@ -118,6 +119,7 @@ contract ArgentWalletDetector is Owned {
     */
 	function isArgentWallet(address _wallet) external view returns (bool) {
 		bytes32 codeHash;    
+		// solhint-disable-next-line no-inline-assembly
     	assembly { codeHash := extcodehash(_wallet) }
 		return acceptedCodes[codeHash].exists && acceptedImplementations[IWallet(_wallet).implementation()].exists;
 	}
--- a/contracts/infrastructure/ArgentWalletDetector.sol
+++ b/contracts/infrastructure/ArgentWalletDetector.sol
@@ -28,99 +28,99 @@ import "../wallet/IWallet.sol";
  */
 contract ArgentWalletDetector is Owned {
 	
-	// The accepted code hashes
-	bytes32[] private codes;
-	// The accepted implementations
-	address[] private implementations;
-	// mapping to efficiently check if a code is accepted
+    // The accepted code hashes
+    bytes32[] private codes;
+    // The accepted implementations
+    address[] private implementations;
+    // mapping to efficiently check if a code is accepted
     mapping (bytes32 => Info) public acceptedCodes;
-	// mapping to efficiently check is an implementation is accepted
-	mapping (address => Info) public acceptedImplementations;
+    // mapping to efficiently check is an implementation is accepted
+    mapping (address => Info) public acceptedImplementations;
 
-	struct Info {
+    struct Info {
         bool exists;
         uint128 index;
     }
 
-	// emits when a new accepted code is added
-	event CodeAdded(bytes32 indexed code);
-	// emits when a new accepted implementation is added 
-	event ImplementationAdded(address indexed implementation);
+    // emits when a new accepted code is added
+    event CodeAdded(bytes32 indexed code);
+    // emits when a new accepted implementation is added 
+    event ImplementationAdded(address indexed implementation);
 
-	constructor(bytes32[] memory _codes, address[] memory _implementations) public {
-		for(uint i = 0; i < _codes.length; i++) {
-			addCode(_codes[i]);
-		}
-		for(uint j = 0; j < _implementations.length; j++) {
-			addImplementation(_implementations[j]);
-		}
-	}
+    constructor(bytes32[] memory _codes, address[] memory _implementations) public {
+        for(uint i = 0; i < _codes.length; i++) {
+            addCode(_codes[i]);
+        }
+        for(uint j = 0; j < _implementations.length; j++) {
+            addImplementation(_implementations[j]);
+        }
+    }
 
-	/**
+    /**
     * @notice Adds a new acceted code hash.
     * @param _code The new code hash.
     */
-	function addCode(bytes32 _code) public onlyOwner {
+    function addCode(bytes32 _code) public onlyOwner {
         require(_code != bytes32(0), "AWR: empty _code");
         Info storage code = acceptedCodes[_code];
-		if(!code.exists) {
-			codes.push(_code);
-			code.exists = true;
-        	code.index = uint128(codes.length - 1);
-			emit CodeAdded(_code);
-		}
+        if(!code.exists) {
+            codes.push(_code);
+            code.exists = true;
+            code.index = uint128(codes.length - 1);
+            emit CodeAdded(_code);
+        }
     }
 	
-	/**
+    /**
     * @notice Adds a new acceted implementation.
     * @param _impl The new implementation.
     */
-	function addImplementation(address _impl) public onlyOwner {
+    function addImplementation(address _impl) public onlyOwner {
         require(_impl != address(0), "AWR: empty _impl");
         Info storage impl = acceptedImplementations[_impl];
-		if(!impl.exists) {
-			implementations.push(_impl);
-			impl.exists = true;
-        	impl.index = uint128(implementations.length - 1);
-			emit ImplementationAdded(_impl);
-		}
+        if(!impl.exists) {
+            implementations.push(_impl);
+            impl.exists = true;
+            impl.index = uint128(implementations.length - 1);
+            emit ImplementationAdded(_impl);
+        }
     }
 
-	/**
+    /**
     * @notice Adds a new acceted code hash and implementation from a deployed Argent wallet.
     * @param _argentWallet The deployed Argent wallet.
     */
     function addCodeAndImplementationFromWallet(address _argentWallet) external onlyOwner {
         bytes32 codeHash;   
-		// solhint-disable-next-line no-inline-assembly
-    	assembly { codeHash := extcodehash(_argentWallet) }
+        // solhint-disable-next-line no-inline-assembly
+        assembly { codeHash := extcodehash(_argentWallet) }
         addCode(codeHash);
         address implementation = IWallet(_argentWallet).implementation(); 
         addImplementation(implementation);
     }
 
-	/**
+    /**
     * @notice Gets the list of accepted implementations.
     */
-	function getImplementations() public view returns (address[] memory) {
-		return implementations;
-	}
+    function getImplementations() public view returns (address[] memory) {
+        return implementations;
+    }
 
-	/**
+    /**
     * @notice Gets the list of accepted code hash.
     */
-	function getCodes() public view returns (bytes32[] memory) {
-		return codes;
-	}
+    function getCodes() public view returns (bytes32[] memory) {
+        return codes;
+    }
 
-	/**
+    /**
     * @notice Checks if an address is an Argent wallet
-	* @param _wallet The target wallet
+    * @param _wallet The target wallet
     */
-	function isArgentWallet(address _wallet) external view returns (bool) {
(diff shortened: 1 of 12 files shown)


contracts/infrastructure/ArgentWalletDetector.sol
```diff
@@ -0,0 +1,126 @@
...
+ * @title ArgentWalletDetector
+ * @notice Simple contract to detect if a given address represents an Argent wallet.
+ * The `isArgentWallet` method returns true if the codehash matches one of the deployed Proxy
+ * and if the target implementation matches one of the deployed BaseWallet.
+ * Only the owner of the contract can add code hash and implementations.
+ * @author Julien Niset - <julien@argent.xyz>
+ */
+contract ArgentWalletDetector is Owned {
+	
+	// The accepeted code hash
```
olivdb, line 31: typo accepeted

contracts/infrastructure/ArgentWalletDetector.sol
```diff
@@ -0,0 +1,126 @@
...
+	bytes32[] private codes;
+	// The accepted implementations
+	address[] private implementations;
+	// mapping to efficiently check if a code is accepeted
+    mapping (bytes32 => Info) public acceptedCodes;
```
olivdb, line 36: incorrect indentation?

contracts/infrastructure/ArgentWalletDetector.sol
```diff
@@ -0,0 +1,126 @@
...
+ * Only the owner of the contract can add code hash and implementations.
+ * @author Julien Niset - <julien@argent.xyz>
+ */
+contract ArgentWalletDetector is Owned {
+	
+	// The accepeted code hash
+	bytes32[] private codes;
+	// The accepted implementations
+	address[] private implementations;
+	// mapping to efficiently check if a code is accepeted
```
olivdb, line 35: typo -> accepted

contracts/infrastructure/ArgentWalletDetector.sol
```diff
@@ -0,0 +1,126 @@
...
+    mapping (bytes32 => Info) public acceptedCodes;
+	// mapping to efficiently check is an implementation is accepeted
```
olivdb, line 37: typos: is -> if; accepeted -> accepted

contracts/infrastructure/ArgentWalletDetector.sol
```diff
@@ -0,0 +1,126 @@
...
+	mapping (address => Info) public acceptedImplementations;
+
+	struct Info {
+        bool exists;
+        uint128 index;
+    }
```
olivdb, line 43: incorrect indentation?

contracts/infrastructure/ArgentWalletDetector.sol
```diff
@@ -0,0 +1,126 @@
...
+
+	// emits when a new accepeted code is added
+	event CodeAdded(bytes32 indexed code);
+	// emits when a new accepeted implementation is added 
+	event ImplementationAdded(address indexed implementation);
+
+	constructor(bytes32[] memory _codes, address[] memory _implementations) public {
+		for(uint i = 0; i < _codes.length; i++) {
```
olivdb, line 51: Inconsistent indentation (1/2): using 1 tab here

contracts/infrastructure/ArgentWalletDetector.sol
```diff
@@ -0,0 +1,126 @@
...
+			addImplementation(_implementations[j]);
+		}
+	}
+
+	/**
+    * @notice Adds a new acceted code hash.
+    * @param _code The new code hash.
+    */
+	function addCode(bytes32 _code) public onlyOwner {
+        require(_code != bytes32(0), "AWR: empty _code");
```
olivdb, line 64: Inconsistent indentation (2/2): using 2 spaces here

contracts/infrastructure/ArgentWalletDetector.sol
```diff
@@ -0,0 +1,126 @@
...
+    mapping (bytes32 => Info) public acceptedCodes;
+	// mapping to efficiently check is an implementation is accepeted
+	mapping (address => Info) public acceptedImplementations;
+
+	struct Info {
+        bool exists;
+        uint128 index;
+    }
+
+	// emits when a new accepeted code is added
```
olivdb, line 45: typo: ->accepted

contracts/infrastructure/ArgentWalletDetector.sol
```diff
@@ -0,0 +1,126 @@
...
+	event CodeAdded(bytes32 indexed code);
+	// emits when a new accepeted implementation is added
```
olivdb, line 47: typo: ->accepted

contracts/infrastructure/ArgentWalletDetector.sol
```diff
@@ -0,0 +1,126 @@
...
+		for(uint i = 0; i < _codes.length; i++) {
+			addCode(_codes[i]);
+		}
+		for(uint j = 0; j < _implementations.length; j++) {
+			addImplementation(_implementations[j]);
+		}
+	}
+
+	/**
+    * @notice Adds a new acceted code hash.
```
olivdb, line 60: typo: ->accepted

contracts/infrastructure/ArgentWalletDetector.sol
```diff
@@ -0,0 +1,126 @@
...
+		if(!code.exists) {
+			codes.push(_code);
+			code.exists = true;
+        	code.index = uint128(codes.length - 1);
+			emit CodeAdded(_code);
+		}
+    }
+	
+	/**
+    * @notice Adds a new acceted implementation.
```
olivdb, line 75: typo -> accepted

contracts/infrastructure/ArgentWalletDetector.sol
```diff
@@ -0,0 +1,126 @@
...
+    * @param _impl The new implementation.
+    */
+	function addImplementation(address _impl) public onlyOwner {
+        require(_impl != address(0), "AWR: empty _impl");
+        Info storage impl = acceptedImplementations[_impl];
+		if(!impl.exists) {
+			implementations.push(_impl);
+			impl.exists = true;
+        	impl.index = uint128(implementations.length - 1);
```
olivdb, line 84: Inconsistent indentation: 8 spaces + 1 tab vs 3 tabs on the previous line

contracts/infrastructure/ArgentWalletDetector.sol
```diff
@@ -0,0 +1,126 @@
...
+			emit ImplementationAdded(_impl);
+		}
+    }
+
+	/**
+    * @notice Adds a new acceted code hash and implementation from a deployed Argent wallet.
```
olivdb, line 90: typo -> accepted

contracts/infrastructure/ArgentWalletDetector.sol
juniset: No idea what happened with indentation, there is no problem locally.

juniset: Should be fixed now.

test/argentWalletDetector.js
```diff
@@ -0,0 +1,81 @@
...
+      await detector.addCode(argentCode);
+      const isArgent = await detector.isArgentWallet(detector.contractAddress);
+      assert.isFalse(isArgent, "should return false when the code is not correct");
+    });
+
+    it("should return false when the implementation is not correct", async () => {
+      await detector.addImplementation(implementation1.contractAddress);
+      await detector.addCode(argentCode);
+      const isArgent = await detector.isArgentWallet(proxy2.contractAddress);
+      assert.isFalse(isArgent, "should return false when the implementation is not correct");
```
olivdb, line 78: This test is not succeeding for the right reason. The code hash of proxy2 is actually `0xb150b4da11d3a51d25092296fb52138fc2d22faa4191d25d4d2e99043f9a28b1`, which is not what is stored in `argentCode`. As a result, the test succeeds because `acceptedCodes[codeHash].exists` is `false`, not because `acceptedImplementations[IWallet(_wallet).implementation()].exists` is `false`. I'd suggest replacing
```
await detector.addCode(argentCode);
```
with:
```
await detector.addCode(ethers.utils.keccak256(proxy2._contract.deployedBytecode));
```
which will make sure we are adding the correct code hash for `proxy2`.

contracts/wallet/BaseWallet.sol
```diff
@@ -27,7 +27,7 @@ import "./IWallet.sol";
 contract BaseWallet is IWallet {
 
     // The implementation of the proxy
-    address public implementation;
+    address public override implementation;
```
olivdb, line 30: Note that this slightly changes the bytecode of the BaseWallet, so maybe it'd be better to keep BaseWallet unchanged and use `BaseWallet(_wallet).implementation()` in `ArgentWalletDetector.sol` instead of `IWallet(_wallet).implementation()`.

juniset: Are you sure it changes the bytecode? My understanding is that `override` is just used by the compiler to enforce a specific behaviour.

I know I could use BaseWallet instead but would prefer to code against the interface.

olivdb: Yes, I checked it both in Remix and in etherlime. It changes 32 bytes close to the end of the bytecode.

juniset: I really think we should use an interface here and not a specific `BaseWallet` implementation. Should I define a new IWallet interface locally until the next time we update the BaseWallet?

@elenadimitrova any opinion on this?

olivdb: A local interface definition would work, though I think you'd need to call is something different than IWallet to avoid conflicts in the build directory

scripts/deploy_wallet_detector.js
```diff
@@ -0,0 +1,63 @@
+// ///////////////////////////////////////////////////////////////////
+// Script to deploy a new ArgentWalletDectector.
+//
+// To deploy a new ArgentWalletDectector:
+// ./execute_script.sh deploy_wallet_detector.js <network>
```
elenadimitrova, line 5: I get a `Error: Cannot find module '/Users/Elena/Source/argent-contracts/NO-COMPILE'` error when running this

test/argentWalletDetector.js
```diff
@@ -0,0 +1,81 @@
+const ArgentWalletDetector = require("../build/ArgentWalletDetector");
```
elenadimitrova, line 1: Branch coverage is only 50% for the `ArgentWalletDetector` contract https://1568-166830546-gh.circle-artifacts.com/0/coverage/infrastructure/ArgentWalletDetector.sol.html
Shouldn't be difficult to get it to 100% by looking at which logic branches aren't covered

scripts/update_wallet_detector.js
```diff
@@ -0,0 +1,82 @@
...
+    await multisigExecutor.executeCall(ArgentWalletDetectorWrapper, "addCodeAndImplementationFromWallet", [wallet]);
+  }
+  if (code) {
+    console.log(`Adding wallet code ${code}`);
+    await multisigExecutor.executeCall(ArgentWalletDetectorWrapper, "addCode", [code]);
+  }
+  if (implementation) {
+    console.log(`Adding wallet implementation ${implementation}`);
+    await multisigExecutor.executeCall(ArgentWalletDetectorWrapper, "addImplementation", [implementation]);
+  }
```
elenadimitrova, line 77: Can we switch this if logic to be 
```
if (wallet) { 
...
} else if (code) { 
... 
} else if (implementation) {
...
}
```
to indicate we can have 1 of 3 option only as opposed to any combination of them. Also do the same for the nested `if`s  [above](https://github.com/argentlabs/argent-contracts/pull/156/files?file-filters%5B%5D=.js&file-filters%5B%5D=.sol#diff-745dc33a43b9e53984f71fe20dc8aa40435bab5d8fa4d737a41253491234249eR36) in 

```
if (idx > 0) {
    wallet = process.argv[idx + 1];
  } else {
    idx = process.argv.indexOf("--implementation");
    if (idx > 0) {
...
```

juniset: Agreed and done for the first part of your comment, but the second part is already in the form of an `if else`.

scripts/deploy_wallet_detector.js
juniset: Good catch, there was a typo in the execute_script.sh script.

utils/config-schema.json
```diff
@@ -145,7 +148,7 @@
         "MakerRegistry",
         "DexRegistry"
       ],
-      "additionalProperties": false
+      "additionalProperties": true
```
elenadimitrova, line 151: This should no longer be necessary.

elenadimitrova (approved):

PR #158, opened 2020-09-30
Addition of script to make tokens tradable

juniset:
Simple JS script to call the `TokenPriceRegistry` and set a list of tokens as `tradable`.

### DIFF
--- a/scripts/set_tradable_tokens.js
+++ b/scripts/set_tradable_tokens.js
@@ -1,4 +1,17 @@
+// Usage: ./execute.sh set_tradable_tokens.js staging --input [filename] (--dry)
+//
+// --input [filename]: path to json file formatted like this:
+//      {
+//        "0xE41d2489571d322189246DaFA5ebDe1F4699F498": true,
+//        "0xc00e94Cb662C3520282E6f5717214004A7f26888": true,
+//        "0x0d438f3b5175bebc262bf23753c1e53d03432bde": false
+//      }
+//
+// --dry: if set, will not send the transaction, just output what will be updated
+//
+
 const fs = require("fs");
+const ethers = require("ethers");
 const MultiSig = require("../build/MultiSigWallet");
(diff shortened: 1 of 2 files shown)


scripts/set_tradable_tokens.js
```diff
@@ -0,0 +1,39 @@
...
+  const { deployer } = manager;
+  const deploymentWallet = deployer.signer;
+  const { config } = configurator;
+
+  const TokenPriceStorageWrapper = await deployer.wrapDeployedContract(TokenPriceRegistry, config.modules.TokenPriceRegistry);
+
+  const tokens = JSON.parse(fs.readFileSync(input, "utf8"));
+  const tradable = Array(tokens.length).fill(true);
+
+  // const res = await TokenPriceStorageWrapper.getTradableForTokenList(tokens);
```
elenadimitrova, line 29: We should check the `tradable` flag is not set on any of the submitted tokens as the contract reverts when a manager disables a tradable token.

scripts/set_tradable_tokens.js
```diff
@@ -0,0 +1,39 @@
...
+  // const res = await TokenPriceStorageWrapper.getPriceForTokenList(tokens);
```
elenadimitrova, line 30: We should also check there is a price set for the token we set the flag for as the contracts don't check existence of token in list before setting `tradable` flag on it meaning we can end up with tokens with 0 prices that are tradable.

elenadimitrova (changes requested):

gergold:
I've addressed all your comments except the last one (we can end up with tokens with 0 prices that are tradable), I'm checking with Tom and Adam what will be the process, set tradable or price first.

elenadimitrova:
Thanks @gergold can you also fix build pls `npm run lint:js` is failing

elenadimitrova (approved):

PR #161, opened 2020-10-08
Cleaning the scripts we use to manage the contracts owned by the multisig

juniset:


scripts/execute_script.sh
```diff
@@ -6,7 +6,10 @@
 
 set -e # stop the script if any subprocess fails
 
-etherlime compile --runs 999
+# rm -rf build
+# npm run compile:lib
+# npm run compile
+# npm run compile:legacy
```
elenadimitrova, line 12: I'm for this change considering how long compilation takes currently however the obvious risk is that we execute scripts without compiling to latest version of the contracts. As a prevention, can we add some warning text in the comments header of scripts using `execute_script` to say `COMPILE CONTRACTS BEFORE RUNNING` or something similar?

scripts/update_compound_registry.js
```diff
@@ -1,3 +1,16 @@
...
+// Script to add/remove CTokens on Compound
+//
+// To add a token:
+// ./execute_script.sh update_compound_registry.js <network> --add --token <token address> --ctoken <ctoken address>
+//
+// To remove a token:
+// ./execute_script.sh update_compound_registry.js <network> --remove --token <token address>
+//
+// where:
+//    - network = [test, staging, prod]
```
elenadimitrova, line 11: Can the network not be `ganache` as well for local testing?

scripts/update_wallet_in_factory.js
```diff
@@ -1,4 +1,12 @@
-// Usage: ./execute.sh updateBaseWallet.js staging
+// ///////////////////////////////////////////////////////////////////
+// Script to update the BaseWalet implementation in the WalletFactory contract
+//
+// Can be executed as:
+// ./execute_script.sh update_wallet_in_factory.js <network>
+//
+// where:
+//     - network = [test, staging, prod]
```
elenadimitrova, line 8: Can we remove `defaultNetwork = "test"`  in this script to mandate passing the network property

scripts/execute_script.sh
juniset: Agreed. Actually I've added a `--no-compile` flag in #156 (https://github.com/argentlabs/argent-contracts/pull/156/files#diff-0aa9ab67bc9291891ca0b035f24dce95R12) to address that. We can either wait for #156 to be merged or I can include that change in this PR.

elenadimitrova: I don't mind either way.

elenadimitrova (approved):

PR #162, opened 2020-10-08
Add configuration reader script

elenadimitrova:
Allows us to read configuration files easily. Note that config is read without verification to allow us to read potentially invalid config. The script does the validation at the end and logs any schema validation errors. I added this logic after noticing the current staging configuration is invalid due to the missing `ArgentWalletDetector` schema property.

### DIFF
--- a/scripts/configReader.js
+++ b/scripts/configReader.js
@@ -37,7 +37,7 @@ async function main() {
   // This will allow the config to be printed regardless of whether it's valid or not
   await configurator.load(false);
   const configuration = configurator.copyConfig();
-  console.log(configuration);
+  console.log(JSON.stringify(configuration, null, 4));
 
   // Validate the configuration. Prints any validation error.
   configurator._validate();
--- a/scripts/configReader.js
+++ b/scripts/configReader.js
@@ -11,15 +11,12 @@
 
 require("dotenv").config();
 
-const path = require("path");
-
 const ConfiguratorLoader = require("../utils/configurator-loader.js");
 const Configurator = require("../utils/configurator.js");
 
 async function main() {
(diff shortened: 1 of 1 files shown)


scripts/configReader.js
```diff
@@ -0,0 +1,36 @@
...
+
+  let configLoader;
+  if (remotelyManagedNetworks.includes(network)) {
+    const bucket = `${process.env.S3_BUCKET_PREFIX}-${network}`;
+    const key = process.env.S3_CONFIG_KEY;
+    configLoader = new ConfiguratorLoader.S3(bucket, key);
+  } else {
+    const fileName = env ? `${network}.${env}.json` : `${network}.json`;
+    const filePath = path.join(__dirname, "./config", fileName);
+    configLoader = new ConfiguratorLoader.Local(filePath);
```
olivdb, line 26: Is this really useful? If there is a local json config file, we could just open this file directly without requiring a script, no?

olivdb:
Instead of having to read it via a script, could we not simply have an unauthenticated GET request that returns the file from S3? @gergold

juniset (approved):

juniset:
@olivdb The config file contains some information about our AWS configuration which we might not want to expose publicly.

scripts/configReader.js
```diff
@@ -0,0 +1,36 @@
...
+  }
+  const configurator = new Configurator(configLoader);
+  await configurator.load();
+  const configuration = configurator.copyConfig();
+  console.log(configuration);
```
olivdb, line 31: Could we `JSON.stringify` the `configuration` so that all subfields display properly?

olivdb (approved):

PR #172, opened 2020-11-09
Preventing non-static calls in VersionManager's fallback

olivdb:
This fix prevents the owner of the VersionManager from exploiting the VersionManager's fallback method to upgrade a wallet without its owner's consent.


contracts/modules/common/StaticDelegateCaller.sol
```diff
@@ -0,0 +1,82 @@
+// Copyright (C) 2018  Argent Labs Ltd. <https://argent.xyz>
```
juniset, line 1: No longer needed?

olivdb: Oops it was not supposed to be pushed

test/nftTransfer.js
```diff
@@ -112,7 +112,7 @@ describe("Token Transfer", function () {
         }
       } else {
         const txPromise = nftFeature.from(owner1)
-          .transferNFT(wallet1.contractAddress, nftContract.contractAddress, recipientAddress, nftId, safe, ZERO_BYTES32);
+          .transferNFT(wallet1.contractAddress, nftContract.contractAddress, recipientAddress, nftId, safe, ZERO_BYTES32, { gasLimit: 300000 });
```
juniset, line 115: Have you benchmarked the extra gas cost for static calls that are initiated on-chain (e.g. safe NFT transfers)?

contracts/modules/VersionManager.sol
```diff
@@ -175,13 +175,32 @@ contract VersionManager is IVersionManager, IModule, BaseFeature, Owned {
...
 
+    /* ***************** Static Call Delegation ************************* */
+
+    /**
+     * @notice This method is used by the VersionManager's fallback (via an internal call) to determine whether
+     * the current transaction is a staticcall or not. The method succeeds if the current transaction is a static call, 
+     * and reverts otherwise. 
+     * @dev The use of an if/else allows to encapsulate the whole logic in a single function.
+     */
+    function verifyIsInStaticCall() public {
```
juniset, line 186: Maybe renaming that method to `onlyStaticCall` would make its intention even clearer?

test/nftTransfer.js
olivdb: It adds 110k gas to static calls

contracts/modules/VersionManager.sol
olivdb: We generally followed the naming convention of using:
- `isX()` for methods returning a boolean when X is true
- `verifyX()` for methods reverting when `isX()` returns false
- `onlyX()` for modifiers calling `verifyX()`

but I could simplify the name to `verifyStaticCall()` if clearer?

test/nftTransfer.js
juniset: 110k??? Is it for the entire static call, or on top of what we had in the earlier implementation (e.g. 1.6)?

olivdb: I managed to reduce it to 1-2k by limiting the gas sent to the inner call. It looks like a write operation in a static context  cause an `assert` instead of a `revert`.

juniset: Given that it only saves 1-2k of gas I think we shouldn't specify the gas passed to the call because it will start failing for non-static calls the day the cost of the CALL opcode is increased.

olivdb: Oh no it does save saves 99% of the gas. I meant that the *increase* (relative to 1.6) of the gas cost is now only 1-2k instead of 110k :)

juniset: Ok that's a lot then :-)

But shouldn't the minimum gas stipend for a call be 2300? I just want to make sure we provide enough gas for the call to fail because of the non-static context, and not for any other reason (specifically an out of gas).

olivdb: The logic of the subcall costs 425 gas to execute. In my tests I found that passing 625 gas to the subcall was enough to make it succeed. I'm not sure why 200 gas extra is required but it seems like the cost of the CALL itself does not need to be included in the stipend. A stipend of 1000 leaves a margin of 1000-625=375 gas to account for opcode repricing. I could increase it but keep in mind that all the gas passed to the subcall is fully consumed by the `assert`. 

Also note that if the stipend was too little for the subcall to succeed in a non-static context, you would notice the last test of versionManager.js fail

olivdb: Anyways, I've increased the stipend to 3000 to be more comfy.

juniset (approved):

PR #177, opened 2020-11-19
Add coverage script for truffle experiment

cgewecke:
[solidity-coverage 559][1]

Hi @elenadimitrova. 

PR adds a draft script to run coverage. 

**Good news**
+ It compiles additively, runs the tests and generates a report

**Bad news**
+ ~170 tests fail 
  + A few calls in the tests use hard coded gasLimits - these could be a problem
  + It's possible I misconfigured the ganache client (see `providerOptions` in `.solcover.js`)
+ Legacy files compilation seems to hang forever (skipped those)
+ Script file not linted (sorry)
+ Migrations are deleted for now...

**Other**
+ I disabled docker compilation in the coverage script because it seems super slow or broken - maybe [truffle 2005][2] was never really fixed

But...think this is part-way there and a baseline for figuring out what further problems there might be.

Just lmk whatever questions you have /  issues you discover.

[1]: https://github.com/sc-forks/solidity-coverage/issues/559
[2]: https://github.com/trufflesuite/truffle/issues/2005


elenadimitrova:
Thanks a lot for this work @cgewecke 🎉  ❤️  I'm really happy for the good news and not worried about the bad news because:
- The failing tests are due to `ganache-accounts.json` not being generated in the coverage context. I've updated the `providerOptions` to do that.
- The legacy files compilation I've managed to solve in a commit I've not yet pushed on the main branch so that's not a problem as well. 
- Not linting the script files is hardly bad news :) 
- We are not using `Migrations` so I won't miss it.

I can't push to this fork so merging as-is (only 2 failing tests remain to be sorted) and picking the fixes above in main branch.

cgewecke:
Ok awesome!

PR #180, opened 2020-12-02
Accurate gas estimation for relayed transactions

elenadimitrova:
Estimating relayed transactions with refunds logic has two core issues:
1) The `gasLimit` parameter used in the refunds logic depends on the estimate of the overall transaction, yet it is also included in that transaction. This means that essentially users need to sign two transactions, one for estimation and second one with the calculated `gasLimit`. 
2) Due to the refunds logic using a `gasleft()` opcode which depends on the overall gas sent for the transaction, [eth_estimateGas returns bad estimations](https://github.com/trufflesuite/ganache-core/issues/617).  

These have led us to break down the `RelayerManager.exececute()` logic into seperate parts, estimate those individually and use that logic for calculating the `gasLimit` for relayed transactions. This logic is encapsulated in the `relay-manager.js` component used in tests where we've enabled all relayed transactions to use gas estimation. See https://www.notion.so/argenthq/Gastimation-WC-Investments-893edacddfe24678ae6dae803e20bd46 for details on the maths behind this.

To enable further debugging of gas estimations, a `debugGasLimits` function is introduces which can be uncommented to start printing gas estimate and spending information in the `relay-manager.js`
```
    // await this.debugGasLimits(_module, _method, _params, _wallet, _signers);
```

The relayer estimation logic in `relay-manager.relay()` defaults to **_no refunds_** as these affect daily limits, account and contract balances thus skewing test results.  Refunds are tested in isolation in `relayer.js` tests. To enable refunds for a relayed transaction, simply pass in the following additional parameters  `_module, _method, _params, _wallet, _signers` to `relay-manager.relay()`.

### DIFF
--- a/test/approvedTransfer.js
+++ b/test/approvedTransfer.js
@@ -75,7 +75,7 @@ contract("ApprovedTransfer", (accounts) => {
       limitStorage.address,
       ethers.constants.AddressZero,
       versionManager.address);
-    manager.setRelayerManager(relayerManager);
+    await manager.setRelayerManager(relayerManager);
     walletImplementation = await BaseWallet.new();
 
     limitFeature = await TestLimitFeature.new(
--- a/test/approvedTransfer.js
+++ b/test/approvedTransfer.js
@@ -1,4 +1,5 @@
 /* global artifacts */
+const { assert } = require("chai");
 const ethers = require("ethers");
 const truffleAssert = require("truffle-assertions");
 
@@ -318,8 +319,9 @@ contract("ApprovedTransfer", (accounts) => {
           const txReceipt = await manager.relay(approvedTransfer, "approveTokenAndCallContract",
             [wallet.address, erc20.address, wallet.address, amountToApprove, target.address, invalidData],
             wallet, [owner, ...sortWalletByAddress([guardian1, guardian2])]);
-          const { success } = parseRelayReceipt(txReceipt);
+          const { success, error } = parseRelayReceipt(txReceipt);
           assert.isFalse(success);
+          assert.equal(error, "BT: Forbidden contract");
         }
 
         it("should revert when target contract is the wallet", async () => {
--- a/test/approvedTransfer.js
+++ b/test/approvedTransfer.js
@@ -1,5 +1,4 @@
 /* global artifacts */
-const { assert } = require("chai");
 const ethers = require("ethers");
 const truffleAssert = require("truffle-assertions");
 
--- a/utils/relay-manager.js
+++ b/utils/relay-manager.js
@@ -10,13 +10,11 @@ class RelayManager {
   async relay(_module, _method, _params, _wallet, _signers,
     _relayerAccount,
     _estimate = false,
-    _gasLimit = 2000000,
+    _gasLimit = 1,
     _nonce,
     _gasPrice = 0,
     _refundToken = ETH_TOKEN,
-    _refundAddress = ethers.constants.AddressZero,
-    _gasLimitRelay) {
-    if (!_gasLimitRelay) { _gasLimitRelay = (_gasLimit * 1.1); } // eslint-disable-line no-param-reassign
+    _refundAddress = ethers.constants.AddressZero) {
     const relayerAccount = _relayerAccount || await utils.getAccount(9);
     const nonce = _nonce || await utils.getNonceForRelay();
     const methodData = _module.contract.methods[_method](..._params).encodeABI();
@@ -35,21 +33,21 @@ class RelayManager {
       _refundAddress,
     );
 
-    if (_estimate === true) {
-      const gasUsed = await this.relayerManager.estimate.execute(
-        _wallet.address,
-        _module.address,
-        methodData,
-        nonce,
-        signatures,
-        _gasPrice,
-        _gasLimit,
-        _refundToken,
-        _refundAddress,
-        { gas: _gasLimitRelay, gasPrice: _gasPrice },
-      );
-      return gasUsed;
-    }
+    const gasEstimate = await this.relayerManager.estimate.execute(
+      _wallet.address,
+      _module.address,
+      methodData,
+      nonce,
+      signatures,
+      _gasPrice,
+      _gasLimit,
+      _refundToken,
+      _refundAddress,
+      { gasPrice: _gasPrice },
+    );
+
+    console.log("method", _method);
+    console.log("gasEstimate", gasEstimate.toString());
 
     const tx = await this.relayerManager.execute(
       _wallet.address,
@@ -61,8 +59,9 @@ class RelayManager {
       _gasLimit,
       _refundToken,
       _refundAddress,
-      { gas: _gasLimitRelay, gasPrice: _gasPrice, from: relayerAccount },
+      { gas: gasEstimate, gasPrice: _gasPrice, from: relayerAccount },
     );
+    console.log("gasUsed    ", tx.receipt.gasUsed.toString());
 
     return tx.receipt;
   }
--- a/utils/relay-manager.js
+++ b/utils/relay-manager.js
@@ -8,11 +8,12 @@ class RelayManager {
   }
 
   async relay(_module, _method, _params, _wallet, _signers,
-    _gasLimit = 1,
+    _gasLimit = 2000000,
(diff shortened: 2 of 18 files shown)


utils/relay-manager.js
```diff
@@ -61,11 +91,119 @@ class RelayManager {
...
+      if (_signers.length > 1) {
+        refundGas = 30000;
+      } else {
+        refundGas = 40000;
+      }
+
+      // If the refund is with a token, add transfer cost.
+      // We are using a simple ERC20 transfer cost, however this varies by supported token, e.g. ZRX, BAT, REP, DAI, USDC, or USDT
+      if (_refundToken !== ETH_TOKEN) {
+        refundGas += 30000;
```
juniset, line 163: I don't understand where this additional 30k comes from.

elenadimitrova: This was the option for using better relayer-level gasLimit estimate instead of adding the 50,000 buffer. I've commented it out with some description in case we decide to come back to it.

utils/relay-manager.js
```diff
@@ -1,26 +1,48 @@
...
+    let _gasLimit = gasLimit;
+    if (_refundToken !== ETH_TOKEN) {
+      const tokenPriceRegistryAddress = await this.relayerManager.tokenPriceRegistry();
+      const tokenPriceRegistry = await ITokenPriceRegistry.at(tokenPriceRegistryAddress);
+
+      const tokenPrice = await tokenPriceRegistry.getTokenPrice(_refundToken);
+      const refundCost = new BN(gasLimit).muln(_gasPrice);
+      // tokenAmount = refundCost * 10^18 / tokenPrice
+      const tokenAmount = new BN(10).pow(new BN(18)).mul(refundCost).div(tokenPrice);
+      _gasLimit = tokenAmount.toNumber();
```
juniset, line 43: I'm confused by this block. We should not change the `_gasLimit` when the refund is in token as the gas consumed doesn't change. But we may want to adapt the `_gasPrice` which represent the amount of token the client is paying per unit of gas consumed.

utils/relay-manager.js
```diff
@@ -35,21 +57,29 @@ class RelayManager {
...
+    // Get the number of zero and non-zero bytes in the relayer.execute calldata
+    const nonZerosString = executeData.toString().slice(2).replace(/0/g, "");
+    const nonZeros = nonZerosString.length;
+    const zeros = executeData.length - nonZeros;
+
+    /* Calculation used by the back end to relay transactions
+      gasLimit =
+      + 21k (base transaction)
+      + 16 * non-empty calldata bytes
+      + 4 * empty calldata bytes
```
juniset, line 80: Maybe mention that we add a buffer of 50k.

utils/relay-manager.js
```diff
@@ -61,11 +91,123 @@ class RelayManager {
...
+    console.log("ratio", _gasLimit / tx.receipt.gasUsed);
+
     return tx.receipt;
   }
+
+  /* Returns the gas limit to use as gasLimit parameter in execute function
+    + 5200  (isFeatureAuthorisedInVersionManager check)
+    + 0 / 1000 / 4800 based on which contract implements getRequiredSignatures()
+    + 2052 (getSignHash call)
+    + 45144 (checkAndUpdateUniqueness call)
```
juniset, line 109: Should say something like ` 6200 / 22200 based on which nonce strategy is used'

utils/relay-manager.js
```diff
@@ -61,11 +91,123 @@ class RelayManager {
...
+      requiredSigsGas = 1000;
+    }
+
+    // Estimate cost of checkAndUpdateUniqueness call
+    let nonceCheckGas = 0;
+    if (_signers.length === 1) {
+      nonceCheckGas = 6200;
+      // Most calls here are the first for wallet so default to 20K gas extra for a storage slot update
+      // We could check if the wallet nonce is empty before we add this
+      nonceCheckGas += 20000;
```
juniset, line 131: We should add 15k and not 20k because 5k is already accounted for in the `6200` .

utils/relay-manager.js
```diff
@@ -61,11 +91,123 @@ class RelayManager {
...
+      if (gasEstimateFeatureCall <= 0) {
+        gasEstimateFeatureCall = 140000;
+      }
+    }
+
+    let refundGas = 0;
+    const methodData = _module.contract.methods[_method](..._params).encodeABI();
+    const requiredSignatures = await _module.getRequiredSignatures(_wallet.address, methodData);
+
+    if (_gasPrice > 0 && requiredSignatures[1].toNumber() === 1) {
```
juniset, line 156: For clarity, maybe add a comment to say that we only refund when `gasPrice > 0 ` and the owner is signing. Took me some time to understand why you where adding that condition.

juniset (approved):

utils/relay-manager.js
```diff
@@ -29,43 +52,166 @@ class RelayManager {
...
+      methodData,
+      nonce,
+      signatures,
+      gasPrice,
+      gasLimit,
+      _refundToken,
+      _refundAddress).encodeABI();
+
+    // Get the number of zero and non-zero bytes in the relayer.execute calldata
+    const nonZerosString = executeData.toString().slice(2).replace(/0/g, "");
```
olivdb, line 73: This regex will remove all zero characters but we want to only remove zero bytes (i.e. `00` at even positions). We could use that regex instead:
```
const nonZerosString = executeData.toString().slice(2).replace(/00(?=(..)*$)/g,"")
```

utils/relay-manager.js
```diff
@@ -29,43 +52,166 @@ class RelayManager {
...
+    + 5200  (isFeatureAuthorisedInVersionManager check)
+    + 0 / 1000 / 4800 based on which contract implements getRequiredSignatures()
+    + 2052 (getSignHash call)
+    + 6200 / 22200 (checkAndUpdateUniqueness call based on which nonce strategy is used
+      Only in tests we add 15k provision for the first setting of nonce on wallet
+    + 10000 * number of signatures (validateSignatures call, should best be estimated but this is also close enough)
+    + Function call estimate
+    + 40000 / 30000 refund cost for 1 signatures and >1 signatures respectively
+
+    Ignoring multiplication and comparisson as that is <10 gas per operation
```
olivdb, line 117: typo: comparison

utils/relay-manager.js
```diff
@@ -29,43 +52,166 @@ class RelayManager {
...
+  */
+  async getGasLimitRefund(_module, _method, _params, _wallet, _signers, _gasPrice) {
```
olivdb, line 119: We should maybe rename `_module` to `_feature`

utils/relay-manager.js
```diff
@@ -1,26 +1,49 @@
...
+    let gasPrice = _gasPrice;
+    if (_refundToken !== ETH_TOKEN) {
+      const tokenPriceRegistryAddress = await this.relayerManager.tokenPriceRegistry();
+      const tokenPriceRegistry = await ITokenPriceRegistry.at(tokenPriceRegistryAddress);
+      const tokenPrice = await tokenPriceRegistry.getTokenPrice(_refundToken);
+
+      // How many tokens to pay per unit of gas spent
+      // refundCost = gasLimit * gasPrice
+      // tokenAmount = refundCost * 10^18 / tokenPrice
+      const tokenAmount = new BN(10).pow(new BN(18)).muln(_gasPrice).div(tokenPrice);
```
olivdb, line 43: This "tokenAmount" variable is in fact the gasPrice so maybe we could instead write 
```
gasPrice = new BN(10).pow(new BN(18)).muln(_gasPrice).div(tokenPrice).toNumber();
```

test/approvedTransfer.js
```diff
@@ -1,4 +1,5 @@
 /* global artifacts */
+const { assert } = require("chai");
```
olivdb, line 2: Is this needed?

utils/relay-manager.js
elenadimitrova: Good find!

test/approvedTransfer.js
elenadimitrova: Nope

olivdb (approved):

PR #181, opened 2020-12-07
Allow anyone to create wallets

olivdb:
This PR removes the restriction on who can create new wallets. Any account possessing a valid signature from a WalletFactory manager can now create a wallet.

This PR also constructs the CREATE2 salt in a way that is compatible with zkSync. It assumes that the order of the salt parameters used in zkSync follows the order in https://github.com/matter-labs/zksync/pull/259.


contracts/infrastructure/WalletFactory.sol
```diff
@@ -218,8 +219,33 @@ contract WalletFactory is Owned, Managed {
      */
     function validateInputs(address _owner, address _versionManager, address _guardian, uint256 _version) internal view {
         require(_owner != address(0), "WF: owner cannot be null");
-        require(IModuleRegistry(moduleRegistry).isRegisteredModule(_versionManager), "WF: invalid _versionManager");
```
elenadimitrova, line 221: We accept the fact that wallets could be created with a non-Argent implementation of `IVersionManager` do we?

elenadimitrova (commented):
Spec doc should be updated as well

contracts/infrastructure/WalletFactory.sol
juniset: I guess we need to remove this check otherwise we have a way to censure the creation of a wallet (by deregistering the target `VersionManager` from the `ModuleRegistry`). And it remains secure for the user since we have a signature from Argent on the wallet address (which includes the target versionManager address in the hash).

In that case the `ModuleRegistry` is no longer needed in the factory and can be removed entirely from this contract (constructor and `changeModuleRegistry` method).

contracts/infrastructure/WalletFactory.sol
```diff
@@ -29,19 +29,22 @@ import "../modules/common/IVersionManager.sol";
  * @notice The WalletFactory contract creates and assigns wallets to accounts.
  * @author Julien Niset - <julien@argent.xyz>
  */
-contract WalletFactory is Owned, Managed {
+contract WalletFactory is Owned {
```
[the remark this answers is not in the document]
juniset: I would prefer that as well, but it will require to add the `virtual` keyword to the methods of the `Managed` base contract. We would need to make sure it doesn't change the bytecode of the contracts that inherit from `Managed` .

contracts/infrastructure/WalletFactory.sol
juniset: Also, the method can now be marked as `pure`.

contracts/infrastructure/WalletFactory.sol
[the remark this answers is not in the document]
olivdb: Adding `virtual` to `revokeManager` does change the bytecode, which is why I had initially decided against changing `Managed`.  But upon closer inspection, only one 256 bit word in the bytecode is changed, close to the end. This might be some sort of hash of the source code which would be inconsequential to the execution logic. Indeed I was able to verify in Etherscan the bytecode produced by using `virtual` with a source file that didn't use `virtual` and vice versa. So if Etherscan doesn't care about this difference in bytecode, I think we could ignore it as well and make the change to `Managed` without having to maintain some legacy `Managed` for our existing contracts.

contracts/infrastructure/WalletFactory.sol
```diff
@@ -199,14 +197,17 @@ contract WalletFactory is Owned, Managed {
...
      * @param _salt The slat provided.
      * @param _owner The owner address.
      * @param _versionManager The version manager module
      * @param _guardian The guardian address.
      * @param _version The version of feature bundle
      */
-    function newSalt(bytes32 _salt, address _owner, address _versionManager, address _guardian, uint256 _version) internal pure returns (bytes32) {
-        return keccak256(abi.encodePacked(_salt, _owner, _versionManager, _guardian, _version));
+    function newSalt(bytes20 _salt, address _owner, address _versionManager, address _guardian, uint256 _version) internal pure returns (bytes32) {
+        return keccak256(abi.encodePacked(keccak256(abi.encodePacked(_owner, _versionManager, _guardian, _version)), _salt));
```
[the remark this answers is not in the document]
olivdb: I think it's correct, zksync's PR has `bytes32 salt = keccak256(abi.encodePacked(saltArg, _newPkHash));`. In our case `saltArg = keccak256(abi.encodePacked(_owner, _versionManager, _guardian, _version))` and `_newPkHash = _salt`.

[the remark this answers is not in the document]
juniset: Yes, to me this is correct as well.

contracts/infrastructure/WalletFactory.sol
```diff
@@ -21,65 +21,45 @@ import "../wallet/BaseWallet.sol";
...
     // The address of the base wallet implementation
     address public walletImplementation;
     // The address of the GuardianStorage
     address public guardianStorage;
 
     // *************** Events *************************** //
 
-    event ModuleRegistryChanged(address addr);
     event WalletCreated(address indexed wallet, address indexed owner, address indexed guardian);
+    event ManagerAdded(address indexed _manager);
```
juniset, line 41: No longer needed.

contracts/infrastructure/WalletFactory.sol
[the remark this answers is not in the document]
juniset: But shouldn't it be:
```
return keccak256(abi.encodePacked(abi.encodePacked(keccak256(abi.encodePacked(_owner, _versionManager, _guardian, _version)), _salt)));
```
with an additional `abi.encodePacked`?

deployment/7_deploy_2.4.js
```diff
@@ -0,0 +1,345 @@
...
+    config.modules.GuardianStorage,
+    config.modules.LimitStorage,
+    VersionManagerWrapper.address,
+    config.defi.weth,
+  );
+  // Deploy the TransferManager module
+  const TransferManagerWrapper = await TransferManager.new(
+    config.modules.LockStorage,
+    config.modules.TransferStorage,
+    config.modules.LimitStorage,
```
juniset, line 114: I'm a bit confused with what will happen to the daily limit of existing wallets with these settings. I think we should update the `init` method of the `TransferManager` to no longer migrate existing limits and save some gas.

Could be part of a separate PR if needed.

juniset (approved):

contracts/infrastructure_0.5/base/Managed.sol
```diff
@@ -0,0 +1,63 @@
...
+pragma solidity ^0.5.4;
+
+import "../../infrastructure/base/Owned.sol";
+
+/**
+ * @title Managed
+ * @notice Basic contract that defines a set of managers. Only the owner can add/remove managers.
+ * @author Julien Niset - <julien@argent.xyz>
+ */
+contract Managed is Owned {
```
elenadimitrova, line 26: Why is this added? I can't see a difference between this and the parallel instance in `infrastructure/base`?

olivdb: the 0.5 compiler doesn't understand `virtual` so this file is the same as the one in `infrastructure/base` but without the `virtual`.

deployment/7_deploy_2.4.js
olivdb: For future reference, we agreed to change `TransferManager.init()` to only setup the default limit for new wallets. Recently updated wallets will have their limits stored in `LimitStorage` and we will no longer migrate the limits of older wallets that had their limits directly stored in the old `TransferManager`.

contracts/infrastructure_0.5/base/Managed.sol
elenadimitrova: Okay then maybe update the pragma in `infrastructure` from `pragma solidity >=0.5.4 <0.7.0;` to `pragma solidity ^0.6.12;` might as well do that to `Owned` in there as well for consistency.

olivdb: Ok for `infrastructure/base/Managed.sol` but why change `Owned.sol` ? The same file can still be used by both `0.5` and `0.6`.

deployment/7_deploy_2.4.js
```diff
@@ -0,0 +1,344 @@
...
+const VersionManager = artifacts.require("VersionManager");
+const Upgrader = artifacts.require("UpgraderToVersionManager");
+const WalletFactory = artifacts.require("WalletFactory");
+
+const deployManager = require("../utils/deploy-manager.js");
+const MultisigExecutor = require("../utils/multisigexecutor.js");
+
+const utils = require("../utils/utilities.js");
+
+const TARGET_VERSION = "2.4.0";
```
elenadimitrova, line 28: This is now 2.5.0. Also rename file.

test/multisig.js
```diff
@@ -243,7 +243,7 @@ contract("MultiSigWallet", (accounts) => {
 
     it("should fail with the wrong nonce", async () => {
       const nonceOffset = 1;
-      await executeSendFailure([owner1, owner2], nonceOffset, true, false, "MSW: Badly ordered signatures");
+      await executeSendFailure([owner1, owner2], nonceOffset, true, false, "MSW");
```
elenadimitrova, line 246: Why has this changed?

test/factory.js
```diff
@@ -61,74 +60,46 @@ contract("WalletFactory", (accounts) => {
...
-    // Restore the good state of factory (we set these to bad addresses in some tests)
-    await factory.changeModuleRegistry(moduleRegistry.address);
+    // Restore the good state of factory (we use bad addresses in some tests)
     await factory.changeRefundAddress(refundAddress);
 
     versionManager = await deployVersionManager();
     await moduleRegistry.registerModule(versionManager.address, ethers.utils.formatBytes32String("versionManager"));
   });
 
   describe("Create and configure the factory", () => {
```
elenadimitrova, line 70: Just one test missing for the overwritten `revokeManager` that it reverts.

test/multisig.js
olivdb: Because a bad nonce can lead to either a "MSW: Badly ordered signatures" revert or a "MSW: Not enough valid signatures" revert.

elenadimitrova (approved):

PR #135, opened 2020-07-15
Make ENS label optional in the wallet factory

juniset:
We make the ENS label optional when creating a wallet. If the `_label` parameter of `createWallet` or `createCounterfactualWallet` is empty the wallet is created without an ENS. The ENS can be registered later.

PR #184, opened 2020-12-26
Register ENS subname for wallet

elenadimitrova:
Since we've removed the ability to auto register an ENS label on a wallet in #135 here we enable that functionality for the wallet owner.

This is implemented via a relayed `multiCall` signed by wallet owner to:
1) ens reverse registrar to `claimWithResolver` 
2) Argent ens manager to `register` - here we pass a signed hash of the wallet owner + ens label which is used for authorising the call in the `ArgentENSManager` instance. 

Gas cost for the above register ENS label multiCall:  368,458


contracts/modules/EnsFeature.sol
```diff
@@ -0,0 +1,68 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.6.12;
+
+import "./common/BaseFeature.sol";
+import "../infrastructure/ens/IENSManager.sol";
+
+/**
+ * @title EnsManagerFeature
```
olivdb, line 23: should be `EnsFeature` ?

contracts/modules/EnsFeature.sol
```diff
@@ -0,0 +1,68 @@
...
+ * @notice Feature to manage wallet's ENS record.
+ */
+contract EnsFeature is BaseFeature {
+  bytes32 constant NAME = "EnsFeature";
+
+  // The address of the ENS manager
+  address public ensManager;
+
+  constructor(ILockStorage _lockStorage,
```
olivdb, line 32: `ILockStorage _lockStorage,` should be indented on the next line

contracts/modules/EnsFeature.sol
```diff
@@ -0,0 +1,68 @@
...
+        ensManager = _ensManager;
+    }
+
+  modifier validateENSLabel(string memory _label) {
+      bytes memory labelBytes = bytes(_label);
+      require(labelBytes.length != 0, "EF: ENS label must be defined");
+      _;
+    }
+
+  function registerWalletENS(address payable _wallet, string memory _label)
```
olivdb, line 48: why `payable` ?

contracts/modules/EnsFeature.sol
```diff
@@ -0,0 +1,68 @@
...
+        IVersionManager _versionManager,
+        address _ensManager
+    )
+        BaseFeature(_lockStorage, _versionManager, NAME)
+        public
+    {
+        ensManager = _ensManager;
+    }
+
+  modifier validateENSLabel(string memory _label) {
```
olivdb, line 42: What's the reason of making this a modifier?

juniset:
This PR raises 2 potential problems or questions that we need to address:
- we (i.e. our backend) no longer controls the label registered for a wallet. Upon registration users select an ENS label that we store in the backend but with this PR users can register a different label which could create problems. We could decide that it is ok and let the backend update the stored label based on the `Registered` event, or add an additional parameter to the `registerWalletENS` containing e.g. a signature of `hash[owner,label]` by a manager account on the backend.
- it is possible for a user to call `registerWalletENS` multiple times and register multiple `argent.xyz` subdomains. We need to decide if that's desirable or update the PR accordingly.

contracts/infrastructure/ens/ArgentENSResolver.sol
```diff
@@ -24,7 +24,7 @@ import "./ENSResolver.sol";
  * to the list of resolved names.
  * @author Julien Niset - <julien@argent.im>
  */
-contract ArgentENSResolver is Owned, Managed, ENSResolver {
+contract ArgentENSResolver is Owned, Managed, IENSResolver {
```
juniset, line 27: Is there any value in updating the ENSResolver other than changing the solidity version and "cleaning" our folder structure? Not sure it is compelling enough to add a second instance of a perfectly working contract.

package.json
```diff
@@ -19,7 +19,7 @@
     "compile:test": "npx truffle compile --config truffle-config-contracts-test.js",
     "compile": "npm run compile:infrastructure && npm run compile:modules && npm run compile:wallet",
     "cc": "rm -rf build && rm -rf build-legacy && npm run compile:lib && npm run compile && npm run compile:legacy && npm run compile:test && npm run provision:lib:artefacts",
-    "ganache": "npx ganache-cli --chainId 1895 --networkId 1597649375983 --gasLimit=20700000 -e 10000 --acctKeys=\"./ganache-accounts.json\" --deterministic",
+    "ganache": "npx ganache-cli --chainId 1895 --networkId 1597649375983 --gasLimit=20700000 --allowUnlimitedContractSize -e 10000 --acctKeys=\"./ganache-accounts.json\" --deterministic",
```
juniset, line 22: `allowUnlimitedContractSize` should not be needed.

test/ens.js
```diff
@@ -155,11 +187,92 @@ contract("ENS contracts", (accounts) => {
...
+      transaction = await encodeTransaction(ensManager.address, 0, data, false);
+      transactions.push(transaction);
+
+      const txReceipt = await manager.relay(
+        module,
+        "multiCall",
+        [wallet.address, transactions],
+        wallet,
+        [owner],
+        0,
```
juniset, line 261: Since we are printing the gas cost of the registration we might as well include an ETH refund to have a more accurate value of what it will cost.

contracts/infrastructure/ens/ArgentENSManager.sol
```diff
@@ -87,34 +89,50 @@ contract ArgentENSManager is IENSManager, Owned, Managed {
...
         bytes32 node = keccak256(abi.encodePacked(rootNode, labelNode));
         address currentOwner = ensRegistry.owner(node);
-        require(currentOwner == address(0), "AEM: _label is alrealdy owned");
+        require(currentOwner == address(0), "AEM: label is already owned");
 
         // Forward ENS
         ensRegistry.setSubnodeRecord(rootNode, labelNode, _owner, address(ensResolver), 0);
         ensResolver.setAddr(node, _owner);
 
         // Reverse ENS
```
juniset, line 107: Just an idea but maybe here we could check if the current instance is the owner of the reverse node and only set the reverse ENS if it is. This will give the users the flexibility to only register the forward ENS if they want to.

elenadimitrova: That could be a problem for our own name resolution though as the clients are currently using the reverse resolver to check wallet address registrations. Or are you saying that functionality will just not work for them and that's okay?

contracts/infrastructure/ens/ArgentENSResolver.sol
elenadimitrova: We don't have to redeploy it. As you said, just cleaning up the contracts.

contracts/infrastructure/ens/ArgentENSManager.sol
juniset: Yes, the clients shouldn't assume the reverse ENS is set for an address. And we've had users not wanting their reverse ENS to be set for privacy reasons, so I believe it would be nice to have that flexibility (on top of being cheaper).

deployment/7_update_ens.js
```diff
@@ -0,0 +1,71 @@
+/* global artifacts */
```
juniset, line 1: Maybe rename this file `8_update_ens.js` so it is decoupled from the wallet upgrade?

contracts/infrastructure/ens/ArgentENSManager.sol
```diff
@@ -87,34 +89,53 @@ contract ArgentENSManager is IENSManager, Owned, Managed {
...
-        string memory name = ".".toSlice().join(parts);
-        ENSReverseRegistrar reverseRegistrar = ENSReverseRegistrar(_getENSReverseRegistrar());
-        bytes32 reverseNode = reverseRegistrar.node(_owner);
-        ensResolver.setName(reverseNode, name);
+        string memory name = string(abi.encodePacked(_label, ".", rootName));
+
+        // Optionally set the reverse ENS
+        bytes32 reverseNode = IReverseRegistrar(_getENSReverseRegistrar()).node(_owner);
+
+        if(ensRegistry.owner(reverseNode) == address(this)) {
```
juniset, line 112: I believe this condition should be `if(ensRegistry.resolver(reverseNode) == ensResolver)` because we should only update the resolver if it has been set for the reverse record with `claimWithResolver` .

juniset (changes requested):

juniset (approved):

PR #187, opened 2021-01-26
Minimal wallet

olivdb:
This PR aim to significantly reduce the gas cost of operating the wallet by simplifying the scope and the code of the smart-contracts.

- The security features designed to protect the wallet from a loss of the owner key remain unchanged (guardians, lock, recovery and transferOwnership)
- The security features designed to protect the wallet from a theft of the owner key are strengthen yet simplified by:
    - pushing the daily limit to 0 which enables us to remove all the logic associated to it
    - generalising whitelisting by adding, on top of trusted contacts, the ability to opt-in/opt-out from registries of authorised dapps, and introducing a first default registry managed by Argent (to include Compound, Maker, Paraswap, and all the dapps currently integrated in the app).
- Merging `transferToken`, `approve`, `callContracts`, and `approveAndCall` into a generic `multiCall` method that can take as input an array of transactions (to, value data).
- Removing all the specific modules (Compound, TokenExchanger, NFT, etc) as their logic can now be implemented through the new `multiCall` method.
- Extending the concept of guardian approved operations (e.g. one shot operations such as ApprovedTransfer.transferToken) by introducing sessions: additional signing keys that, once approved by a majority of guardians, can be used to bypass all the security checks during a pre-defined validity period.

Thanks to this reduced scope all the functionalities of the wallet are implemented in a single module which further reduces the cost of creation and operation.

To save some gas on calldata we also change the value of the `ETH-TOKEN` constant from `0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE` to `address(0)`.

**Todos (not exhaustive)**:
- add tests for the core functionalities of the wallet
- add `Filter` objects for all the dapps currently integrated in Argent
- shorten reason strings to reduce the size of the bytecode
- add timelock to the `DappRegistry`


elenadimitrova (commented):
This is only a preliminary review as the PR is not in a final state.

contracts/infrastructure/DappRegistry.sol
```diff
@@ -0,0 +1,78 @@
+pragma solidity ^0.6.12;
+
+import "./IAuthoriser.sol";
+import "./dapp/IFilter.sol";
+import "./base/Owned.sol";
+import "./storage/Storage.sol";
+
+contract DappRegistry is IAuthoriser, Storage, Owned {
+
+    bytes32 constant internal ARGENT_DEFAULT = keccak256("argent_default");
```
olivdb, line 10: No need for the keccak IMO. Using the raw string for ARGENT_DEFAULT (and other registry ids) will make debugging easier and saves about 1k of gas when calling `toggle()` thanks to the reduction of non-zero calldata bytes.

contracts/infrastructure/DappRegistry.sol
```diff
@@ -0,0 +1,78 @@
...
+        }
+        authorisations[_registry][_contract] = Authorisation(true, _filter);
+    }
+
+    function getFilter(address _wallet, address _contract) internal view returns (bool, address) {
+        // check argent default registry first (enbaled by default, i.e. false => true)
+        if (!isEnabled[_wallet][ARGENT_DEFAULT] && authorisations[ARGENT_DEFAULT][_contract].isActive) {
+            return (true, authorisations[ARGENT_DEFAULT][_contract].filter);
+        } else {
+            for (uint i = 0; i < registries.length; i++) {
```
olivdb, line 70: This is going to be costly for calls to contracts authorised only in recent registries (as well as many reverts) when the number of registries grows. I think it would be better to store a list of enabled registries for each wallet. It will make disabling a registry in `toggleDappRegistry()` a little bit more expensive but `toggleDappRegistry()` should be a rare operation anyways.

contracts/infrastructure/DappRegistry.sol
```diff
@@ -0,0 +1,78 @@
...
+            return _data.length == 0 || filter == address(0) || IFilter(filter).validate(_data);
+        }
+        return false;
+    }
+
+    function toggle(address _wallet, bytes32 _registry) external override onlyModule(_wallet) returns (bool) {
+        require(_registry == ARGENT_DEFAULT || registryManager[_registry] != address(0), "AR: unknow registry");
+        bool current = isEnabled[_wallet][_registry];
+        isEnabled[_wallet][_registry] = !current;
+        return !current;
```
olivdb, line 41: This should probably be `return (_registry == ARGENT_DEFAULT) ? current : !current` to make the `ToggledDappRegistry(_wallet, _registry, isEnabled)` event easier to interpret for clients.

contracts/infrastructure/DappRegistry.sol
```diff
@@ -0,0 +1,78 @@
...
+    mapping (address => mapping (bytes32 => bool)) public isEnabled;
+    // mapping [name][address][authorisation]
+    mapping (bytes32 => mapping (address => Authorisation)) public authorisations;
+
+    bytes32[] public registries;
+
+    mapping (bytes32 => address) public registryManager;
+
+    event RegistryCreated(bytes32 registry, address manager);
+    event RegistryRemoved(bytes32 registry);
```
olivdb, line 27: Currently unused

contracts/infrastructure/DappRegistry.sol
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    function toggle(address _wallet, bytes32 _registry) external override onlyModule(_wallet) returns (bool) {
+        require(_registry == ARGENT_DEFAULT || registryManager[_registry] != address(0), "AR: unknow registry");
+        bool current = isEnabled[_wallet][_registry];
+        isEnabled[_wallet][_registry] = !current;
+        return !current;
+    }
+
+    // Do we want to let the owner to delete a registry?
```
olivdb, line 44: I think it's fine to allow deletion of a registry (and it's probably necessary for scalability if the list of registries is stored globally as is the case now instead of _per wallet_ as I suggest here https://github.com/argentlabs/argent-contracts/pull/187/files#r573675996). Users can always use their own whitelist to circumvent registry restrictions. The ability to delete a registry would allow us (and eventually a community-controller governance contract) to remove unsafe registries and protect users.

contracts/modules/RelayerManager.sol
```diff
@@ -368,4 +386,31 @@ contract RelayerManager is BaseFeature {
...
+
+    function startSession(address _wallet, bytes calldata _data) internal returns (bool) {
+        require(_data.length >= 68, "RM: Invalid session parameters");
+        (address key, uint64 expires) = abi.decode(_data[36:], (address, uint64));
+        if (key != address(0)) {
+            if (expires > 0) {
+                sessions[_wallet] = Session(key, expires);
+                emit SessionCreated(_wallet, key, expires);
+            }
+            return true;
```
elenadimitrova, line 406: When doing an ordinary approved transfer without subsequently needing a session, we are forced to pass in any address to `startSession` for that to work. It does work but feels dirty.

contracts/modules/TransactionManager.sol
```diff
@@ -0,0 +1,226 @@
...
+    )
+        external
+        returns (bytes4)
+    {
+        return ERC721_RECEIVED;
+    }
+
+    // *************** Internal Functions ********************* //
+
+    function enableStaticCall(address _wallet) internal {
```
olivdb, line 199: Reminder to add a method that the owner can call to enable a static call implemented in an external contract and executed via this module's fallback.

olivdb: Maybe the fallback could be implemented by reading an entry in a `mapping(bytes4 => StaticCallExecution)` where `StaticCallExecution` is a 
```
struct {
    bytes4 result;
    address implementation;
}
``` 
where in most cases `implementation` is `address(0)` in which case `result` is simply returned by the fallback (e.g. for  `onERC721Received` and `onERC1155Received`) but in some special cases `result` is null and `implementation` contains an implementing contract (as would be the case for `isValidSignature`).

olivdb: This will also allow us to save a bunch of bytecode by not having to store any specific static call logic in the TransactionManager.

contracts/infrastructure/DappRegistry.sol
```diff
@@ -0,0 +1,78 @@
...
+
+    mapping (bytes32 => address) public registryManager;
+
+    event RegistryCreated(bytes32 registry, address manager);
+    event RegistryRemoved(bytes32 registry);
+
+    function authorise(address _wallet, address _contract, bytes calldata _data) external view override returns (bool) {
+        (bool isActive, address filter) = getFilter(_wallet, _contract);
+        if (isActive) {
+            return _data.length == 0 || filter == address(0) || IFilter(filter).validate(_data);
```
olivdb, line 32: I think that passing the `_wallet` to `validate()` would allow for more useful filters, e.g. a filter that checks that the recipient (`to`) of a UniswapV2Router trade (or the `beneficiary` of a Paraswap trade) is either the wallet itself or an address whitelisted by the wallet.

contracts/modules/TransactionManager.sol
```diff
@@ -0,0 +1,226 @@
...
+            (address first, address second) = abi.decode(_transaction.data[4:], (address, address));
+            return first == _wallet ? second : first;
+        }   
+        return _transaction.to;
+    }
+
+    function isAuthorised(address _wallet, address _spender, address _to, bytes calldata _data) internal view returns (bool) {
+        if (_to == _spender) {
+            return authoriser.authorise(_wallet, _to, _data); // do we need to block calls to the wallet or modules?
+        } else {
```
olivdb, line 218: Not sure this distinction is a good idea. If `spender != to`, then the only cases where a generic registry would be useful seems to be for ERC20 approvals. But in such cases, the filter would need to be able to distinguish an `approve()` call from another type of call. For example, a filter should deny a `transfer()` to the Paraswap Proxy but allow an `approve()` to it. Or a filter may allow any approval to the `UniswapV2Router` but only allow calls that do not use a beneficiary other than the wallet. So in general, the full `data` will be needed by the filter to make such distinctions. In conclusion I would only do `authoriser.authorise(_wallet, _to, _data);` here and let the filter figure out the details.

contracts/modules/RelayerManager.sol
```diff
@@ -106,17 +90,20 @@ contract RelayerManager is BaseFeature {
         external
         returns (bool)
     {
-        uint startGas = gasleft();
+        // initial gas = 21k + non_zero_bytes * 8 + zero_bytes * 4
```
olivdb, line 93: Typo: non_zero_bytes * 8 -> non_zero_bytes * 16

contracts/modules/RelayerManager.sol
```diff
@@ -106,17 +90,20 @@ contract RelayerManager is BaseFeature {
...
+        //            ~= 21k + calldata.length * [1/3 * 16 + 2/3 * 4]
+        uint256 startGas = gasleft() + 21000 + msg.data.length * 8;
         require(startGas >= _gasLimit, "RM: not enough gas provided");
         require(verifyData(_wallet, _data), "RM: Target of _data != _wallet");
-        require(isFeatureAuthorisedInVersionManager(_wallet, _feature), "RM: feature not authorised");
         StackExtension memory stack;
-        (stack.requiredSignatures, stack.ownerSignatureRequirement) = IFeature(_feature).getRequiredSignatures(_wallet, _data);
+        (stack.requiredSignatures, stack.ownerSignatureRequirement) = getRequiredSignatures(_wallet, _data);
```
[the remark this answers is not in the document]
olivdb: Upgrades will be calling `ArgentModule.addModule()` for which there is an entry on L65 of `ArgentModule.getRequiredSignatures()`.

contracts/modules/RelayerManager.sol
```diff
@@ -368,4 +386,31 @@ contract RelayerManager is BaseFeature {
...
+        uint numberOfSignatures = 1 + numberOfGuardians;
+        return (numberOfSignatures, OwnerSignature.Required);
+    }
+
+    function startSession(address _wallet, bytes calldata _data) internal returns (bool) {
+        require(_data.length >= 68, "RM: Invalid session parameters");
+        (address key, uint64 expires) = abi.decode(_data[36:], (address, uint64));
+        if (key != address(0)) {
+            if (expires > 0) {
+                sessions[_wallet] = Session(key, expires);
```
elenadimitrova, line 403: We are overwriting any existing sessions here. Might be better to throw instead and require owner to explicitly clear existing valid session before replacing?

contracts/modules/TransactionManager.sol
juniset: Agreed.

contracts/modules/TransactionManager.sol
```diff
@@ -0,0 +1,273 @@
...
+    )
+        external 
+        onlySelf()
+        onlyWhenUnlocked(_wallet)
+        returns (bytes[] memory)
+    {
+        multiCallWithApproval(_wallet, _transactions);
+    }
+
+    function multiCallWithGuardiansAndStartSession(
```
th3m477, line 101: StartingSession instead of AndStartSession?

"And" feels like session happens after (and order not relevant)

contracts/modules/TransactionManager.sol
```diff
@@ -0,0 +1,273 @@
...
+    struct Call {
+        address to;
+        uint256 value;
+        bytes data;
+        bool isSpenderInData;
+    }
+
+    // *************** Events *************************** //
+
+    event Transfer(address indexed wallet, address indexed token, uint256 indexed amount, address to, bytes data);
```
th3m477, line 46: are we still using this event anywhere?

contracts/modules/SimpleUpgrader.sol
```diff
@@ -57,7 +54,6 @@ contract SimpleUpgrader is IModule {
      */
     function init(address _wallet) external override {
         require(msg.sender == _wallet, "SU: only wallet can call init");
-        require(!lockStorage.isLocked(_wallet), "SU: wallet locked");
```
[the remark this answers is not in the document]
juniset: We don't, the check is on the `addModule` method.

contracts/modules/RelayerManager.sol
```diff
@@ -106,17 +90,20 @@ contract RelayerManager is BaseFeature {
         external
         returns (bool)
     {
-        uint startGas = gasleft();
+        // initial gas = 21k + non_zero_bytes * 8 + zero_bytes * 4
+        //            ~= 21k + calldata.length * [1/3 * 16 + 2/3 * 4]
```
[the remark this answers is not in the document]
juniset: Yes, the assumption is 1/3 non-empty bytes and 2/3 empty bytes.

package.json
```diff
@@ -13,7 +13,7 @@
     "compile:infrastructure_0.5": "npx truffle compile --config truffle-config-infrastructure-0.5.js",
     "compile:infrastructure_0.6": "npx truffle compile --config truffle-config-infrastructure.js",
     "compile:infrastructure": "npm run compile:infrastructure_0.5 && npm run compile:infrastructure_0.6",
-    "compile:modules": "npx truffle compile --config truffle-config-modules.js && rm build/contracts/ApprovedTransfer.json && npx truffle compile --contracts_directory=./contracts/modules/ApprovedTransfer*.sol && rm build/contracts/RelayerManager.json && npx truffle compile --contracts_directory=./contracts/modules/RelayerManager*.sol && rm build/contracts/TransferManager.json && npx truffle compile --contracts_directory=./contracts/modules/TransferManager*.sol && rm build/contracts/TokenExchanger.json && npx truffle compile --contracts_directory=./contracts/modules/TokenExchanger*.sol",
+    "compile:modules": "npx truffle compile --config truffle-config-modules.js",
```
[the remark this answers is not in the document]
elenadimitrova: Confirmed on Ropsten contracts are verify-able

test/argentWalletDetector.js
```diff
@@ -1,110 +0,0 @@
...
-const utils = require("../utils/utilities.js");
-
-const RANDOM_CODE = "0x880ac7547a884027b93f5eaba5ff545919fdeb3c23ed0d2094db66303b3a80ac";
-const ZERO_BYTES32 = ethers.constants.HashZero;
-const ZERO_ADDRESS = ethers.constants.AddressZero;
-
-const EMPTY_CODE_MSG = "AWR: empty _code";
-const EMPTY_IMPL_MSG = "AWR: empty _impl";
-
-describe("ArgentWalletDetector", () => {
```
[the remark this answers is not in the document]
elenadimitrova: Core tests were reinstated

utils/relay-manager.js
```diff
@@ -180,40 +175,40 @@ class RelayManager {
     return gasLimit;
   }
 
-  async debugGasLimits(_feature, _method, _params, _wallet, _signers) {
```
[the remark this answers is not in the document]
elenadimitrova: done

contracts/modules/RelayerManager.sol
elenadimitrova: Logic refactored in #196

contracts/modules/TransactionManager.sol
elenadimitrova: Implemented in #195

elenadimitrova (approved):

PR #189, opened 2021-02-03
Propose change to reduce some gas on registration

graemecode:
Curious what you think about this change. Allows removing the strings library altogether.


elenadimitrova (approved):
Thanks for the contribution! Indeed this is an outdated implementation using a lib that is obsoleted. It comes at a good time too as we are redeploying the ENSManager with the next release. Please just rebase.

elenadimitrova (approved):

PR #191, opened 2021-02-10
Enable refund for wallet creation

juniset:
This PR updates the `WalletFactory` to make the creation of wallet payable to the user. It is achieved by using the fact that during creation the `WalletFactory` is briefly an authorised module of the wallet and can call its `invoke` method to trigger a refund.

The amount and the token to use for the refund must be approved by the owner of the wallet through a signature that must be provided to the `createCounterfactualWallet` method.


contracts/infrastructure/WalletFactory.sol
```diff
@@ -31,56 +32,41 @@ import "../modules/common/IVersionManager.sol";
...
     address public guardianStorage;
+    // The recipient of the refund
+    address public refundAddress; 
 
     // *************** Events *************************** //
 
     event ModuleRegistryChanged(address addr);
+    event RefundAddressChanged(address addr);
     event WalletCreated(address indexed wallet, address indexed owner, address indexed guardian);
+    event Ok(bytes res);
```
olivdb, line 51: Forgot to remove that line

contracts/infrastructure/WalletFactory.sol
```diff
@@ -222,4 +223,60 @@ contract WalletFactory is Owned, Managed {
         require(_guardian != (address(0)), "WF: guardian cannot be null");
         require(_version > 0, "WF: invalid _version");
     }
+
+    /**
+     * @notice Refunds the creation of the wallet when provided with a vild signature from the wallet owner.
```
olivdb, line 228: Typo: vild

contracts/infrastructure/WalletFactory.sol
```diff
@@ -222,4 +223,60 @@ contract WalletFactory is Owned, Managed {
...
+                "\x19Ethereum Signed Message:\n32",
+                keccak256(abi.encodePacked(_refundAmount, _refundToken))
+            ));
+        address signer = Utils.recoverSigner(signedHash, _ownerSignature, 0);
+        if (signer == _owner) {
+            if (_refundToken == ETH_TOKEN) {
+				invokeWallet(_wallet, refundAddress, _refundAmount, "");
+            } else {
+                bytes memory methodData = abi.encodeWithSignature("transfer(address,uint256)", refundAddress, _refundAmount);
+                invokeWallet(_wallet, _refundToken, 0, methodData);
```
elenadimitrova, line 254: We should check this is successful similarly to how it's done in the Relayer.

contracts/infrastructure/WalletFactory.sol
```diff
@@ -222,4 +223,60 @@ contract WalletFactory is Owned, Managed {
...
+        internal
+    {
+        bytes32 signedHash = keccak256(abi.encodePacked(
+                "\x19Ethereum Signed Message:\n32",
+                keccak256(abi.encodePacked(_refundAmount, _refundToken))
+            ));
+        address signer = Utils.recoverSigner(signedHash, _ownerSignature, 0);
+        if (signer == _owner) {
+            if (_refundToken == ETH_TOKEN) {
+				invokeWallet(_wallet, refundAddress, _refundAmount, "");
```
elenadimitrova, line 251: Something with spacing here

contracts/infrastructure/WalletFactory.sol
```diff
@@ -222,4 +223,60 @@ contract WalletFactory is Owned, Managed {
...
+        address _to,
+        uint256 _value,
+        bytes memory _data
+    )
+        internal
+        returns (bytes memory _res) 
+    {
+        bool success;
+        (success, _res) = _wallet.call(abi.encodeWithSignature("invoke(address,uint256,bytes)", _to, _value, _data));
+        if (success && _res.length > 0) { //_res is empty if _wallet is an "old" BaseWallet that can't return output values
```
olivdb, line 270: Given that (1) we know `_wallet` is not an old BaseWallet and (2) we don't care about the return value, this can be simplified to:
```
if(!success) {
            assembly {
                returndatacopy(0, 0, returndatasize())
                revert(0, returndatasize())
            }
}
```

contracts/infrastructure/WalletFactory.sol
```diff
@@ -222,4 +223,60 @@ contract WalletFactory is Owned, Managed {
...
+            if (_refundToken == ETH_TOKEN) {
+				invokeWallet(_wallet, refundAddress, _refundAmount, "");
+            } else {
+                bytes memory methodData = abi.encodeWithSignature("transfer(address,uint256)", refundAddress, _refundAmount);
+                invokeWallet(_wallet, _refundToken, 0, methodData);
+            }
+        }
+    }
+
+    function invokeWallet(
```
elenadimitrova, line 259: Can't we call `_versionManager.checkAuthorisedFeatureAndInvokeWallet` instead of duplicating code?

olivdb (approved):

contracts/infrastructure/WalletFactory.sol
olivdb: The `WalletFactory` is added as a _module_, not as a _feature_ so the `_isFeatureAuthorisedForWallet` check would fail.

contracts/infrastructure/WalletFactory.sol
juniset: Following @elenadimitrova comment we do care about the return value for ERC20 tokens who's `transfer` method returns a bool. But I've simplified the code to account for (1).

contracts/infrastructure/WalletFactory.sol
juniset: I don't understand, it's properly indented on my IDE... not the first time it happens.

contracts/infrastructure/WalletFactory.sol
olivdb: Fine but note that the vast majority of tokens don't actually use the bool to indicate a failure but revert instead. That includes all the tokens you want to support for the refund.

contracts/infrastructure/WalletFactory.sol
```diff
@@ -267,16 +277,12 @@ contract WalletFactory is Owned, Managed {
     {
         bool success;
         (success, _res) = _wallet.call(abi.encodeWithSignature("invoke(address,uint256,bytes)", _to, _value, _data));
-        if (success && _res.length > 0) { //_res is empty if _wallet is an "old" BaseWallet that can't return output values
-            (_res) = abi.decode(_res, (bytes));
-        } else if (_res.length > 0) {
+        if (!success) {
```
olivdb, line 280: You still need ` (_res) = abi.decode(_res, (bytes));` here.

contracts/infrastructure/WalletFactory.sol
juniset: You're right. And even if that was not true, the only risk is to create a wallet when the refund fails because of a lack of funds... which should never happens if the backend monitors deposits correctly.

contracts/infrastructure/WalletFactory.sol
juniset: Ignoring this comment as we are now ignoring the returned bytes.

contracts/infrastructure/WalletFactory.sol
elenadimitrova: Okay yes, we should be able to change that in the minimal wallet PR though.

contracts/infrastructure/WalletFactory.sol
elenadimitrova: I am checking some of the refund tokens we support and so far I see they all return `bool`, e.g. ZRX and BAT (which are defined as refund tokens in the gas estimation document). Can we be specific here as to what tokens we'll support? Despite that considering the minimal cost of this check and that it's part of ERC20 I would prefer to leave it.

olivdb: According to the [spec](https://www.notion.so/argenthq/Make-L1-wallets-paid-666c018f4413452d9503cf07c513022d) the supported refund tokens are ETH, USDC, DAI, USDT, which revert on failure. ZRX and BAT do indeed return `false` on failure.

juniset: We are planning on removing ZRX and BAT from the transaction refund tokens, but in any-case for creation the refund tokens will be limited to ETH, DAI, USDC and USDT. These tokens either return `true` or revert. The added logic will literally never been used so I would prefer to not include it.

contracts/infrastructure/WalletFactory.sol
juniset: It's fixed.

contracts/infrastructure/WalletFactory.sol
juniset: And we can always revisit this choice in a coupe on month when do the next wallet upgrade.

elenadimitrova: I don't understand why is it a problem to properly support ERC20 instead of relying on configuration of what refund tokens we support. Alternatively we should remove it from the Relayer logic as well although I would prefer to keep it as a safeguard.

juniset: I don't feel strongly either way. I've re-added the returned bytes to be consistent with the Relayer logic.

elenadimitrova (approved):

PR #193, opened 2021-02-15
Updating refund mechanim

juniset:
This PR updates the refund mechanism in the RelayerManager module.

To determine the amount to refund, the contract now uses the minimum between the `_gasLimit` signed by the user and the amount of gas consumed by the transaction (as computed by the module) multiplied by the minimum between the `_gasPrice` signed by the user and the actual `tx.gasprice` as specified by the relayer.

If the refund is to be executed in an ERC20 token, the RelayerManager converts the `tx.gasprice` to token using the current price on the corresponding ETH/Token UniswapV2 pool. Refunding in ERC20 adds ~14500 gas compared to a refund in ETH (~10k to update the balances in the token contract and ~4.5k to get the token price from UniswapV2).


contracts/modules/RelayerManager.sol
```diff
@@ -352,12 +359,16 @@ abstract contract RelayerManager is BaseModule {
                         require(whitelistAfter > 0 && whitelistAfter < block.timestamp, "RM: refund not authorised");
                     }
             }
-            uint256 gasConsumed = _startGas.sub(gasleft()).add(14000);
-            uint256 refundAmount = Utils.min(gasConsumed, _gasLimit).mul(_gasPrice);
+            uint256 refundAmount;
             if (_refundToken == ETH_TOKEN) {
+                uint256 gasConsumed = _startGas.sub(gasleft()).add(14000);
```
elenadimitrova, line 364: Where has the 14K value come from? Maybe update the pr description with that. The logic we had before was dependent on the number of signatures, i.e. 30k/40k

utils/relay-manager.js
```diff
@@ -113,14 +112,8 @@ class RelayManager {
...
-  async getGasLimitRefund(_feature, _method, _params, _wallet, _signers, _gasPrice) {
-    let requiredSigsGas = 0;
-    const { contractName } = _feature.constructor;
-    if (contractName === "ApprovedTransfer" || contractName === "RecoveryManager") {
-      requiredSigsGas = 4800;
-    } else if (contractName === "GuardianManager") {
-      requiredSigsGas = 1000;
-    }
+  async getGasLimitRefund(_module, _method, _params, _wallet, _signers, _gasPrice) {
+    const requiredSigsGas = 4800;
```
elenadimitrova, line 116: The cost of `getRequiredSignatures` function hasn't changed. Why the change here?

elenadimitrova (changes requested):
This branch doesn't compile atm so it's not possible to test but still left a few preliminary comments.

contracts/modules/common/SimpleOracle.sol
```diff
@@ -0,0 +1,60 @@
...
+
+    function getReservesForTokenPool(address _token) internal view returns (uint256 wethReserve, uint256 tokenReserve) {
+        if (weth < _token) {
+            address pair = getPairForSorted(weth, _token);
+            (wethReserve, tokenReserve,) = IUniswapV2Pair(pair).getReserves();
+        } else {
+            address pair = getPairForSorted(_token, weth);
+            (tokenReserve, wethReserve,) = IUniswapV2Pair(pair).getReserves();
+        }
+        require(wethReserve != 0 && tokenReserve != 0, "SO: no liquidity");
```
olivdb, line 48: Unlikely scenario but if the liquidity is removed, perhaps we should skip the refund (i.e. return 0 here) instead of reverting the whole transaction.

contracts/modules/common/SimpleOracle.sol
```diff
@@ -0,0 +1,60 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.6.12;
+
+import "@uniswap/v2-periphery/contracts/interfaces/IUniswapV2Router01.sol";
+import "@uniswap/v2-core/contracts/interfaces/IUniswapV2Pair.sol";
+import "@openzeppelin/contracts/math/SafeMath.sol";
+
+contract SimpleOracle {
```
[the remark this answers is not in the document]
juniset: Why do you want to make it abstract and what would be the benefit? It has no unimplemented methods so I dont see the point.

contracts/modules/ArgentModule.sol
```diff
@@ -19,18 +19,19 @@ contract ArgentModule is BaseModule, RelayerManager, SecurityManager, Transactio
 
     constructor (
         IModuleRegistry _registry,
-        ILockStorage _lockStorage,
```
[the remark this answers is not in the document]
juniset: Since there is only one module it is more efficient to store the lock state locally. The needed code has been moved to the `BaseModule` .

contracts/modules/ArgentModule.sol
```diff
@@ -19,18 +19,19 @@ contract ArgentModule is BaseModule, RelayerManager, SecurityManager, Transactio
...
+        address _uniswapRouter,
         uint256 _securityPeriod,
         uint256 _securityWindow,
         uint256 _recoveryPeriod,
         uint256 _lockPeriod
     )
-        BaseModule(_registry, _lockStorage, _guardianStorage, _userWhitelist, _authoriser, _securityPeriod, NAME)
-        SecurityManager(_recoveryPeriod, _lockPeriod, _securityWindow)
         public
+        BaseModule(_registry, _guardianStorage, _userWhitelist, _authoriser, _securityPeriod, NAME)
```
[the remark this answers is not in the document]
juniset: I've re-added the `ModuleCreated` event that had disappeared.

contracts/modules/RelayerManager.sol
juniset: It was dependent on the use or not of the daily limit, which we determined using the  number of signatures. This is no longer relevant.

Now the cost of the remaining logic (i.e. `Utils.min()` twice, `invokeWallet` and emitting the `Refund` event) is constant and estimated at 14k gas when the refund is in ETH based on existing mainnet transactions. When the refund is in ERC20 we need to increase that number by 14500 gas (10k to update the 2 balances on the ERC20 contract and 4.5k to get the token price from the UniswapV2 pool).

I'll add a card to make sure we verify these numbers when testing on Ropsten and Mainnet.

utils/relay-manager.js
```diff
@@ -10,30 +10,28 @@ const IGuardianStorage = artifacts.require("IGuardianStorage");
 
 class RelayManager {
   constructor(guardianStorage, tokenPriceRegistry) {
-    this.guardianStorage = IGuardianStorage.at(guardianStorage);
-    if (tokenPriceRegistry !== ethers.constants.AddressZero) {
-      this.tokenPriceRegistry = ITokenPriceRegistry.at(tokenPriceRegistry);
-    }
+    this.guardianStorage = guardianStorage;
+    this.tokenPriceRegistry = tokenPriceRegistry;
```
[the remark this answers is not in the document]
juniset: What do you mean? JS is loosely typed.

contracts/modules/common/SimpleOracle.sol
juniset: I would prefer to be strict so it cannot be abused. If there is no liquidity then we should not let the user refund with that token (i.e. the transaction should not be relayed) but if it happens then the transaction should fail.

However, you raise a valid point: what happens when Uniswap releases its V3?

utils/relay-manager.js
juniset: Because we had logic to adapt the cost based on the module being called... and it is longer relevant with 1 module.

I don't want to add too much complexity here and start parsing data to determine the called being made so I think that choosing an upper bound to make sure the tests won't fail is ok for now.

I've added a card to revisit the `relay-manager.js` logic in a separate PR.

utils/configurator.js
```diff
@@ -44,6 +44,10 @@ class Configurator {
     this._config.defi.uniswap.factory = address;
   }
 
+  updateUniswapV2Router(address) {
+    this._config.defi.uniswap.v2Router = address;
```
elenadimitrova, line 48: We should add this to the schema

utils/relay-manager.js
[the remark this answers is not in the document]
elenadimitrova: I mean the type of these parameters being contract addresses here vs being of the appropriate contract types, e.g. https://github.com/argentlabs/argent-contracts/pull/193/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json#diff-a6485253a66ea540c4d8b0fb4c6f1328e5e17e3c9f2b476dddc4f4452c3d0388R34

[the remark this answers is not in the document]
juniset: I see. That's because you cannot use `async/await` in a `constructor` in JS.

test/relayer.js
```diff
@@ -0,0 +1,351 @@
...
+      const balanceEnd = await utils.getBalance(wallet.address);
+      expect(balanceEnd.sub(balanceStart)).to.be.lt.BN(0);
+    });
+
+    it("should refund in ERC20", async () => {
+      // token balance
+      const balanceStart = await token.balanceOf(relayer);
+      // send ETH
+      const transaction = await encodeTransaction(recipient, 10, ZERO_BYTES32, false);
+      const txReceipt = await manager.relay(
```
elenadimitrova, line 287: This is failing on `(wethReserve, tokenReserve,) = IUniswapV2Pair(pair).getReserves();` but I haven't investigated further

juniset: Thanks, that helps and I know what the problem is.

elenadimitrova (approved):

contracts-test/ArgentModuleTest.sol
```diff
@@ -0,0 +1,52 @@
+pragma solidity ^0.6.12;
+pragma experimental ABIEncoderV2;
+
+import "@uniswap/v2-periphery/contracts/interfaces/IUniswapV2Router01.sol";
+import "../contracts/modules/ArgentModule.sol";
+
+/**
+ * @notice Extends the ArgentModule to get the creation code of uniswap pairs locally
+ * and enbale ERC20 refunds in tests.
```
olivdb, line 9: Typo: enbale

olivdb (approved):

PR #194, opened 2021-02-16
Extended WalletCreated event with refund info

elenadimitrova:
Extend the `WalletCreated` event to include `refundToken` and `refundAmount` information to help clients with refund cost calculation.

Fixes to the `abi-uploader` to use for the truffle contract syntax.

Additionally there were some inconsistencies with wallet factory tests which were fixed.

### DIFF
--- a/contracts/infrastructure/WalletFactory.sol
+++ b/contracts/infrastructure/WalletFactory.sol
@@ -47,8 +47,7 @@ contract WalletFactory is Owned, Managed {
 
     event ModuleRegistryChanged(address addr);
     event RefundAddressChanged(address addr);
-    event WalletCreated(address indexed wallet, address indexed owner, address indexed guardian);
-    event WalletFeeRefund(address indexed wallet, address indexed refundAddress, address refundToken, uint256 refundAmount);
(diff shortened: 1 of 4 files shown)


contracts/infrastructure/WalletFactory.sol
```diff
@@ -48,6 +48,7 @@ contract WalletFactory is Owned, Managed {
     event ModuleRegistryChanged(address addr);
     event RefundAddressChanged(address addr);
     event WalletCreated(address indexed wallet, address indexed owner, address indexed guardian);
+    event WalletFeeRefund(address indexed wallet, address indexed refundAddress, address refundToken, uint256 refundAmount);
```
olivdb, line 51: Does refundAddress really need to be indexed? If not we could just add the refund info the the WalletCreated event instead of adding an extra event.

elenadimitrova: Matt agreed to drop the `refundAddress` and use `WalletCreated`so I'll do that.

olivdb (approved):

PR #195, opened 2021-02-17
Minimal wallet: static calls

olivdb:
- New BaseWallet storing a single module for the execution of all supported static calls
- ERC1155 support (included by default for new BaseWallet, and opt-in via the `enableERC1155TokenReceiver() ` method for old BaseWallet)


juniset (commented):
I really like this implementation and the update to the `BaseWallet` as it will reduce the cost of creating and upgrading wallets while keeping the `BaseWallet` ABI consistent.

My only comment is that I believe we also need to support ERC165 to support ERC1155, see https://eips.ethereum.org/EIPS/eip-1155.

contracts/modules/TransactionManager.sol
```diff
@@ -153,6 +157,51 @@ abstract contract TransactionManager is BaseModule {
...
+    function enableERC1155TokenReceiver(
+        address _wallet
+    )   
+        external
+        onlySelf()
+    {
+        IWallet(_wallet).enableStaticCall(address(this), ERC165_INTERFACE);
+        IWallet(_wallet).enableStaticCall(address(this), ERC1155_RECEIVED);
+        IWallet(_wallet).enableStaticCall(address(this), ERC1155_BATCH_RECEIVED);
+        emit ERC1155TokenReceiverEnabled(_wallet);
```
juniset, line 176: Do we need to emit an event given that it will not be emitted by new wallets that do support ERC1155 and it can be emitted multiple times if you call the method more than once?

contracts/modules/TransactionManager.sol
```diff
@@ -153,6 +157,51 @@ abstract contract TransactionManager is BaseModule {
...
+    }
+
+    /**
+     * @notice Returns true if this contract implements the interface defined by
+     * `interfaceId` (see https://eips.ethereum.org/EIPS/eip-165).
+     */
+    function supportsInterface(bytes4 interfaceID) external view returns (bool) {
```
juniset, line 183: Maybe move this method in the Callback section with `isValidSignature` and the `fallback` ?

juniset (approved):

contracts/modules/TransactionManager.sol
olivdb: Checking with clients

olivdb: Confirmed that it's not needed by clients

contracts/modules/TransactionManager.sol
```diff
@@ -177,33 +225,26 @@ abstract contract TransactionManager is BaseModule {
...
-    {
-        return ERC721_RECEIVED;
+
+    fallback() external {
+        bytes4 methodId = Utils.functionPrefix(msg.data);
+        if(methodId == ERC721_RECEIVED || methodId == ERC1155_RECEIVED || methodId == ERC1155_BATCH_RECEIVED) {
+            // solhint-disable-next-line no-inline-assembly
+            assembly {                
+                calldatacopy(0, 0, 0x04)
+                return (0, 0x20)
```
elenadimitrova, line 238: The return value here is expected to be `bytes4` why return `bytes32`?

olivdb: I tried but it looks like in assembly you can't return less than 32 bytes.

contracts/modules/TransactionManager.sol
```diff
@@ -163,9 +167,53 @@ abstract contract TransactionManager is BaseModule {
...
 
     /** ******************* Callbacks ************************** */
 
+    /**
+     * @notice Returns true if this contract implements the interface defined by
+     * `interfaceId` (see https://eips.ethereum.org/EIPS/eip-165).
+     */
+    function supportsInterface(bytes4 interfaceID) external view returns (bool) {
+        return  interfaceID == ERC165_INTERFACE || interfaceID == ERC1155_INTERFACE;          
+    }
```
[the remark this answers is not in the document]
olivdb: I measured it at less than 7500 gas. Adding a test to show that.

contracts/modules/TransactionManager.sol
elenadimitrova: I don't think there's any limitation around assembly return. Doesn't `return (0, 0x04)` work?

test/static-calls.js
```diff
@@ -164,8 +165,11 @@ contract("Static Calls", (accounts) => {
       assert.isTrue(success, "enableERC1155TokenReceiver failed");
       await checkStaticCalls({ _wallet: oldWallet, _supportERC1155: true });
     });
-    it("supports ERC1155TokenReceiver by default (new wallet)", async () => {
-      await checkStaticCalls({ _wallet: wallet, _supportERC1155: true });
+
+    it("requires less than 10000 gas to call supportsInterface()", async () => {
+      // this call will fail if supportsInterface() consumes more than 10000 gas
+      const txReceipt = await (await ERC165Tester.new()).testERC165Gas(wallet.address, "0x4e2312e0");
+      console.log(`supportsInterface() costs less than ${txReceipt.logs[0].args._gas.toString()} gas`);
```
elenadimitrova, line 172: Great, just the messaging is a bit off > supportsInterface() costs less than x gas -> supportsInterface() costs x gas

contracts/modules/TransactionManager.sol
olivdb: No :(

test/static-calls.js
olivdb: It's actually an upper bound. The `gasleft()` adds a bit of gas that makes the amount emitted in the event slightly higher than the real cost (the amount logged is 8876 but you can see that the method still succeeds when called with only 7500 gas).

juniset (approved):

contracts/modules/TransactionManager.sol
olivdb: Summary of discussion with Elena:
1) `return (0, 0x04)` works in `truffle` but makes `web3.eth.call()` return a `byte4` for static calls implemented via the module's fallback, which is inconsistent with the `byte32` normally returned by  `web3.eth.call()` when calling static methods returning a `byte4` but implemented _outside_ the fallback (i.e. with a proper method declaration).
2) `return (0, 0x04)` does not work in remix and yields `Failed to decode output: Error: data out-of-bounds (length=4, offset=32, code=BUFFER_OVERRUN, version=abi/5.0.7)` whereas `return (0, 0x20)` gives the exact same result as if the method was implemented with a proper method declaration.

Based on these observations, it seems to me that the EVM normally returns a "zero-padded `bytes32`" for when returning `bytes4`, even though libraries may reformat this "zero-padded `bytes32`" as a smaller type (e.g. bytes4) based on the available ABI.

In conclusion, to be safe, and since the goal of the fallback is to emulate the usage of a proper method declaration, I propose to use `return (0, 0x20)`.

elenadimitrova (approved):

PR #196, opened 2021-02-18
Authorisation sessions

elenadimitrova:
Makes the sessions logic skeleton implemented in #187 cleaner and more explicit. There are two methods now:
- `multiCallWithGuardians` - doing a multi-call with guardian approval - the equivalent of the previous `ApprovedTransfer` functionality. This additionally can include a call to `startSession` to authorise a certain user to transaction for a certain period of time without guardian approval. 
- `multiCallWithSession` - transactions signed by a pre-authorised session user (via `startSession`)

There is at most one session per wallet allowed at any given time. Wallet owner alone can cancel existing session.

Additionally we've disable adding the wallet and modules to the whitelist in order to prevent bypassing security for `startSession` calls.


contracts/modules/ArgentModule.sol
```diff
@@ -101,6 +98,19 @@ contract ArgentModule is BaseModule, RelayerManager, SecurityManager, Transactio
             // any guardian
             return (1, OwnerSignature.Disallowed);
         }
+        if (methodId == TransactionManager.multiCallWithApproval.selector) {
```
juniset, line 101: This should be moved up the IFs as we anticipate it will be a frequent call.

contracts/modules/RelayerManager.sol
```diff
@@ -268,19 +254,12 @@ abstract contract RelayerManager is BaseModule {
...
-    function validateSignatures(
-        address _wallet,
-        bytes32 _signHash,
-        bytes memory _signatures,
-        OwnerSignature _option
-    )
-        internal
-        view
-        returns (bool)
+    function validateSignatures(address _wallet, bytes32 _signHash, bytes memory _signatures, OwnerSignature _option) internal view
```
juniset, line 261: Why the new format for long method definitions? Can we stick to the current convention which we use everywhere in the code?

contracts/modules/RelayerManager.sol
```diff
@@ -343,7 +322,7 @@ abstract contract RelayerManager is BaseModule {
     )
         internal
     {
-        if (_gasPrice > 0 && (_option == OwnerSignature.Required || _option == OwnerSignature.Session)) {
+        if (_gasPrice > 0 && (_option == OwnerSignature.Required || ownerSignatureRequirement == OwnerSignature.Session)) {
```
juniset, line 325: Should be `_option` and not `ownerSignatureRequirement`.

contracts/modules/TransactionManager.sol
```diff
@@ -214,7 +213,9 @@ abstract contract TransactionManager is BaseModule {
 
     function isAuthorised(address _wallet, address _spender, address _to, bytes calldata _data) internal view returns (bool) {
         if (_to == _spender) {
-            return authoriser.authorise(_wallet, _to, _data); // do we need to block calls to the wallet or modules?
+            return authoriser.authorise(_wallet, _to, _data); 
+            // do we need to block calls to the wallet or modules?
+            // we do need to block calls to modules for the purposes of not being able to bypass calls to startSession
```
juniset, line 218: If we have to do that then it will add some gas overhead to every call going through `multiCall`.

juniset (commented):
I did some quick gas benchmark and it seems that this PR adds ~16k gas to create a session but also to do a simple ETH transfer with a session key (still need to be confirmed). We can live with the added gas to create a session, but probably not to such a gas increase to every call with a session key.

juniset:
If the goal is to be cleaner than I think we should split `multiCallWithApproval` into 2 external methods to get rid of the `useSession` flag:
- function `multiCallWithGuardians(address _wallet, Call[] calldata _transactions) external `
- function `multiCallWithSession(address _wallet, Call[] calldata _transactions) external `
And keep the logic in an internal method to limit the added bytecode:
- function `multiCallWithApproval(address _wallet, Call[] calldata _transactions) internal `

Also, the definition of the `Session` struct should be moved to the `RelayerManager`.

contracts/modules/TransactionManager.sol
juniset: I think it is ok to let the `Authoriser` do the authorisation and not add any constraint here. If the wallet and the module are not explicitly authorised in the `Authoriser` then the call will be blocked as expected.

However, we should probably prevent the owner to add the wallet and the module in its whitelist, i.e. add a check in `addToWhitelist`.

contracts/modules/RelayerManager.sol
elenadimitrova: That is the convention we agreed a year ago but isn't applied here to use 145 chars and/or 5 parameters on single line and only drop to multi line otherwise.

juniset: As per the discussion we agreed to "... have either everything on the same line, of drop everything on their own lines. But I don't like dropping only the modifiers.".

So to me me explicitly agreed not to use the above convention.

contracts/modules/RelayerManager.sol
```diff
@@ -387,25 +371,26 @@ abstract contract RelayerManager is BaseModule {
...
-        require(numberOfGuardians > 0, "RM: no guardians set on wallet");
-        uint numberOfSignatures = 1 + numberOfGuardians;
-        return (numberOfSignatures, OwnerSignature.Required);
+    function startSession(address _wallet, address _sessionUser, uint64 _duration) external 
+    onlyWallet(_wallet) returns (bool)
+    {
+        require(_sessionUser != address(0), "RM: Invalid session user");
+        require(_duration > 0, "RM: Invalid session duration");
+
+        uint64 expiry = Utils.safe64(block.timestamp + _duration);
```
juniset, line 380: I'm not sure if basing the expiration of a session on the current timestamp is a good idea in terms of security. What happens if I'm starting an 1 hour session but the transaction takes 30 minutes to be mined... I may end up having an open session when I didn't want to. Maybe more secure to have the expiry absolute?

contracts/modules/TransactionManager.sol
```diff
@@ -48,36 +48,40 @@ abstract contract TransactionManager is BaseModule {
...
+
+    function multiCallWithGuardians(address _wallet, Call[] calldata _transactions) external
+    onlySelf()
+    onlyWhenUnlocked(_wallet)
+    returns (bytes[] memory)
+    {
+        multiCallWithApproval(_wallet, false, _transactions);
+    }
+
+    function multiCallWithApproval(address _wallet, bool _useSession, Call[] calldata _transactions) internal
```
juniset, line 83: `_useSession` is no longer necessary.

contracts/modules/TransactionManager.sol
```diff
@@ -91,15 +95,13 @@ abstract contract TransactionManager is BaseModule {
...
-        external
-        onlyWalletOwnerOrSelf(_wallet)
-        onlyWhenUnlocked(_wallet)
+    function addToWhitelist(address _wallet, address _target) external
+    onlyWalletOwnerOrSelf(_wallet)
+    onlyWhenUnlocked(_wallet)
     {
-        require(!isWhitelisted(_wallet, _target), "TT: target already whitelisted");
+        require(_target != _wallet, "TM: Cannot whitelist wallet");
+        require(!IWallet(_wallet).authorised(_target), "TM: Cannot whitelist module");
```
juniset, line 103: Shouldn't we just do `require(_target != address(this))` to save some gas since there is only 1 module?

juniset (changes requested):

contracts/modules/RelayerManager.sol
elenadimitrova: I've changed them to one line then.
Unfortunately I can't find a solhint rule for that so we can apply it consistently.

juniset: Sorry to insist but can we have everything in one line then? The `returns` is still in a separate line for several methods (all the `multiCall`s and `validateSignatures`) which is not consistent and harder to read.

contracts/modules/RelayerManager.sol
elenadimitrova: I changed this logic for a similar reason that if a block takes 30 min to mine and I want my session to be 30 min long then it would have expired by the time it's committed. I think it's going to be more difficult and error prone if we make the client adjust session lengths based on current block times rather than the opposite.

contracts/modules/TransactionManager.sol
elenadimitrova: To help us make a decision the measured difference in gas is 4,580 (88,404 vs 83,824).

contracts/modules/RelayerManager.sol
```diff
@@ -387,25 +370,24 @@ abstract contract RelayerManager is BaseModule {
         assembly { chainId := chainid() }
     }
 
-    function getRequiredSignaturesToStartSession(address _wallet) internal view returns (uint256, OwnerSignature) {
-        // owner  + [n/2] guardians
-        uint numberOfGuardians = Utils.ceil(guardianStorage.guardianCount(_wallet), 2);
-        require(numberOfGuardians > 0, "RM: no guardians set on wallet");
-        uint numberOfSignatures = 1 + numberOfGuardians;
-        return (numberOfSignatures, OwnerSignature.Required);
+    function startSession(address _wallet, address _sessionUser, uint64 _duration) external onlyWallet(_wallet) returns (bool)
```
olivdb, line 373: `return (bool)` is no longe necessary

contracts/modules/RelayerManager.sol
```diff
@@ -387,25 +370,24 @@ abstract contract RelayerManager is BaseModule {
...
-        }
-        return false;
+    /**
+    * @notice Clears the active session of a wallet if any.
+    * @param _wallet The target wallet.
+    */
+    function clearSession(address _wallet) external onlySelf()
+    {
+        delete sessions[_wallet];
+        emit SessionCleared(_wallet, sessions[_wallet].key);
```
olivdb, line 390: The second argument logged will always be `address(0)` since the session has been deleted.

contracts/modules/RelayerManager.sol
olivdb: I'm probably missing something but why don't we make this method `internal` and change `multiCallWithGuardians` to:

```
function multiCallWithGuardians(address _wallet, Call[] calldata _transactions, Session calldata _session) external onlySelf() onlyWhenUnlocked(_wallet) returns (bytes[] memory)
{
        multiCallWithApproval(_wallet, _transactions);
        if(_session.key != address(0)) {
            startSession(_wallet, _session.key, _session.expires);
        }
 }
 ```
 in order to save the gas required for the Proxy->BaseWallet->ArgentModule calls?

**Review by juniset — APPROVED** [review]
Approving it provided that we fix the method declaration inconsistencies.

File: contracts/modules/TransactionManager.sol [hunk 10]
```diff
@@ -62,44 +64,61 @@ abstract contract TransactionManager is BaseModule {
...
+        onlyWhenUnlocked(_wallet)
+        returns (bytes[] memory)
+    {
+        multiCallWithApproval(_wallet, _transactions);
+    }
+
+    function multiCallWithGuardiansAndStartSession(
+        address _wallet,
+        address _sessionUser,
+        uint64 _duration,
```
juniset, line 100: Would make sense to move the session arguments after the `_transactions` argument to be consistent with the other multicall methods. Also, why not use a `Session` object and have:
```
function multiCallWithGuardiansAndStartSession(
        address _wallet,
        Call[] calldata _transactions,
        Session _session
    )
```

elenadimitrova: Ordering was only because when you just want to start a session, the `transactions` object will be empty and made sense to leave it at the end. I don't mind either way.

elenadimitrova: The `Session` object contains the expiration time whereas here we are passing the duration. They are both uint's so it will work but it's confusing to use these interchangeably.

olivdb (approved):

PR #197, opened 2021-02-23
Registry of authorised contracts

olivdb:


contracts/infrastructure/DappRegistry.sol
```diff
@@ -7,69 +7,72 @@ import "./storage/Storage.sol";
...
         return false;
     }
 
-    function toggle(address _wallet, bytes32 _registry) external override onlyModule(_wallet) returns (bool) {
-        require(_registry == ARGENT_DEFAULT || registryManager[_registry] != address(0), "AR: unknow registry");
-        bool current = isEnabled[_wallet][_registry];
-        isEnabled[_wallet][_registry] = !current;
-        return !current;
+    function toggleRegistry(address _wallet, uint8 _registryId, bool _enabled) external override onlyModule(_wallet) returns (bool) {
+        require(_registryId == 0 /* Argent Default Registry */ || registryManagers[_registryId] != address(0), "DR: unknow registry");
```
th3m477, line 33: ==?

contracts/modules/TransactionManager.sol
```diff
@@ -214,9 +215,9 @@ abstract contract TransactionManager is BaseModule {
 
     function isAuthorised(address _wallet, address _spender, address _to, bytes calldata _data) internal view returns (bool) {
         if (_to == _spender) {
```
juniset, line 217: Coming back to your comment https://github.com/argentlabs/argent-contracts/pull/187#discussion_r573975163 , I completely agree that this distinction is not a good idea and we should not have restrictions in the `DappRegistry` but let the `Filter` decide for maximum flexibility.

However, in the current implementation of the registry the `getFilter` method checks the existence of a filter for the `_to` address which in the case of an ERC20 method will be the token, while the filter is associated to the `_spender`. So we need to update the logic for these scenarios.

PR #199, opened 2021-02-25
Minimal wallet: DappRegistry

olivdb:
- uses bit vectors for more gas-efficient reads of authorisations and writes of enabled registries
- implements timelocks for updating dapp filters and changing timelock


contracts/infrastructure/DappRegistry.sol
```diff
@@ -7,72 +7,193 @@ import "./storage/Storage.sol";
...
     
-    // mapping [wallet][name][status]
-    mapping (address => mapping (bytes32 => bool)) public isEnabled;
-    // mapping [name][address][authorisation]
-    mapping (bytes32 => mapping (address => Authorisation)) public authorisations;
-
-    bytes32[] public registries;
+    event RegistryCreated(uint8 registryId, address manager);
+    event RegistryRemoved(uint8 registryId);
+    event TimelockChangePended(uint64 newSecurityPeriod);
```
th3m477, line 28: what is Pended? :)

contracts/infrastructure/DappRegistry.sol
```diff
@@ -7,72 +7,193 @@ import "./storage/Storage.sol";
...
-    // Do we want to let the owner to delete a registry?
-    function createRegistry(bytes32 _registry, address _manager) external onlyOwner {
-        require(_registry != bytes32(0) && _manager != address(0), "AR: invalid parameters");
-        require(registryManager[_registry] == address(0), "AR: duplicate registry");
-        registries.push(_registry);
-        registryManager[_registry] = _manager;
-        emit RegistryCreated(_registry, _manager);
+    function confirmTimelockChange() external {
+        uint64 newPeriod = newSecurityPeriod;
+        require(newPeriod > 0 && securityPeriodChangeAfter <= block.timestamp, "AR: can't (yet) change timelock");
```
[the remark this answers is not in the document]
olivdb: This check was meant to only allow calls to this method (which can be called by anyone) when a timelock change is pending. I wanted to allow changing the timelock to 0 if needed, so I should have used `securityPeriodChangeAfter` instead of `newPeriod` here:
```
require(securityPeriodChangeAfter > 0 && securityPeriodChangeAfter <= block.timestamp, "AR: can't (yet) change timelock");
```

contracts/infrastructure/DappRegistry.sol
```diff
@@ -7,72 +7,193 @@ import "./storage/Storage.sol";
...
+        require(_registryId > 0 && _manager != address(0), "AR: invalid parameters");
+        require(registryManagers[_registryId] == address(0), "AR: duplicate registry");
+        registryManagers[_registryId] = _manager;
+        emit RegistryCreated(_registryId, _manager);
+    }
+
+    // Note: removeRegistry is not supported because that would allow the owner to replace registries that 
+    // have already been enabled by users with a new (potentially maliciously populated) registry 
+
+    function changeManager(uint8 _registryId, address _newManager) external {
```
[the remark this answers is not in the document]
olivdb: As discussed, I think this is best handled by the manager smart contract. I would prefer not to bloat DappRegistry more than necessary.

contracts/infrastructure/DappRegistry.sol
```diff
@@ -7,72 +7,193 @@ import "./storage/Storage.sol";
...
+    function removeDapp(uint8 _registryId, address _dapp) external {
+        validateManager(_registryId);
+        require(authorisations[_registryId][_dapp] != bytes32(0), "AR: unknown dapp");
+        delete authorisations[_registryId][_dapp];
+        emit DappRemoved(_registryId, _dapp);
+    }
+
+    /********  Internal Functions ***********/
+
+    function validateManager(uint8 _registryId) internal {
```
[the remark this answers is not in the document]
elenadimitrova: Also `registryManagers` can be `registryOwners`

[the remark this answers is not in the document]
elenadimitrova: Maybe even make that internal function a modifier.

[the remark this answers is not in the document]
olivdb: Done

contracts/infrastructure/DappRegistry.sol
```diff
@@ -7,72 +7,193 @@ import "./storage/Storage.sol";
...
         return false;
     }
 
-    function toggle(address _wallet, bytes32 _registry) external override onlyModule(_wallet) returns (bool) {
-        require(_registry == ARGENT_DEFAULT || registryManager[_registry] != address(0), "AR: unknow registry");
-        bool current = isEnabled[_wallet][_registry];
-        isEnabled[_wallet][_registry] = !current;
-        return !current;
+    function toggleRegistry(address _wallet, uint8 _registryId, bool _enabled) external override onlyModule(_wallet) returns (bool) {
+        require(_registryId == 0 /* Argent Default Registry */ || registryManagers[_registryId] != address(0), "AR: unknow registry");
```
elenadimitrova, line 55: typo -> unknow

contracts/infrastructure/DappRegistry.sol
```diff
@@ -2,77 +2,197 @@ pragma solidity ^0.6.12;
...
-        address filter;
-    }
+    // bit vector of enabled registry ids for each wallet
+    mapping (address => bytes32) public enabledRegistryIds; // [wallet] => [bit vector of 256 registry ids]
+    // authorised dapps and their filters for each registry id
+    mapping (uint8 => mapping (address => bytes32)) public authorisations; // [registryId] => [dapp] => [{filter:160}{validAfter:64}]
+    // pending authorised dapps and their filters for each registry id
+    mapping (uint8 => mapping (address => bytes32)) public pendingFilterUpdates; // [registryId] => [dapp] => [{filter:160}{validAfter:64}]
+    // owners for each registry id
+    mapping (uint8 => address) public registryOwners; // [registryId] => [owner]
```
elenadimitrova, line 23: Since we cannot remove or overwrite registries is it possible that we will run out of these 256 registry ids at some point?

juniset: No, that shouldn't be an issue since we currently plan to use only 2 registries. Even if we change our minds it will be more than enough.

olivdb: Yes, in theory. In practice we do not expect using more than 2 or 3 registries.

contracts/infrastructure/DappRegistry.sol
```diff
@@ -2,77 +2,197 @@ pragma solidity ^0.6.12;
...
-    function toggle(address _wallet, bytes32 _registry) external override onlyModule(_wallet) returns (bool) {
-        require(_registry == ARGENT_DEFAULT || registryManager[_registry] != address(0), "AR: unknow registry");
-        bool current = isEnabled[_wallet][_registry];
-        isEnabled[_wallet][_registry] = !current;
-        return !current;
+    function toggleRegistry(address _wallet, uint8 _registryId, bool _enabled) external override onlyModule(_wallet) returns (bool) {
+        require(registryOwners[_registryId] != address(0), "AR: unknown registry");
+        uint registries = uint(enabledRegistryIds[_wallet]);
+        bool current = ((registries >> _registryId) & 1) > 0;
+        require(current != _enabled, "AR: bad state change" );
```
elenadimitrova, line 66: We can do without these two lines and save ~2,000 gas. I can't see any harm that could be done with someone trying to toggle a registry setting to an existing value. The `^` would just result in the same value being passed in `enabledRegistryIds[_wallet]`. The cost of toggling between differencing values is ~47,000 and to an existing value is ~15,000 gas.

contracts/infrastructure/DappRegistry.sol
```diff
@@ -2,77 +2,197 @@ pragma solidity ^0.6.12;
...
+    }
 
-    mapping (bytes32 => address) public registryManager;
+    /********* Wallet-centered functions *************/
 
-    event RegistryCreated(bytes32 registry, address manager);
-    event RegistryRemoved(bytes32 registry);
+    function isEnabledRegistry(address _wallet, uint8 _registryId) external view returns (bool) {
+        uint registries = uint(enabledRegistryIds[_wallet]);
+        return ((registries >> _registryId) & 1) > 0;
```
elenadimitrova, line 51: Remix does something odd and allows this to overflow, i.e. this returns `true` there for a registryId > 255. I'm pretty sure this can't be the case with the right shift here but putting a note to add a test for that.

contracts/infrastructure/DappRegistry.sol
```diff
@@ -2,77 +2,197 @@ pragma solidity ^0.6.12;
...
+    * that has previously been authorised _without_ a filter. To change a non-zero filter, use `requestFilterUpdate`
+    * and `confirmFilterUpdate` instead.
+    * @param _registryId The id of the registry to modify
+    * @param _dapp The address of the dapp contract to authorise.
+    * @param _filter The address of the filter contract to use.
+    */
+    function addFilter(uint8 _registryId, address _dapp, address _filter) external onlyOwner(_registryId) {
+        uint auth = uint(authorisations[_registryId][_dapp]); // {filter:160}{validFrom:64}
+        require((auth >> 64) == 0, "DR: filter already set");
+        uint validFrom = auth & 0xffffffffffffffff;
```
elenadimitrova, line 116: Just for clarity can we rename `validFrom` to `validAfter`? One is inclusive of the timestamp and the other is not, and that's the name used in declaration notes and other functions here.

contracts/infrastructure/DappRegistry.sol
```diff
@@ -2,77 +2,197 @@ pragma solidity ^0.6.12;
...
+    * @notice Confirm the filter change requested by `requestFilterUpdate`
+    * @param _registryId The id of the registry to modify
+    * @param _dapp The address of the dapp contract to authorise.
+    */
+    function confirmFilterUpdate(uint8 _registryId, address _dapp) external {
+        uint newAuth = uint(pendingFilterUpdates[_registryId][_dapp]);
+        require(newAuth > 0, "AR: no pending filter update");
+        uint validFrom = newAuth & 0xffffffffffffffff;
+        require(validFrom <= block.timestamp, "AR: too early to confirm auth");
+        pendingFilterUpdates[_registryId][_dapp] = bytes32(newAuth);
```
elenadimitrova, line 157: This should be updating `authorisations`

elenadimitrova (commented):
I'm finding the bit vector logic a bit of an overkill especially for the filters logic. It might make sense for the registries logic in case we end up having many more than 2 in there, as we can more cheaply compose changes to enabled registries. However for the filters logic the storage is already going to be optimised into a single storage slot for `(address,uint64)`.

contracts/infrastructure/DappRegistry.sol
olivdb: This was introduced to meet a requirement from @th3m477 of having explicit enable/disable calls instead of allowing the same function call to do both. I assume this is to mitigate the risk of having a client mistakenly sign the same toggle transaction twice? I think this was discussed last Monday when I was off so I'll let Matt explain the his reasons :)

contracts/infrastructure/DappRegistry.sol
```diff
@@ -2,77 +2,197 @@ pragma solidity ^0.6.12;
...
+        return ((registries >> _registryId) & 1) > 0;
+    }
 
-    function authorise(address _wallet, address _contract, bytes calldata _data) external view override returns (bool) {
-        (bool isActive, address filter) = getFilter(_wallet, _contract);
+    function isAuthorised(address _wallet, address _spender, address _to, bytes calldata _data) external view override returns (bool) {
+        (bool isActive, address filter) = getFilter(_wallet, _spender);
         if (isActive) {
-            return _data.length == 0 || filter == address(0) || IFilter(filter).validate(_data);
+            return _data.length == 0 || filter == address(0) || IFilter(filter).validate(_wallet, _spender, _to, _data);
```
juniset, line 57: I think we should remove the condition _data.length == 0 and let the filter decide if it accepts ETH transfers.

contracts/infrastructure/DappRegistry.sol
```diff
@@ -2,77 +2,197 @@ pragma solidity ^0.6.12;
...
         }
         return false;
     }
 
-    function toggle(address _wallet, bytes32 _registry) external override onlyModule(_wallet) returns (bool) {
-        require(_registry == ARGENT_DEFAULT || registryManager[_registry] != address(0), "AR: unknow registry");
-        bool current = isEnabled[_wallet][_registry];
-        isEnabled[_wallet][_registry] = !current;
-        return !current;
+    function toggleRegistry(address _wallet, uint8 _registryId, bool _enabled) external override onlyModule(_wallet) returns (bool) {
```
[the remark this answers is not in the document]
olivdb: Yes, I like the idea of making it usable by other wallets

contracts/infrastructure/DappRegistry.sol
```diff
@@ -2,77 +2,197 @@ pragma solidity ^0.6.12;
...
+        pendingFilterUpdates[_registryId][_dapp] = bytes32((uint(uint160(_filter)) << 64) | validAfter);
+        emit FilterUpdateRequested(_registryId, _dapp, _filter, validAfter);
+    }
+
+    /**
+    * @notice Confirm the filter change requested by `requestFilterUpdate`
+    * @param _registryId The id of the registry to modify
+    * @param _dapp The address of the dapp contract to authorise.
+    */
+    function confirmFilterUpdate(uint8 _registryId, address _dapp) external {
```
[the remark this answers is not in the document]
olivdb: Not necessary I think. If a filter update needs to be cancelled, the registry owner can just call `requestFilterUpdate` with the correct filter, which will automatically override the previous pending filter update, making it unconfirmable.

contracts/infrastructure/DappRegistry.sol
```diff
@@ -2,77 +2,197 @@ pragma solidity ^0.6.12;
...
+
+    /**
+    * @notice Set an authorisation filter for a dapp that has _not_ previously been authorised or 
+    * that has previously been authorised _without_ a filter. To change a non-zero filter, use `requestFilterUpdate`
+    * and `confirmFilterUpdate` instead.
+    * @param _registryId The id of the registry to modify
+    * @param _dapp The address of the dapp contract to authorise.
+    * @param _filter The address of the filter contract to use.
+    */
+    function addFilter(uint8 _registryId, address _dapp, address _filter) external onlyOwner(_registryId) {
```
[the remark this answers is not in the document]
olivdb: Fine, going from 0 to non-zero filter is probably going to be a rare occurrence anyways. We can do without this optimisation.

contracts/infrastructure/DappRegistry.sol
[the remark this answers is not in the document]
juniset: I was thinking of a scenario where we realise the filter needs to be updated during the timelock period but the fix will not be ready when the period ends. But in that case I guess we can always call `requestFilterUpdate` with the existing filter in the registry that we are trying to update until the new one is ready.

Agreed it is not needed.

deployment/2_deploy_contracts.js
```diff
@@ -59,7 +59,7 @@ async function main() {
 
   // Deploy Module Registry
   const ModuleRegistryWrapper = await ModuleRegistry.new();
-  const AuthoriserWrapper = await Authoriser.new();
+  const AuthoriserWrapper = await Authoriser.new(newConfig.settings.securityPeriod);
```
juniset, line 62: We should add a new `timelockPeriod`  entry in the config and that's what the constructor should use.

contracts/infrastructure/DappRegistry.sol
```diff
@@ -2,77 +2,212 @@ pragma solidity ^0.6.12;
...
 
-    function getFilter(address _wallet, address _contract) internal view returns (bool, address) {
-        // check argent default registry first (enbaled by default, i.e. false => true)
-        if (!isEnabled[_wallet][ARGENT_DEFAULT] && authorisations[ARGENT_DEFAULT][_contract].isActive) {
-            return (true, authorisations[ARGENT_DEFAULT][_contract].filter);
-        } else {
-            for (uint i = 0; i < registries.length; i++) {
-                if (isEnabled[_wallet][registries[i]] && authorisations[registries[i]][_contract].isActive) {
-                    return (true, authorisations[registries[i]][_contract].filter);
+    function getFilter(address _wallet, address _dapp) internal view returns (bool, address) {
```
juniset, line 220: A valid point raised by the client teams is that if a dapp is authorised on multiple registries but with different filters (e.g. registry[0] authorises buy/sell on Aave, while registry[1] athorises staking) the second one will be ignored.

We could fix that by having the `getFilter` method return an array of matching filters, and loop through the filters in `isAuthorised` until one of them passes.

contracts/infrastructure/DappRegistry.sol
olivdb: Looks to me like solidity uses `_registryId % 256` for `_registryId` when the function is called with `_registryId` > 256. So it's expected that `isEnabledRegistry(address _wallet, uint8 _registryId)` is going to return true for `uint` values of `_registryId` such that `isEnabledRegistry(_wallet, _registryId % 256) == true`. I verified that in a test:
```
const dataWithBadSig = utils.encodeFunctionCall("isEnabledRegistry(address,uint)", [wallet.address, 256 + CUSTOM_REGISTRY_ID]);
      const correctSig = web3.utils.sha3("isEnabledRegistry(address,uint8)").slice(0, 10);
      const data = `${correctSig}${dataWithBadSig.slice(10)}`;
      console.log({ dataWithBadSig, correctSig, data });
      const output = await web3.eth.call({
        to: authoriser.address,
        data
      });
      assert.equal(output.toString(), "0x0000000000000000000000000000000000000000000000000000000000000001");
```

contracts/infrastructure/DappRegistry.sol
olivdb: I don't think it's worth the added gas cost of having to systematically do N reads for N filters. Besides, philosophically, I think of a filter as the answer to the question "what is the set of method calls on this contract such that all assets stay under the control of the wallet?". There should be a precise answer to this question for each dapp. So in the case of Aave, after reviewing the Aave contract, we may determine that minting, redeeming _and_ staking are all permissible actions, and would define a filter that reflects that. Partial filters split across registries would in my opinion indicate that multiple registries disagree on the above question. But such disagreement should be resolved off-chain IMO.

contracts/infrastructure/DappRegistry.sol
```diff
@@ -2,77 +2,236 @@ pragma solidity ^0.6.12;
...
+    mapping (uint8 => mapping (address => bytes32)) public pendingFilterUpdates; // [registryId] => [dapp] => [{filter:160}{validAfter:64}]
+    // owners for each registry id
+    mapping (uint8 => address) public registryOwners; // [registryId] => [owner]
     
-    // mapping [wallet][name][status]
-    mapping (address => mapping (bytes32 => bool)) public isEnabled;
-    // mapping [name][address][authorisation]
-    mapping (bytes32 => mapping (address => Authorisation)) public authorisations;
+    event RegistryCreated(uint8 registryId, address registryOwner);
+    event RegistryRemoved(uint8 registryId);
```
juniset, line 25: This event is not used and can be removed.

deployment/2_deploy_contracts.js
```diff
@@ -59,7 +59,7 @@ async function main() {
 
   // Deploy Module Registry
   const ModuleRegistryWrapper = await ModuleRegistry.new();
-  const AuthoriserWrapper = await Authoriser.new();
+  const AuthoriserWrapper = await Authoriser.new(newConfig.settings.timelockPeriod);
```
juniset, line 62: Since this contract is deployed for the first time and is part of the 2.5 release I feel this line should be part of the `7_upgrade_2.5.js` file .

contracts/infrastructure/DappRegistry.sol
elenadimitrova: Had a conversation with Matt and Julien on the call and decided we can remove this logic.

th3m477: @olivdb it's ok to not have the check 👍 

i just wanted `_enabled` passed in which i can see you do, so I am happy

juniset (approved):

contracts/infrastructure/DappRegistry.sol
olivdb: I changed the `require` into an `if`

deployment/7_deploy_2.5.js
```diff
@@ -103,6 +107,7 @@ const main = async () => {
   configurator.updateInfrastructureAddresses({
     WalletFactory: WalletFactoryWrapper.address,
     BaseWallet: BaseWalletWrapper.address,
+    Authoriser: AuthoriserWrapper.address,
```
elenadimitrova, line 110: We should add that to the schema as well although might leave it for the cleanup we'll do later.

contracts/infrastructure/DappRegistry.sol
elenadimitrova: To save all the gas we can get rid of that `if`. I don't see the point of that check. If the value has no change, then no storage will be written.

contracts/infrastructure/DappRegistry.sol
elenadimitrova: You mean that it's allowed to overflow?

elenadimitrova (approved):

contracts/infrastructure/DappRegistry.sol
olivdb: Without the `if`, the XOR will flip the bit no matter what, making it inconsistent with the value of `_enabled` when `current == _enabled`. An alternative to the if is to set/unset the bit based on `_enabled`
```
        if(_enabled) {
            enabledRegistryIds[msg.sender] = bytes32(registries | (uint(1) << _registryId));
        } else {
            enabledRegistryIds[msg.sender] = bytes32(registries & ~(uint(1) << _registryId));
        }
  ```
  Same gas cost but maybe a bit more verbose.

contracts/infrastructure/DappRegistry.sol
olivdb: Yes in the sense that the transaction won't revert if your "uint8" is > 255. It will just "wrap around" (i.e. use the value passed modulo 256).

PR #200, opened 2021-02-26
fixing broken test and reducing bytecode size

juniset:
This PR fixes a broken test and reduces the size of some reverts strings in `BaseModule`, `SecurityManager` and `TransactionManager` to make sure they fit in `bytes32`.

It also decreases the number of compiler runs from `999` to `800` to reduce the size of the `ArgentModule` bytecode under 24576 bytes.

### DIFF
--- a/contracts/modules/SecurityManager.sol
+++ b/contracts/modules/SecurityManager.sol
@@ -231,7 +231,7 @@ abstract contract SecurityManager is BaseModule {
      * @param _guardian The guardian to add.
      */
     function addGuardian(address _wallet, address _guardian) external onlyWalletOwnerOrSelf(_wallet) onlyWhenUnlocked(_wallet) {
-        require(!_isOwner(_wallet, _guardian), "SM: _guardian cannot be owner");
+        require(!_isOwner(_wallet, _guardian), "SM: guardian cannot be owner");
         require(!isGuardian(_wallet, _guardian), "SM: duplicate guardian");
         // Guardians must either be an EOA or a contract with an owner()
         // method that returns an address with a 5000 gas stipend.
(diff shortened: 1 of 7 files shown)


contracts/modules/SecurityManager.sol
```diff
@@ -232,13 +231,13 @@ abstract contract SecurityManager is BaseModule {
      * @param _guardian The guardian to add.
      */
     function addGuardian(address _wallet, address _guardian) external onlyWalletOwnerOrSelf(_wallet) onlyWhenUnlocked(_wallet) {
-        require(!_isOwner(_wallet, _guardian), "SM: target guardian cannot be owner");
-        require(!isGuardian(_wallet, _guardian), "SM: target is already a guardian");
+        require(!_isOwner(_wallet, _guardian), "SM: _guardian cannot be owner");
```
elenadimitrova, line 234: Can we avoid parameter names in the revert strings? It's un unnecessary dependency. We can just have _guardian -> guardian

contracts/modules/SecurityManager.sol
```diff
@@ -371,8 +370,8 @@ abstract contract SecurityManager is BaseModule {
     // *************** Internal Functions ********************* //
 
     function validateNewOwner(address _wallet, address _newOwner) internal view {
-        require(_newOwner != address(0), "SM: new owner address cannot be null");
-        require(!isGuardian(_wallet, _newOwner), "SM: new owner address cannot be a guardian");
+        require(_newOwner != address(0), "SM: _newOwner cannot be null");
```
elenadimitrova, line 373: Another parameter name used

contracts/modules/SecurityManager.sol
```diff
@@ -371,8 +370,8 @@ abstract contract SecurityManager is BaseModule {
...
+        require(!isGuardian(_wallet, _newOwner), "SM: _newOwner is guardian");
```
elenadimitrova, line 374: Another parameter name used

olivdb (approved):

elenadimitrova (approved):

PR #203, opened 2021-03-03
basic cleaning

juniset:
This PR:

- removes deprecated contracts and events
- adds or update method descriptions when needed
- makes the security parameters immutable
- add tests for send and receive of core assets (ETH, ERC20, ERC721, ERC1155)
- rename `isSpenderInData` into `isTokenCall` to be a bit more explicit


contracts/infrastructure/WalletFactory.sol
```diff
@@ -147,8 +148,8 @@ contract WalletFactory is Managed {
     }
 
     /**
-     * @notice Inits the module for a wallet by doing nothing.
-     * The method can only be called by the wallet itself.
+     * @notice Required to make the factory a module during the 
+     *  initialisation fo the wallet.
```
olivdb, line 152: typo: fo

contracts/modules/TransactionManager.sol
```diff
@@ -35,23 +36,37 @@ abstract contract TransactionManager is BaseModule {
...
 
     struct Call {
-        address to;
-        uint256 value;
-        bytes data;
-        bool isSpenderInData;
+        address to;         // the target of the call
+        uint256 value;      // the ETH to transfer
+        bytes data;         // the data payload
+        bool isTokenCall;   // true if the target is a token
```
olivdb, line 42: I'd rename `isTokenCall` to `isStandardTokenCall` and update the comment to "true if target is a token and data is a standard ERC20/721/1155 method" since SNX, aTokens, etc have non-ERC20 methods (e.g. `redeem(uint256)`) that would be called with `isTokenCall = false`

juniset: I've updated the comment but not the parameter name to keep it short. I believe the updated comment is sufficient to make its function clear.

elenadimitrova:
Maybe we can make better use of the `SafeMath` library and use their [`toUint128`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/math/SafeCast.sol#L31), `toUint64` and [`min`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/math/Math.sol#L19) functions. Which are exact

package.json
```diff
@@ -61,7 +61,7 @@
   },
   "homepage": "https://github.com/argentlabs/argent-contracts#readme",
   "dependencies": {
-    "@openzeppelin/contracts": "3.0.1",
+    "@openzeppelin/contracts": "3.4.0",
```
elenadimitrova, line 64: We can make better use of the `SafeMath` library and use their SafeCast methods introduced since [v3.1](https://github.com/OpenZeppelin/openzeppelin-contracts/releases/tag/v3.1.0). [`toUint128`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/math/SafeCast.sol#L31), `toUint64` and [`min`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/math/Math.sol#L19) functions. Which are exact replicas of what we have in `Utils` lib.

test/asset.js
```diff
@@ -0,0 +1,267 @@
...
+  }
+
+  async function whitelist(target) {
+    await module.addToWhitelist(wallet.address, target, { from: owner });
+    await utils.increaseTime(3);
+    const isTrusted = await module.isWhitelisted(wallet.address, target);
+    assert.isTrue(isTrusted, "should be trusted after the security period");
+  }
+
+  async function initNonce() {
```
elenadimitrova, line 120: Can we consolidate `encodeTransaction`,  `whitelist` and `initNonce` into `utils` library. They are duplicated across all tests.

utils/utilities.js
```diff
@@ -265,5 +256,34 @@ module.exports = {
...
+    const owner = await wallet.owner();
+    await module.addToWhitelist(wallet.address, target, { from: owner });
+    await utilities.increaseTime(securityPeriod + 2);
+    const isTrusted = await module.isWhitelisted(wallet.address, target);
+    assert.isTrue(isTrusted, "should be trusted after the security period");
+  },
+
+  // set the relayer nonce to > 0
+  initNonce: async (wallet, account, module, manager, securityPeriod) => {
+    await utilities.addTrustedContact(wallet, account, module, securityPeriod);
```
elenadimitrova, line 273: Can we simplify this helper and assume `account` is always `account[8]` for example, similarly with how we default the relayer account to account[9]. It makes no difference but keeps the function interface simpler.

utils/utilities.js
```diff
@@ -265,5 +256,34 @@ module.exports = {
...
+    const owner = await wallet.owner();
+    const transaction = utilities.encodeTransaction(account, 1, ZERO_BYTES32);
+    const txReceipt = await manager.relay(
+      module,
+      "multiCall",
+      [wallet.address, [transaction]],
+      wallet,
+      [owner]);
+    const success = await utilities.parseRelayReceipt(txReceipt).success;
```
elenadimitrova, line 282: There's no need to check this. We have the unit tests for verifying this works.

elenadimitrova (approved):

olivdb (approved):

PR #204, opened 2021-03-03
Safer recoverSpender

olivdb:
Changing the logic by which the spender is recovered to match transaction[i].data with ERC20 methods instead of using `transaction[i].isTokenCall`

Note: being able to fit in spender recovery support for ERC721 methods requires further bytecode reduction. Temporarily achieved via a reduction of the number of runs from 800 to 560.

### DIFF
--- a/contracts/modules/TransactionManager.sol
+++ b/contracts/modules/TransactionManager.sol
@@ -30,7 +30,7 @@ import "../../lib/other/ERC20.sol";
 abstract contract TransactionManager is BaseModule {
     // ERC20 & ERC721 transfers & approvals
     bytes4 private constant ERC20_TRANSFER = bytes4(keccak256("transfer(address,uint256)"));
-    bytes4 private constant ERC20_APPROVE = bytes4(keccak256("transferFrom(address,address,uint256)"));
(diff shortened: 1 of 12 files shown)


contracts/modules/TransactionManager.sol
```diff
@@ -275,9 +282,22 @@ abstract contract TransactionManager is BaseModule {
     }
 
     function recoverSpender(address _wallet, Call calldata _transaction) internal pure returns (address) {
-        if (_transaction.isTokenCall) {
+        bytes4 methodId;
```
juniset, line 285: Since the data is in calldata I think we can get the methodId by doing `bytes4 methodId = abi.decode(_transaction.data[:4], (bytes4));` which might be more readable and doesn't require to store the data in memory.

Also, we should skip all the logic if `_transaction.data.length < 4` which would correspond to a ETH transfer.

olivdb: We'd need to do `bytes32 res = abi.decode(abi.encodePacked(bytes28(0), _transaction.data), (bytes32));` which costs 500 extra gas compared to the assembly code. Agreed with adding the `_transaction.data.length < 4` check though.

juniset (approved):

PR #205, opened 2021-03-04
Fix legacy wallet factory

elenadimitrova:
Overload `ArgentENSManager.register` function to fix legacy factory integration. 

Update deployment script to set legacy wallet factory as manager of new `ArgentENSManager`

### DIFF
--- a/contracts/infrastructure/ens/ArgentENSManager.sol
+++ b/contracts/infrastructure/ens/ArgentENSManager.sol
@@ -66,8 +66,7 @@ contract ArgentENSManager is IENSManager, Owned, Managed {
     // *************** External Functions ********************* //
 
     /**
-     * @notice This function must be called when the ENS Manager contract is replaced and the address of the new Manager should be provided.
-     * @param _newOwner The address of the new ENS manager that will manage the root node.
+     * @inheritdoc IENSManager
      */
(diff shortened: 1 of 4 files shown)


contracts/infrastructure/ens/ArgentENSManager.sol
```diff
@@ -85,16 +84,23 @@ contract ArgentENSManager is IENSManager, Owned, Managed {
...
         bytes32 signedHash = keccak256(abi.encodePacked("\x19Ethereum Signed Message:\n32", keccak256(abi.encodePacked(_owner, _label))));
         validateManagerSignature(signedHash, _managerSignature);
 
+        _register(_label, _owner);
+    }
+
+    /**
+     * @inheritdoc IENSManager
+     */
+    function register(string calldata _label, address _owner) external override onlyManager validateENSLabel(_label) {
```
juniset, line 99: In `validateENSLabel` I believe it is not necessary to copy the bytes representation of the string to memory to get its length. We could just do:
``` 
modifier validateENSLabel(string memory _label) {
      require(bytes(_label).length != 0, "AEM: ENS label must be defined");
      _;
    }
```

juniset (approved):

PR #206, opened 2021-03-04
Clear session after a recovery is finalized

elenadimitrova:
When completing a recovery procedure any sessions are cleared.

### DIFF
--- a/contracts/modules/SecurityManager.sol
+++ b/contracts/modules/SecurityManager.sol
@@ -391,9 +391,4 @@ abstract contract SecurityManager is BaseModule {
     function _setLock(address _wallet, uint256 _releaseAfter, bytes4 _locker) internal {
         locks[_wallet] = Lock(SafeCast.toUint64(_releaseAfter), _locker);
     }
-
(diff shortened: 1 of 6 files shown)


contracts/modules/SecurityManager.sol
```diff
@@ -186,6 +188,14 @@ abstract contract SecurityManager is BaseModule {
         return (config.recovery, config.executeAfter, config.guardianCount);
     }
 
+    /**
+    * @notice Clears the active session of a wallet if any.
+    * @param _wallet The target wallet.
+    */
+    function clearSession(address _wallet) public onlyWalletOwnerOrSelf(_wallet) {
```
juniset, line 195: I find it a little inconsistent that creating session is on the `TransactionManager` and clearing a session is on the `SecurityManager`. They should be defined on the same contract.

Since the `session` object is declared on the `BaseModule` I would prefer to have the internal `_clearSession()` defined on the `BaseModule` and maintain the external `clearSession()` on the `TransactionManager`.

juniset (approved):

PR #207, opened 2021-03-04
Optimise guardian validation logic

elenadimitrova:
Slightly cleaner guardian validation logic in that guardians are looked up inside the `_isGuardianOrGuardianSigner` function instead of being passed to it. Also makes property names clearer.

### DIFF
--- a/contracts/modules/RelayerManager.sol
+++ b/contracts/modules/RelayerManager.sol
@@ -263,10 +263,6 @@ abstract contract RelayerManager is BaseModule, SimpleOracle {
             return true;
         }
         address lastSigner = address(0);
-        address[] memory guardians;
-        if (_option != OwnerSignature.Required || _signatures.length > 65) {
-            guardians = guardianStorage.getGuardians(_wallet); // guardians are only read if they may be needed
-        }
 
         for (uint256 i = 0; i < _signatures.length / 65; i++) {
             address signer = Utils.recoverSigner(_signHash, _signatures, i);
@@ -289,7 +285,7 @@ abstract contract RelayerManager is BaseModule, SimpleOracle {
                 return false; // Signers must be different
             }
             lastSigner = signer;
-            bool isGuardian = _isGuardianOrGuardianSigner(guardians, signer);
+            bool isGuardian = _isGuardianOrGuardianSigner(_wallet, signer);
             if (!isGuardian) {
                 return false;
             }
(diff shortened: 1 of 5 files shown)


contracts/modules/common/BaseModule.sol
```diff
@@ -164,6 +165,37 @@ abstract contract BaseModule is IModule {
...
+
+        if (guardians.length == 0 || _user == address(0)) {
+            return false;
+        }
+        bool isFound = false;
+        for (uint256 i = 0; i < guardians.length; i++) {
+            if (!isFound) {
+                // check if _user is an account guardian
+                if (_user == guardians[i]) {
+                    isFound = true;
```
th3m477, line 186: return in loop?

contracts/modules/RelayerManager.sol
```diff
@@ -291,7 +285,7 @@ abstract contract RelayerManager is BaseModule, SimpleOracle {
                 return false; // Signers must be different
             }
             lastSigner = signer;
-            (isGuardian, guardians) = Utils.isGuardianOrGuardianSigner(guardians, signer);
+            bool isGuardian = _isGuardianOrGuardianSigner(_wallet, signer);
```
olivdb, line 288: I think this no longer prevents duplicate guardian signatures

olivdb, line 288: This makes it possible for a smart contract guardian to sign more than once by changing what `owner()` returns between subsequent calls. Note that this is not something we test for, which is why our tests do not fail.

contracts/modules/common/BaseModule.sol
elenadimitrova: This is a copy of the function as we have it today. What is the problem exactly with return in loop?

olivdb: We could just replace the `isFound = true; continue;` with `return true` since we no longer need to iterate through the entire array (something we had to do to update `updatedGuardians`)

contracts/modules/RelayerManager.sol
olivdb: Actually, static call is not a protection. Here is an example of how a malicious smart contract guardian could still return multiple addresses:

```
pragma solidity ^0.8.1;

contract Guardian {
    function owner() external view returns (address) {
        uint idx = gasleft() % 10000;
        if(idx > 8000) return 0x5B38Da6a701c568545dCfcB03FcB875f56beddC4;
        if(idx > 4000) return 0xAb8483F64d9C6d1EcF9b849Ae677dD3315835cb2;
        return 0x4B20993Bc481177ec7E8f571ceCaE8A9e22C02db;
    }
}

contract Module {
    Guardian g;
    constructor() {
        g = new G();
    }
    function callGuardian() external view returns (address[] memory a) {
        a = new address[](3);
        for(uint i = 0; i < 3; i++) {
            a[i] = g.owner();
        }
    }
}
```

PR #208, opened 2021-03-04
Paraswap, AaveV2, Balancer, Yearn, DSR Filters

olivdb:
Includes implementation & tests for the following authorisation filters:
- Paraswap
- AaveV2
- Balancer
- Yearn
- DSR

### DIFF
--- a/contracts/infrastructure/dapp/AaveV2Filter.sol
+++ b/contracts/infrastructure/dapp/AaveV2Filter.sol
@@ -1,0 +1,54 @@
+// Copyright (C) 2021  Argent Labs Ltd. <https://argent.xyz>
+
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.s
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.6.12;
+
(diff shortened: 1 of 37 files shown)


juniset (approved):

contracts/infrastructure/dapp/AaveV2Filter.sol
```diff
@@ -0,0 +1,50 @@
...
+        if(method == DEPOSIT || method == WITHDRAW) {
+            (,, address beneficiary) = abi.decode(_data[4:], (address, uint256, address));   
+            return beneficiary == _wallet;
+        }
+
+        // only allow approve (LendingPool can only transfer a valid asset => no need to validate the token address here)
+        return (method == ERC20_APPROVE);
+    }
+
+    function getMethod(bytes memory _data) internal pure returns (bytes4 method) {
```
juniset, line 44: Maybe we should move this method to an abstract `BaseFilter` contract given that it will be used by all filters.

olivdb: True. But that's a useless require. Also, I like that we have tried so far to write filters in a way that doesn't rely on any other file than `IFilter` and `BaseFilter`, making it easier to publish the filters and `DappRegistry` in a standalone repo, should this become a reference registry used outside Argent.

test/paraswap.js
```diff
@@ -149,16 +150,14 @@ contract("ArgentModule", (accounts) => {
       SECURITY_WINDOW,
       RECOVERY_PERIOD,
       LOCK_PERIOD);
-
     await registry.registerModule(module.address, ethers.utils.formatBytes32String("ArgentModule"));
-
+    filter = await Filter.new(tokenPriceRegistry.address, dexRegistry.address);
+    await dappRegistry.addDapp(0, paraswap.address, filter.address);
+    await dappRegistry.addDapp(0, paraswapProxy, ZERO_ADDRESS);
```
juniset, line 156: Shouldn't we have a filter for the proxy as well?

juniset (approved):

elenadimitrova (commented):
As discussed I think best to integrate test these with a mainnet fork although happy to leave it on the tech debt backlog for now.

juniset (approved):

PR #209, opened 2021-03-04
Compound Filter

juniset:
This PR adds support for lending with Compound by:
-  adding the CompoundFilter
- adding tests to confirm minting and redeeming cTokens

### DIFF
--- a/contracts/infrastructure/dapp/CompoundFilter.sol
+++ b/contracts/infrastructure/dapp/CompoundFilter.sol
@@ -23,7 +23,6 @@ contract CompoundFilter is IFilter {
     bytes4 private constant CETH_MINT = bytes4(keccak256("mint()"));
     bytes4 private constant CERC20_MINT = bytes4(keccak256("mint(uint256)"));
     bytes4 private constant CTOKEN_REDEEM = bytes4(keccak256("redeem(uint256)"));
-    bytes4 private constant CTOKEN_REDEEM_UNDERLYING = bytes4(keccak256("redeemUnderlying(uint256)"));
     bytes4 private constant ERC20_APPROVE = bytes4(keccak256("approve(address,uint256)"));
 
     function isValid(address _wallet, address _spender, address _to, bytes calldata _data) external view override returns (bool) {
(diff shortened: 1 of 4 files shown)


contracts/infrastructure/dapp/CompoundFilter.sol
```diff
@@ -0,0 +1,51 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.6.12;
+
+import "./IFilter.sol";
+
+contract CompoundFilter is IFilter {
```
olivdb, line 21: It seems to me that the only method that needs to be forbidden is `repayBorrowBehalf`. Everything else is safe to allow (note that ERC20 methods of cToken will be caught in `recoverSpender` and do not need to be forbidden here). In particular why not allow `borrow`, `repayBorrow`, `redeemUnderlying`?

contracts/infrastructure/dapp/CompoundFilter.sol
```diff
@@ -0,0 +1,51 @@
...
+
+    bytes4 private constant CETH_MINT = bytes4(keccak256("mint()"));
+    bytes4 private constant CERC20_MINT = bytes4(keccak256("mint(uint256)"));
+    bytes4 private constant CTOKEN_REDEEM = bytes4(keccak256("redeem(uint256)"));
+    bytes4 private constant ERC20_APPROVE = bytes4(keccak256("approve(address,uint256)"));
+
+    function isValid(address _wallet, address _spender, address _to, bytes calldata _data) external view override returns (bool) {
+        // disable ETH transfer
+        if (_data.length < 4) {
```
olivdb, line 30: We could create the filter with a constructor parameter indicating whether this is a cErc20 or a cEther. For a cEther filter, it would be nice to enable ETH transfers (which invoke cEther's fallback) to enable deposits via simple transfers. That way, sending ETH to the cEther contract will work as well.

contracts/infrastructure/dapp/CompoundFilter.sol
juniset: Agreed for the CToken methods, we should only block `reapyBorrowBehalf`.

However I believe ERC20 token methods where the spender is the cToken (i.e. the ones which will end up in the filter) must be blocked because you want to prevent e.g. a transfer of DAI to cBAT. So `dai.transfer(cBAT, 1000)` must be blocked. The only one we should accept is an approve on the underlying to allow minting, i.e. `dai.approve(cDAI,1000)` .

olivdb (approved):

PR #212, opened 2021-03-08
Clean config and fixes to upgrade script

elenadimitrova:
Remove obsoleted properties in config and schema. 

Updates and fixes to v2.5 installation scrips.


deployment/7_deploy_2.5.js
```diff
@@ -76,6 +77,8 @@ const main = async () => {
   const DappRegistryWrapper = await DappRegistry.new(config.settings.timelockPeriod);
```
juniset, line 77: Shouldn't we create the DappRegistry with a timelock of 0 and add the first set of dapps in the script? Then chenage the timelock to its correct value and set the multisig as the owner?

deployment/8_update_ens.js
```diff
@@ -16,9 +21,12 @@ async function main() {
   const { deploymentAccount, configurator, abiUploader } = await deployManager.getProps();
   const { config } = configurator;
   const { domain } = config.ENS;
+
+  WalletFactory16.setProvider(web3.currentProvider);
+
   // ENVIRONMENT SPECIFIC
   // ROPSTEN
-  const walletFactoryVersion16 = "0x802248Ec4d68879Ce62d33748776Db5a1a98C422";
+  const walletFactory16Wrapper = await WalletFactory16.at("0x802248Ec4d68879Ce62d33748776Db5a1a98C422");
```
juniset, line 29: Maybe add a `walletFactoryVersion16` property to the config file?

deployment/7_deploy_2.5.js
elenadimitrova: Sure, #208 already does that so when that's merged I'll rebase to update here.

PR #213, opened 2021-03-09
Lido filter

elenadimitrova:
Adds DappRegistry Filter for staking in Lido as well as selling stETH via Curve. Additionally we introduce integration testing using a mainnet fork.


contracts/infrastructure/dapp/LidoFilter.sol
```diff
@@ -0,0 +1,36 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.6.12;
+
+import "./BaseFilter.sol";
+
+contract LidoFilter is BaseFilter {
+  bytes4 private constant DEPOSIT = bytes4(keccak256("depositBufferedEther()"));
```
olivdb, line 22: The method that we want to approve is the payable `submit(address)` method. `depositBufferedEther()` is used by Lido keepers to periodically move the ETH deposited via `submit` to the ETH2 deposit contract.

contracts/infrastructure/dapp/LidoFilter.sol
```diff
@@ -0,0 +1,36 @@
...
+
+  function isValid(address _wallet, address _spender, address _to, bytes calldata _data) external view override returns (bool) {
+    // Allow sending ETH as well as calls to depositBufferedEther()
+    if (_data.length == 0) {
+        return true;
+    }
+    if(_spender == _to) {
```
olivdb, line 29: We probably need to add `&& _data.length >= 4` here since `getMethod` doesn't do that check

contracts/infrastructure/dapp/LidoFilter.sol
elenadimitrova: I have added explicit calls to `submit`  it in addition to the fallback which implicitly calls it as well.

contracts/infrastructure/dapp/LidoFilter.sol
```diff
@@ -0,0 +1,36 @@
...
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.6.12;
+
+import "./BaseFilter.sol";
+
+contract LidoFilter is BaseFilter {
+  bytes4 private constant SUBMIT = bytes4(keccak256("submit(address)"));
+
+  function isValid(address /*_wallet*/, address _spender, address _to, bytes calldata _data) external view override returns (bool) {
+    // Allow sending ETH as well as calls to send(address)
```
olivdb, line 25: send(address) -> submit(address) ?

contracts/infrastructure/dapp/LidoFilter.sol
```diff
@@ -0,0 +1,36 @@
...
+contract LidoFilter is BaseFilter {
+  bytes4 private constant SUBMIT = bytes4(keccak256("submit(address)"));
+
+  function isValid(address /*_wallet*/, address _spender, address _to, bytes calldata _data) external view override returns (bool) {
+    // Allow sending ETH as well as calls to submit(address)
+    if (_data.length == 0) {
+        return true;
+    }
+    if(_spender == _to && _data.length >= 4) {
+      bytes4 methodId = getMethod(_data);
```
juniset, line 30: Could be written as:
```
if(_spender == _to && _data.length >= 4) {
    return (getMethod(_data) == SUBMIT);
}
```
but not sure it will save anything.

juniset (commented):
Shouldn't we also implement selling StETH for ETH using Curve as currently implemented in the app?

Could also be a separate PR.

contracts/infrastructure/dapp/CurveFilter.sol
```diff
@@ -0,0 +1,36 @@
...
+  bytes4 private constant EXCHANGE = bytes4(keccak256("exchange(int128,int128,uint256,uint256)"));
+  bytes4 private constant ERC20_APPROVE = bytes4(keccak256("approve(address,uint256)"));
+
+  function isValid(address /*_wallet*/, address _spender, address _to, bytes calldata _data) external view override returns (bool) {
+    if (_data.length < 4) {
+        return false;
+    }
+    bytes4 methodId = getMethod(_data);
+    if(_spender == _to) {
+        return (methodId == EXCHANGE);
```
juniset, line 31: Not sure if it is necessary, but maybe we should check that we are swapping stETH for ETH, and not the other way around. Can't imagine how it can be harmful, but our intention with this filter is to only swap stETH for ETH so we might as well enforce it in the code.

If we decide to do so (e.g. by passing the stETH address in the constructor) then we could also check that `_to == stETH` when the method is ERC20_APPROVE).

juniset (approved):

olivdb (approved):

PR #216, opened 2021-03-15
Configuration schema updates to include filters

elenadimitrova:
Miscellaneous cleanup
- configuration schema updates to include filters
- slither db updates to remove the warning for immutable implementation in the wallet Proxy
- deployment fixes for the ArgentWalletDetector and increasing the gas limit for deployments. Also exclude unsupported filters from deployments to Ropsten
- fix for the missing `cancelGuardianRevokation` in `getRequiredSignatures`


utils/config-schema.json
```diff
@@ -78,6 +78,15 @@
           "properties": {
             "tub": {
               "$ref": "#/definitions/ethaddress"
+            },
+            "pot": {
+              "$ref": "#/definitions/ethaddress"
+            },
+            "jug": {
```
olivdb, line 85: Not currently required until we implement Maker Vault filters

utils/config-schema.json
```diff
@@ -99,6 +111,63 @@
...
+        "paraswap": {
+          "type": "object",
+          "properties": {
+            "contract": {
+              "$ref": "#/definitions/ethaddress"
+            },
+            "authorisedExchanges": {
+              "type": "object",
+              "properties": {
+                "Kyber": {
```
olivdb, line 124: Could we keep this empty? Kyber was just used as a placeholder of an example Dex in `dev`.

utils/config-schema.json
```diff
@@ -78,6 +78,15 @@
           "properties": {
             "tub": {
```
olivdb, line 79: `tub` should be optional (along with `vat`, `daiJoin`, `wethJoin`, `cdpManager`) as they are only used in `dev` if provided, to built the mock migration contract.

scripts/deploy_wallet_detector.js
```diff
@@ -2,33 +2,46 @@
...
 // ////////////////////////////////////////////////////////////////////
 
 /* global artifacts */
+global.web3 = web3;
 const ArgentWalletDetector = artifacts.require("ArgentWalletDetector");
 
-const { configurator, abiUploader } = require("deploy-manager.js");
+const deployManager = require("../utils/deploy-manager.js");
 
 const PROXYWALLET_CODEHASH = [
```
juniset, line 17: I believe we have introduced a new `Proxy` when updating the `WalletFactory` to enable refund, and we are introducing a new one now with the `implementation` immutable. 

We shouldn't redeploy the `ArgentWalletDetector` but we need to update the existing instances and add these 2 `Proxy` bytecodes.

elenadimitrova: This script was needed for `test` only which is why I kept it separate and not part of the 2.5 upgrade script which only calls `addImplementation`

contracts/modules/common/Utils.sol
```diff
@@ -31,6 +31,7 @@ library Utils {
     bytes4 private constant ERC721_SAFE_TRANSFER_FROM_BYTES = bytes4(keccak256("safeTransferFrom(address,address,uint256,bytes)"));
     bytes4 private constant ERC1155_SAFE_TRANSFER_FROM = bytes4(keccak256("safeTransferFrom(address,address,uint256,uint256,bytes)"));
 
+    bytes4 private constant OWNER_SIG = 0x8da5cb5b;
```
juniset, line 34: Maybe use `bytes4 private constant OWNER_SIG = bytes4(keccak256("owner()"));` to be consistent with our other constants?

deployment/7_deploy_2.5.js
```diff
@@ -133,35 +133,49 @@ const main = async () => {
...
-
-  // Balancer
-  console.log("Deploying BalancerFilter");
-  const BalancerFilterWrapper = await BalancerFilter.new();
-  console.log(`Deployed BalancerFilter at ${BalancerFilterWrapper.address}`);
-  for (const pool of (config.defi.balancer.pools)) {
-    console.log(`Adding filter for Balancer pool ${pool}`);
-    await DappRegistryWrapper.addDapp(0, pool, BalancerFilterWrapper.address);
-  }
+  // The following filters can't be setup on Ropsten due to tha lack of integrations
```
juniset, line 136: I believe we should also add the `refundAddress` to the DappRegistry with no filter.

contracts/modules/common/Utils.sol
elenadimitrova: This raises a `TypeError: Only direct number constants and references to such constants are supported by inline assembly.` error

juniset: We could probably add `bytes4 sig = OWNER_SIG;` and use `sig` in the assembly to fix the error but no need to add the extra gas.

If we do one more iteration on this PR then we could add `bytes4(keccak256("owner()"));` above the `OWNEER_SIG` declaration to make the code more readable.

juniset (approved):

deployment/7_deploy_2.5.js
juniset: My mistake, already added.

PR #219, opened 2021-03-17
immutable implementation

juniset:
This PR:

- makes the `implementation` variable in `Proxy.sol` immutable to save ~2k gas on every call.
- increases the gas stipend in `Utils.isGuardianOwner` from 5k to 25k to follow the gas increase due to EIP 2929
- updates the hardcoded values used for the gas refund


olivdb (approved):

contracts/wallet/Proxy.sol
```diff
@@ -24,7 +24,7 @@ pragma solidity ^0.6.12;
  */
 contract Proxy {
 
-    address implementation;
+    address immutable public implementation;
```
elenadimitrova, line 27: Just a reminder to remove that ignore from the `slither` db as it's been asking us to do this.

test/ens.js
```diff
@@ -358,7 +362,8 @@ contract("ENS contracts", (accounts) => {
     });
 
     it("should support registering ens for wallets created using the legacy wallet factory v1.6", async () => {
-      const factory = await WalletFactoryV16.new(registry.address, walletImplementation.address, ensManager.address);
+      const walletImpl = await BaseWalletV16.new();
+      const factory = await WalletFactoryV16.new(registry.address, walletImpl.address, ensManager.address);
```
elenadimitrova, line 366: 👍

elenadimitrova: Also we need to restore the BaseWallet legacy tests but can be done in a separate PR.

elenadimitrova (approved):

PR #220, opened 2021-03-19
Audit report fixes

elenadimitrova:
Fixes the preliminary audit notes sent

### DIFF
--- a/contracts/infrastructure/DappRegistry.sol
+++ b/contracts/infrastructure/DappRegistry.sol
@@ -206,6 +206,9 @@ contract DappRegistry is IAuthoriser {
     function removeDapp(uint8 _registryId, address _dapp) external onlyOwner(_registryId) {
         require(authorisations[_registryId][_dapp] != bytes32(0), "DR: unknown dapp");
         delete authorisations[_registryId][_dapp];
+        if(pendingFilterUpdates[_registryId][_dapp] != bytes32(0)) {
+            delete pendingFilterUpdates[_registryId][_dapp];
+        }
         emit DappRemoved(_registryId, _dapp);
     }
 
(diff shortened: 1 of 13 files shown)


contracts/modules/RelayerManager.sol
```diff
@@ -92,6 +92,8 @@ abstract contract RelayerManager is BaseModule, SimpleOracle {
         external
         returns (bool)
     {
+        require(!_isLocked(_wallet) || _gasPrice == 0, "RM: Locked wallet refund");
```
juniset, line 95: We should move this check a few lines down (e.g. after `require(verifyData(_wallet, _data), "RM: Target of _data != _wallet");` so that it is included in the gas refund accounting.

contracts/infrastructure/DappRegistry.sol
```diff
@@ -206,6 +206,9 @@ contract DappRegistry is IAuthoriser {
     function removeDapp(uint8 _registryId, address _dapp) external onlyOwner(_registryId) {
         require(authorisations[_registryId][_dapp] != bytes32(0), "DR: unknown dapp");
         delete authorisations[_registryId][_dapp];
+        if(pendingFilterUpdates[_registryId][_dapp] != bytes32(0)) {
```
juniset, line 209: Wouldn't it be more efficient to just delete and skip the `if`?

contracts/modules/common/Utils.sol
```diff
@@ -65,7 +65,7 @@ library Utils {
     * @param _to The target contract.
     * @param _data The data payload.
     */
-    function recoverSpender(address _wallet, address _to, bytes memory _data) internal pure returns (address) {
+    function recoverSpender(address _wallet, address _to, bytes memory _data) internal pure returns (address spender) {
```
elenadimitrova, line 68: `_wallet` param is no longer necessary.

contracts/modules/common/Utils.sol
```diff
@@ -75,23 +75,29 @@ library Utils {
             if(
```
elenadimitrova, line 75: For extracting the `methodId` shall we use the internal `functionPrefix` function we already have in the library?

olivdb: We could but `functionPrefix` contains an unnecessary data size check and overall slightly increases our gas cost (by 24 to 93 gas in assets.js tests). Since `recoverSpender` is called in every subcall, I'd lean towards saving that small cost.

juniset (approved):

PR #221, opened 2021-03-24
Filter for Paraswap V4

olivdb:
Filter supporting the following AugustusSwapper V4 methods:
- multiSwap
- simpleSwap
- swapOnUniswap
- swapOnUniswapFork
- megaSwap

Tests use a custom mock of [AugustusSwapper](https://github.com/argentlabs/argent-contracts/blob/paraswapV4/lib_0.7/paraswap/V4/AugustusSwapperMock.sol). The original `AugustusSwapper` cannot be compiled without optimisation, partly because the contract is too large, and partly because `simpleSwap()` uses a number of parameters that exceeds the stack size. As a result solidity-coverage cannot be used with `AugustusSwapper` because of https://github.com/sc-forks/solidity-coverage/issues/417. To solve that problem, we use a mock of `AugustusSwapper` that removes unused methods and catches calls to `simpleSwap()` in a `fallback` that decodes `msg.data` in two passes (to circumvent the stack size limit) before calling an internal `simpleSwapWithStruct()` using a single struct as argument.


contracts/infrastructure/dapp/ParaswapFilter.sol
```diff
@@ -33,55 +35,208 @@ interface IParaswap {
...
+    }
+
+    function isValidMegaSwap(address _wallet, bytes calldata _data) internal view returns (bool) {
+        (IParaswap.MegaSwapSellData memory sell) = abi.decode(_data[4:], (IParaswap.MegaSwapSellData));
+        return hasValidBeneficiary(_wallet, sell.beneficiary) &&
+            hasTradableToken(sell.path[0].path[sell.path[0].path.length - 1].to) && 
+            hasValidMegaPath(sell.path);
+    }
+
+    function hasAuthorisedCallees(
```
juniset, line 191: I'm concerned about the gas overhead of calling the DapRegistry for every callee. What's the benefit of this approach over having the callee information locally? In both cases adding a new callee will require a filter addition/update thus take 7 days.

contracts/infrastructure/dapp/ParaswapFilter.sol
```diff
@@ -33,55 +35,208 @@ interface IParaswap {
...
         return success && abi.decode(res, (bool));
     }
 
-    function hasValidExchangeAdapters(bytes calldata _data) internal view returns (bool) {
-        // Note: using uint256[5] instead of (address, address, uint, uint, uint) to avoid "stack too deep" issues
-        (,IParaswap.Path[] memory path) = abi.decode(_data[4:], (uint256[5], IParaswap.Path[]));
+    function hasValidExchangeAdapters(IParaswap.Path[] memory _path) internal view returns (bool) {
         (bool success,) = dexRegistry.staticcall(
-            abi.encodeWithSignature("verifyExchangeAdapters((address,uint256,(address,address,uint256,bytes,uint256)[])[])", path)
+            abi.encodeWithSignature("verifyExchangeAdapters((address,uint256,(address,address,uint256,bytes,uint256)[])[])", _path)
```
juniset, line 228: Feels a bit unfortunate to have both a DappRegistry and a DexRegistry... would be nice to being able to get rid of the latter.

contracts/infrastructure/dapp/ParaswapFilter.sol
```diff
@@ -33,55 +35,208 @@ interface IParaswap {
...
+    function getUniswapProxy() external view returns (address);
 }
 
 contract ParaswapFilter is BaseFilter {
 
-    bytes4 constant internal MULTISWAP = 0xcbd1603e; // bytes4(keccak256("multiSwap(...)"))
+    bytes4 constant internal MULTISWAP = bytes4(keccak256(
+        "multiSwap((address,uint256,uint256,uint256,address,string,bool,(address,uint256,(address,address,uint256,bytes,uint256)[])[]))"
+    ));
+    bytes4 constant internal SIMPLESWAP = bytes4(keccak256(
```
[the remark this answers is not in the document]
olivdb: No, we agreed with clients that only "selling" tokens will be supported at first (as is the case now)

contracts/infrastructure/dapp/ParaswapFilter.sol
olivdb: It avoids duplication of the registry logic and content. 

`simpleSwap` offers almost no extra security over directly doing the trade via a `multiCall` and can just be thought of as a light wrapper over `multiCall`. So I think `AugustusSwapper` should be treated like any other wallet with regard to the `DappRegistry`. 

Ultimately a `simpleSwap` will cost one extra call to the `DappRegistry` compared to an equivalent `multiCall`. We could at some point decide to have the client deconstruct the subcalls of a `simpleSwap` and make them become subcalls of a pure `multiCall` instead (bypassing AugustusSwapper) to save the extra `DappRegistry` call.

olivdb: That being said, we could add a method to the DappRegistry to check an array of (spend, to, data) tuples.

contracts/infrastructure/dapp/ParaswapFilter.sol
olivdb: We could get rid of `DexRegistry` by using filters that authorise only a specific account (in this case AugustusSwapper) to use them. To avoid having wasting gas on `IFilter(filter).isValid()` calls, we could add the rule that if a filter address is the address of the account for which the authorisation is requested, the authorisation should be granted:
```
function isAuthorised(address _wallet, address _spender, address _to, bytes calldata _data) external view override returns (bool) {
        /// ....
                    if(filter == address(0) || filter == _wallet || IFilter(filter).isValid(_wallet, _spender, _to, _data)) {
                        return true;
                    }
        }
```
That would let us verify the exchange adapters at roughly the same cost as we pay now.

olivdb: Actually for adapters I think it's fine to hardcode their list in the filter

contracts/infrastructure/dapp/ParaswapFilter.sol
```diff
@@ -33,55 +36,241 @@ interface IParaswap {
...
     }
 
-    function hasTradableToken(bytes calldata _data) internal view returns (bool) {
-        (, address _destToken) = abi.decode(_data[4:], (address, address));   
+    function isValidSimpleSwap(address _wallet, address _augustus, bytes calldata _data) internal view returns (bool) {
+        (,address toToken,, address[] memory callees,, uint256[] memory startIndexes,, address beneficiary) 
+            = abi.decode(_data[4:], (address, address, uint256[3],address[],bytes,uint256[],uint256[],address));
+        return hasValidBeneficiary(_wallet, beneficiary) &&
+            hasTradableToken(toToken) &&
+            hasAuthorisedCallees(_augustus, callees, startIndexes, _data);
```
juniset, line 192: Do we really need to check the callees given that the calls are executed by AugustusSwapper? In a nutshell what `simpleSwap` does is:
- 1/ transfer the `from` tokens from the wallet to AugustusSwapper
- 2/ execute calls with the callees
- 3/ check that it received the intended amount of `to` tokens
- 4/ transfer the received `to` tokens to the beneficiary

Given the above I have the impression that `hasValidBeneficiary` and `hasTradableToken` combined with check 3/ are sufficient to prevent the wallet from being abused irrespective of what the calls to the callees are. In other words the wallet doesn't care what AugustusSwapper does provided that the above 3 conditions are satisfied.

contracts/infrastructure/dapp/ParaswapFilter.sol
```diff
@@ -33,55 +36,241 @@ interface IParaswap {
...
     }
 
-    function hasValidBeneficiary(address _wallet, bytes calldata _data) internal pure returns (bool) {
-        (address beneficiary) = abi.decode(_data[228:], (address)); // skipping 4 + 7*32 = 228 bytes
-        return (beneficiary == address(0) || beneficiary == _wallet);
+    function isValidMultiSwap(address _wallet, bytes calldata _data) internal view returns (bool) {
+        (IParaswap.SellData memory sell) = abi.decode(_data[4:], (IParaswap.SellData));
+        return hasValidBeneficiary(_wallet, sell.beneficiary) &&
+            hasTradableToken(sell.path[sell.path.length - 1].to) && 
+            hasValidExchangeAdapters(sell.path);
```
juniset, line 184: Given the comment for `simpleSwap` I believe the `hasValidExchangeAdapters` check is not needed either.

contracts/infrastructure/dapp/ParaswapFilter.sol
```diff
@@ -33,55 +36,241 @@ interface IParaswap {
...
+            (factory == uniForkFactory2 && initCode == uniForkInitCode2) ||
+            (factory == uniForkFactory3 && initCode == uniForkInitCode3)
+        );
+    }
+
+    function isValidMegaSwap(address _wallet, bytes calldata _data) internal view returns (bool) {
+        (IParaswap.MegaSwapSellData memory sell) = abi.decode(_data[4:], (IParaswap.MegaSwapSellData));
+        return hasValidBeneficiary(_wallet, sell.beneficiary) &&
+            hasTradableToken(sell.path[0].path[sell.path[0].path.length - 1].to) && 
+            hasValidMegaPath(sell.path);
```
juniset, line 219: Same as above; is `hasValidMegaPath` really needed?

contracts/infrastructure/dapp/ParaswapFilter.sol
olivdb: The issue is that check 3/ is not really doable because a compromised owner can pass a `toAmount` value that is as low as they want and trivially satisfy check 3/

PR #222, opened 2021-03-25
Extend MultiCallHelper with isAuthorisedInRegistry function

elenadimitrova:
- Adds a  new helper function for the clients with the following signature: `isAuthorisedInRegistry(address _wallet, Call[] calldata _transactions, uint8 _registryId)` which returns `boolean` of whether all the transactions are authorised in the given registry
-  Makes the `sender` event param in `ToggledRegistry` event indexed

### DIFF
--- a/contracts/infrastructure/helper/MultiCallHelper.sol
+++ b/contracts/infrastructure/helper/MultiCallHelper.sol
@@ -113,6 +113,25 @@ contract MultiCallHelper {
         return MAX_UINT;
     }
 
+    function isAuthorisedInRegistry(address _wallet, Call[] calldata _transactions, uint8 _registryId) external view returns (bool) {
+        for(uint i = 0; i < _transactions.length; i++) {
+            address spender = Utils.recoverSpender(_transactions[i].to, _transactions[i].data);
+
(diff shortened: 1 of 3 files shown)


olivdb (approved):

contracts/infrastructure/helper/MultiCallHelper.sol
```diff
@@ -113,6 +113,25 @@ contract MultiCallHelper {
...
+
+            uint auth = uint(dappRegistry.authorisations(_registryId, spender)); 
+            uint validAfter = auth & 0xffffffffffffffff;
+            if (0 < validAfter && validAfter <= block.timestamp) { // if the current time is greater than the validity time
+                address filter = address(uint160(auth >> 64));
+                if(filter != address(0) && !IFilter(filter).isValid(_wallet, spender, _transactions[i].to, _transactions[i].data)) {
+                    return false;
+                }
+            } else {
+                return false;
```
sgc-code, line 128: i think we should remove this `return`, if the filter is not active yet, we should ignore it and continue checking the other filters

sgc-code: it's fine, sorry

sgc-code (approved):
LGTM

PR #224, opened 2021-03-26
Aave V1 filter

elenadimitrova:
Allow calls to deposit ETH via `LendingPool.deposit(address,uint256,uint16)`

And redeeming aETH via `aToken.redeem(uint256)`

Integration tests working with mainnet contracts:
```
LendingPoolCore: 0x3dfd23A6c5E8BbcFc9581d2E864a68feb6a076d3
LendingPool     0x398eC7346DcD622eDc5ae82352F02bE94C62d119
aETH Token      0x3a3A65aAb0dd2A17E3F1947bA16138cd37d08c04
```

Deployment configuration setup with Ropsten instances:
```
LendingPoolCore: 0x4295ee704716950a4de7438086d6f0fbc0ba9472
LendingPool     0x9E5C7835E4b13368fd628196C4f1c6cEc89673Fa
aETH Token      0x2433A1b6FcF156956599280C3Eb1863247CFE675
```

### DIFF
--- a/contracts/infrastructure/dapp/AaveETHTokenFilter.sol
+++ b/contracts/infrastructure/dapp/AaveETHTokenFilter.sol
@@ -1,0 +1,30 @@
+// Copyright (C) 2021  Argent Labs Ltd. <https://argent.xyz>
+
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.6.12;
+
+import "./BaseFilter.sol";
+
+contract AaveETHTokenFilter is BaseFilter {
+
+    bytes4 private constant REDEEM = bytes4(keccak256("redeem(uint256)"));
+
+    function isValid(address /*_wallet*/, address /*_spender*/, address /*_to*/, bytes calldata _data) external view override returns (bool) {
+        bytes4 method = getMethod(_data);
+
+        return (method == REDEEM);
+    }
+}
--- a/deployment/7_deploy_2.5.js
+++ b/deployment/7_deploy_2.5.js
@@ -31,8 +31,8 @@ const ScdMcdMigration = artifacts.require("ScdMcdMigration");
 const UniswapV2Filter = artifacts.require("UniswapV2UniZapFilter");
 const LidoFilter = artifacts.require("LidoFilter");
 const CurveFilter = artifacts.require("CurveFilter");
-const AaveV1Filter = artifacts.require("AaveV1Filter");
-const AaveETHTokenFilter = artifacts.require("AaveETHTokenFilter");
+const AaveV1LendingPoolFilter = artifacts.require("AaveV1LendingPoolFilter");
+const AaveV1ATokenFilter = artifacts.require("AaveV1ATokenFilter");
 
 const deployManager = require("../utils/deploy-manager.js");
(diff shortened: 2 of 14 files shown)


utils/config-schema.json
```diff
@@ -122,6 +122,12 @@
           "properties": {
             "contract": {
               "$ref": "#/definitions/ethaddress"
+            },
+            "lendingPool": {
+              "$ref": "#/definitions/ethaddress"
+            },
+            "aaveToken": {
```
olivdb, line 129: There are several aTokens so I think this should be an array `aTokens` of addresses.

deployment/7_deploy_2.5.js
```diff
@@ -203,6 +205,16 @@ const main = async () => {
...
 
+  // Aave V1
+  console.log("Deploying AaveV1");
+  const AaveV1FilterWrapper = await AaveV1Filter.new();
+  console.log(`Deployed AaveV1Filter at ${AaveV1FilterWrapper.address}`);
+  await DappRegistryWrapper.addDapp(0, config.defi.aave.lendingPool, AaveV1FilterWrapper.address);
+
+  const AaveETHTokenFilterWrapper = await AaveETHTokenFilter.new();
+  console.log(`Deployed AaveETHTokenFilter at ${AaveETHTokenFilterWrapper.address}`);
+  await DappRegistryWrapper.addDapp(0, config.defi.aave.aaveToken, AaveETHTokenFilterWrapper.address);
```
olivdb, line 216: I think we should add that filter for each aToken in `config.defi.aave.aTokens`

contracts/infrastructure/dapp/AaveETHTokenFilter.sol
```diff
@@ -0,0 +1,30 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.6.12;
+
+import "./BaseFilter.sol";
+
+contract AaveETHTokenFilter is BaseFilter {
```
olivdb, line 21: Why `AaveETHTokenFilter`? Should we name this filter `AaveV1ATokenFilter` (and the other one `AaveV1LendingPoolFilter` maybe)?

deployment/7_deploy_2.5.js
```diff
@@ -203,6 +205,16 @@ const main = async () => {
   console.log(`Deployed UniswapV2Filter at ${UniswapV2FilterWrapper.address}`);
   await DappRegistryWrapper.addDapp(0, config.defi.uniswap.unizap, UniswapV2FilterWrapper.address);
 
+  // Aave V1
+  console.log("Deploying AaveV1");
```
olivdb, line 209: I think we should also authorise the [LendingPoolCore] (https://etherscan.io/address/0x3dfd23A6c5E8BbcFc9581d2E864a68feb6a076d3) contract (used as the target of an ERC20 approval prior to depositing the ERC20 into the lending pool) with our OnlyApproveFilter.

olivdb: And we could also add that "ERC20 deposit" use case to our tests as well

olivdb (commented):
Could we add simple unit tests of this filter (and the Lido filter)? I think it's fine to just use dumb mock contracts that don't do anything, just so we don't reduce our `infrastructure` coverage score.

utils/config-schema.json
elenadimitrova: Okay, only did that because Tom was saying we only support aETH atm. Is that not the case?

elenadimitrova:
I was thinking of doing the opposite for testing, i.e. move all filters tests to be integration and run coverage on those

olivdb:
> I was thinking of doing the opposite for testing, i.e. move all filters tests to be integration and run coverage on those. It'll shrink out lib and test contracts estate significantly and also reduce risk we get out of date with production and also flag integration issues early.

I don't think we should get rid of (non-integration) unit tests for filters. Unit tests run faster and are not at risk of breaking when some state changes in a third party contract. Also, in some cases we are going to have a proactive integration of a protocol that is not yet deployed on any testnet or mainnet (as is the case for Paraswap).

elenadimitrova:
In terms of speed, integration tests are faster end-to-end because a) they don't need us to compile mocked contracts and b) we reuse existing instances rather than create new ones. Proactive integration is fine, but that's more the exception than the rule for filters development.

olivdb (approved):

PR #227, opened 2021-03-30
Removed all price logic from TokenPriceRegistry and renamed it TokenRegistry

juniset:
This PR:
- Removed all price logic from `TokenPriceRegistry` and renamed it `TokenRegistry` as only the `tradable` flag is used by the new contracts.
- Introduced a new `TestSimpleOracle` contract that can be passed to the `relay-manager.js` to convert ETH to Token when the refund is in ERC20.
- Cleaned all test files from a price oracle when not needed.
- Updated the deployment script to deploy the new `TokenRegistry`.


contracts-test/TestSimpleOracle.sol
```diff
@@ -0,0 +1,45 @@
...
+pragma solidity ^0.6.12;
+
+import "../contracts/modules/common/SimpleOracle.sol";
+
+contract TestSimpleOracle is SimpleOracle {
+
+    bytes32 internal creationCode;
+
+    constructor(address _uniswapRouter) SimpleOracle(_uniswapRouter) public {
+        address uniswapV2Factory = IUniswapV2Router01(_uniswapRouter).factory();
```
olivdb, line 26: I think this line is unnecessary as you already inherit `uniswapV2Factory` from `SimpleOracle`

olivdb (approved):

PR #231, opened 2021-04-02
Fix compound and unizap filter

juniset:
- Compound: we explicitly list the cToken methods that the filter should accept
- Unizap: we check that the target pair has been flagged as `tradable` in the `TokenRegistry`

We also split the deployment of 2.5 into infrastructure and module to keep the script readable.


contracts/infrastructure/dapp/CompoundCTokenFilter.sol
```diff
@@ -37,8 +43,21 @@ contract CompoundCTokenFilter is BaseFilter {
         bytes4 method = getMethod(_data);
         // cToken methods
         if (_spender == _to) {
-            // block repayBorrowBehalf
-            return (method != CTOKEN_REPAY_BORROW_BEHALF);
+            if (underlying == address(0)) {
```
elenadimitrova, line 46: Shouldn't this address be checked for `0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE`?

elenadimitrova: Same for line 41 above.

juniset: We have moved to `0x00...00` for the ETH address in the new contracts but you are right that we still use `0xEe...eE` in the configuration.

I suggest to just update the config file for consistency?

elenadimitrova: okay I've updated the test config with the 0 address

elenadimitrova (approved):

PR #232, opened 2021-04-05
Remove legacy code

elenadimitrova:
- Add tests for legacy wallets.
- Remove the compile-from-source for legacy contracts and use pre-compiled artefacts. 

Also sneaking in here are the following maintenance changes:
- Fix an bad update to `BaseWallet` addresses in production in the `ArgentWalletDetector`
- Configure `ganache-cli` to run with 0 `gasPrice` for testing
- Remove `docker` from CI as we're not using it for contracts compilation any more
- Move `ArgentModuleTest` to a mocks subfolder to make it clear it is not part of the production wallet and only used for testing. This has to be kept in the contracts folder due to https://github.com/sc-forks/solidity-coverage/issues/620
- Review slither warnings and updates to db


juniset (commented):
For the sake of completeness we should probably also add the `Proxy.json` files of the different versions in `build-legacy`. 

Also, to be cleaner we could remove most of the content of the json files and only keep the `abi`, `compiler` `bytecode` and `deployedBytecode` properties.

juniset (approved):

PR #236, opened 2021-04-14
Paraswap sub-filters for simpleSwap

olivdb:
Includes filters that allow `simpleSwap` to interact with:
- Paraswap's custom UniV2 router (named "UniswapV3Router")
- ParaswapPools


contracts/infrastructure/dapp/paraswap/ParaswapUniV2RouterFilter.sol
```diff
@@ -0,0 +1,58 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.8.3;
+
+import "../BaseFilter.sol";
+import "./ParaswapUtils.sol";
+
```
juniset, line 21: I think we should briefly describe to which contract filters will be associated to make reviewing the code easier. Something like " This contract filters calls to the <name of the contract> currently deployed at <address of the contract>".

Took me some time to remember that this contract was associated to the `UniswapProxy` contract of Paraswap and not the official `UniswapRouter`...

juniset: BTW shouldn't the name of the contract be `ParaswapUniV3RouterFilter`?

contracts/infrastructure/dapp/paraswap/ParaswapUniV2RouterFilter.sol
```diff
@@ -0,0 +1,58 @@
...
+    function isValid(address /*_wallet*/, address /*_spender*/, address /*_to*/, bytes calldata _data) external view override returns (bool valid) {
+        // disable ETH transfer
+        if (_data.length < 4) {
+            return false;
+        }
+
+        bytes4 methodId = getMethod(_data);
+
+        if(methodId == SWAP) {
+            (, address[] memory path) = abi.decode(_data[4:], (uint256[2], address[]));
```
juniset, line 52: I think it would be cleaner to stick to the method signature and use   `(, ,address[] memory path) = abi.decode(_data[4:], (uint256, uint256, address[]))`, or use `address[] memory path = abi.decode(_data[68:], (address[]))` as an alternative.

contracts/infrastructure/dapp/paraswap/ZeroExV2Filter.sol
```diff
@@ -0,0 +1,58 @@
...
+            ParaswapUtils.ZeroExV2Order[] memory orders = abi.decode(_data[4:], (ParaswapUtils.ZeroExV2Order[]));
+            for(uint i = 0; i < orders.length; i++) {
+                if(!marketMakers[orders[i].makerAddress]) {
+                    return false;
+                }
+            }
+            return true;
+        }
+        
+        return methodId == ERC20_APPROVE;
```
juniset, line 56: Shouldn't we use `return methodId == ERC20_APPROVE && _spender != _to;` to be explicit about that fact that we are approving the target contract on an ERC20? Might not be strictly needed but it makes the intentions clearer at almost no extra gas cost.

As a side note, isn't 0x using an asset proxy contract instead of approving the exchange directly?

contracts/infrastructure/dapp/paraswap/ZeroExV4Filter.sol
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        bytes4 methodId = getMethod(_data);
+
+        if(methodId == FILL) {
+            ParaswapUtils.ZeroExV4Order memory order = abi.decode(_data[4:], (ParaswapUtils.ZeroExV4Order));
+            return marketMakers[order.maker];
+        } 
+        
+        return methodId == ERC20_APPROVE;
```
juniset, line 51: Same as above.

contracts/infrastructure/dapp/paraswap/ZeroExV2Filter.sol
```diff
@@ -0,0 +1,58 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.8.3;
+
+import "../BaseFilter.sol";
+import "./ParaswapUtils.sol";
+
+contract ZeroExV2Filter is BaseFilter {
```
juniset, line 22: Maybe we should add a reference to Paraswap in the name of the filter to make it explicit that this is not a generic ZeroX filter (in case we decide to add a generic ZeorX filter in the future). E.g. ParaswapPoolZeroxV2Filter?

contracts/infrastructure/dapp/paraswap/ParaswapUniV2RouterFilter.sol
olivdb: Agreed to add the description of the spender.

Regarding the name, I found the name given to this router by Paraswap to be too confusing. Uniswap V3 is coming out soon and this "UniswapV3Router" sounds like it's the official Uniswap V3 router when in fact it's a custom router made by Paraswap for Uniswap V2 (hence why I call it "Paraswap's UniV2 Router" i.e. `ParaswapUniV2Router`).

contracts/infrastructure/dapp/paraswap/ParaswapUniV2RouterFilter.sol
olivdb: `(, ,address[] memory path) = abi.decode(_data[4:], (uint256, uint256, address[]))` would work in this particular case but in many other instances I had to aggregates the initial method parameters in an array to avoid hitting a Stack Too Deep compilation error. In general, every additional decoded parameter adds to the stack size, even when its value is ignored.

`address[] memory path = abi.decode(_data[68:], (address[]))` doesn't work when the decoded param is an array because the offset of the array in the data would be incorrect.

contracts/infrastructure/dapp/paraswap/ZeroExV2Filter.sol
olivdb: Can add `_spender != _to`.

Not 0xV4, but you are right for 0xV2. Will remove the `return methodId == ERC20_APPROVE;` for it.

contracts/infrastructure/dapp/paraswap/ZeroExV4Filter.sol
olivdb: 0xV4 no longer use a separate proxy.

contracts/infrastructure/dapp/paraswap/ZeroExV2Filter.sol
olivdb: It is actually a generic filter with a whitelist of market makers. It could be used outside of Paraswap, to trade on ZeroEx with approved market makers.

juniset: The fact that we whitelist market makers is what makes it non generic, but I agree that we may want to add market makers outside Paraswap pools in the future. Maybe `WhitelistedZeroxV2Filter` ?

contracts/infrastructure/dapp/paraswap/ParaswapUniV2RouterFilter.sol
juniset: Ok makes sense.

deployment/7_deploy_2.5_infrastructure.js
```diff
@@ -114,20 +119,54 @@ const main = async () => {
...
+    await DappRegistryWrapper.addDapp(0, routers[i], ParaswapUniV2RouterFilterWrapper.address);
+  }
+
   // Paraswap Proxy
   console.log("Deploying OnlyApproveFilter");
   const OnlyApproveFilterWrapper = await OnlyApproveFilter.new();
   console.log(`Deployed OnlyApproveFilter at ${OnlyApproveFilterWrapper.address}`);
   const AugustusSwapperWrapper = await IAugustusSwapper.at(config.defi.paraswap.contract);
-  await DappRegistryWrapper.addDapp(0, await AugustusSwapperWrapper.getTokenTransferProxy(), OnlyApproveFilterWrapper.address);
+  const proxies = [await AugustusSwapperWrapper.getTokenTransferProxy(), ...Object.values(config.defi.paraswap.proxies || {})];
```
juniset, line 151: Did you add the ZeroxV2 asset proxy to the config?

contracts/infrastructure/dapp/paraswap/CurveFilter.sol
```diff
@@ -0,0 +1,37 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.8.3;
+
+import "../BaseFilter.sol";
+
+contract CurveFilter is BaseFilter {
```
juniset, line 26: Why didn't you update the existing `CurveFilter` to make it more generic? We should only have 1.

contracts/infrastructure/dapp/paraswap/CurveFilter.sol
```diff
@@ -0,0 +1,37 @@
...
+
+    function isValid(address /*_wallet*/, address /*_spender*/, address /*_to*/, bytes calldata _data) external view override returns (bool valid) {
+        // disable ETH transfer
+        if (_data.length < 4) {
+            return false;
+        }
+
+        bytes4 methodId = getMethod(_data);
+
+        return methodId == ERC20_APPROVE || methodId == EXCHANGE || methodId == EXCHANGE_UNDERLYING;
```
juniset: As mentioned earlier I would prefer to separate the cases where the target of the `_data` is the pool (`_to == _spender`) and where it is an ERC20 (`_to != _spender`) to make the filter logic more explicit.

juniset: Sorry, did not realise it was already addressed.

contracts/infrastructure/dapp/paraswap/CurveFilter.sol
olivdb: True, totally forgot about the old one.

deployment/7_deploy_2.5_infrastructure.js
olivdb: Done

juniset (approved):

PR #240, opened 2021-04-27
Supporting AugustusSwapper's utility functions for singleSwaps

olivdb:
Supporting AugustusSwapper's utility functions for singleSwaps


contracts/infrastructure/dapp/WethFilter.sol
```diff
@@ -0,0 +1,40 @@
...
+ * @author Olivier VDB - <olivier@argent.xyz>
+ */
+contract WethFilter is BaseFilter {
+
+    bytes4 private constant DEPOSIT = bytes4(keccak256("deposit()"));
+    bytes4 private constant WITHDRAW = bytes4(keccak256("withdraw(uint256)"));
+
+    function isValid(address /*_wallet*/, address _spender, address _to, bytes calldata _data) external view override returns (bool valid) {
+        // disable ETH transfer
+        if (_data.length < 4) {
```
juniset, line 33: Wallets should be able to deposit ETH via the fallback function so we should accept calls where `_data.length == 0`.

contracts/infrastructure/dapp/paraswap/ParaswapFilter.sol
```diff
@@ -269,14 +280,49 @@ contract ParaswapFilter is BaseFilter {
         uint256 exchangeDataOffset = 36 + abi.decode(_data[196:228], (uint256)); 
         address[] memory spenders = new address[](_callees.length);
         bytes[] memory allData = new bytes[](_callees.length);
-        for(uint256 i = 0; i < _callees.length; i++) {
+        uint256 i;
+        while(i < _callees.length) {
```
juniset, line 284: I think this doesn't work as you are looping over `callees.length` but you are reducing the length if the callee is augustus so you might never reach the last callee. Also you are not incrementing `i` when you reduce the size of `callees` so you will check the same callee twice. Maybe I'm wrong but it's just a sign that this logic is too complicated.

A simpler version of the same logic that I find simpler to read is
```
function hasAuthorisedCallees(
        address _augustus,
        address[] memory _callees,
        uint256[] memory _startIndexes,
        bytes calldata _data
    )
        internal
        view
        returns (bool)
    {
        // _data = {sig:4}{six params:192}{exchangeDataOffset:32}{...}
        // we add 4+32=36 to the offset to skip the method sig and the size of the exchangeData array
        uint256 exchangeDataOffset = 36 + abi.decode(_data[196:228], (uint256)); 
        address[] memory to = new address[](_callees.length);
        address[] memory spenders = new address[](_callees.length);
        bytes[] memory allData = new bytes[](_callees.length);
        uint256 i;
        while(i < _callees.length) {
            bytes calldata slicedExchangeData = _data[exchangeDataOffset+_startIndexes[i] : exchangeDataOffset+_startIndexes[i+1]];
            if (_callees[i] == _augustus) {
                if (slicedExchangeData.length >= 4 && getMethod(slicedExchangeData) == WITHDRAW_ALL_WETH) {
                    uint newLength = _callees.length - 1;
                    assembly {
                        mstore(to, newLength)
                        mstore(spenders, newLength)
                        mstore(allData, newLength)
                    }
                } else {
                    return false;
                }
            } else {
                to[i] = _callees[i];
                allData[i] = slicedExchangeData;
                spenders[i] = Utils.recoverSpender(_callees[i], slicedExchangeData);
            }
            i++;
        }
        return authoriser.areAuthorised(_augustus, spenders, to, allData);
    }
 ```

File: contracts/infrastructure/dapp/paraswap/ParaswapFilter.sol [hunk 3]
```diff
@@ -97,6 +97,8 @@ contract ParaswapFilter is BaseFilter {
     bytes4 constant internal MEGASWAP = bytes4(keccak256(
         "megaSwap((address,uint256,uint256,uint256,address,string,bool,(uint256,(address,uint256,(address,address,uint256,bytes,uint256)[])[])[]))"
     ));
+    bytes4 constant internal APPROVE = bytes4(keccak256("approve(address,address,uint256)"));
```
juniset, line 100: This is no longer needed.

contracts/infrastructure/dapp/paraswap/ParaswapFilter.sol
olivdb: Agreed that using a temporary array for callees is much cleaner

contracts/infrastructure/dapp/paraswap/ParaswapFilter.sol
```diff
@@ -137,6 +138,15 @@ contract ParaswapFilter is BaseFilter {
     bytes32 public immutable uniForkInitCode2; // linkswap
     bytes32 public immutable uniForkInitCode3; // defiswap
 
+    
+    // Stack Extension for hasAuthorisedCallees()
+    struct CalleeInfo {
```
juniset, line 143: Sholul be removed.

juniset (approved):

juniset (approved):

contracts/infrastructure/dapp/WethFilter.sol
Muzcool: 👌

Muzcool:
👌

Muzcool:
👌

ghost (approved):

PR #244, opened 2021-05-05
Gro filters

olivdb:
- Add filters for Gro
- Add benchmark/test for addToWhitlist via multicall


test/gro.js
```diff
@@ -0,0 +1,227 @@
...
+const ZERO_ADDRESS = ethers.constants.AddressZero;
+const SECURITY_PERIOD = 2;
+const SECURITY_WINDOW = 2;
+const LOCK_PERIOD = 4;
+const RECOVERY_PERIOD = 4;
+const AMOUNT = web3.utils.toWei("0.01");
+
+const RelayManager = require("../utils/relay-manager");
+
+contract("yEarn Filter", (accounts) => {
```
juniset, line 45: Should be Gro and not yEarn.

contracts/infrastructure/dapp/GroDepositFilter.sol
```diff
@@ -0,0 +1,45 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.8.3;
+
+import "./BaseFilter.sol";
+
+/**
+ * @title GroDepositFilter
+ * @notice Filter used for deposits to Gro Protocol
```
juniset, line 23: Maybe add the name of the target Gro contract (`DepositHandler`) while the deployed address is not available.

contracts/infrastructure/dapp/GroWithdrawFilter.sol
```diff
@@ -0,0 +1,42 @@
...
+pragma solidity ^0.8.3;
+
+import "./BaseFilter.sol";
+
+/**
+ * @title GroWithdrawFilter
+ * @notice Filter used for withdrawals from Gro Protocol
+ * @author Olivier VDB - <olivier@argent.xyz>
+ */
+contract GroWithdrawFilter is BaseFilter {
```
juniset, line 26: Maybe add the name of the target Gro contract (`WithdrawHandler`) while the deployed address is not available.

juniset (commented):
Would be good to have a deployment script for the Filter as well.

juniset (approved):

PR #249, opened 2021-05-11
YearnV2 filter

olivdb:
YearnV2 filter


contracts/infrastructure/dapp/YearnV2Filter.sol
```diff
@@ -0,0 +1,64 @@
...
+pragma solidity ^0.8.3;
+
+import "./BaseFilter.sol";
+
+/**
+ * @title YearnV2Filter
+ * @notice Filter used for deposits & withdrawals into/from YearnV2
+ * @author Olivier VDB - <olivier@argent.xyz>
+ */
+contract YearnV2Filter is BaseFilter {
```
juniset, line 27: Can you add the deployment address of one (or more) vaults in the description of the filter.

contracts/infrastructure/dapp/YearnV2Filter.sol
```diff
@@ -0,0 +1,64 @@
...
+    // (see https://github.com/yearn/yearn-vaults/blob/master/contracts/Vault.vy).
+    // Each Vyper method therefore results in N+1 ABI methods for N parameters
+    // These ABI methods are denoted (DEPOSIT|WITHDRAW){i} below where i is the number of parameters for the method
+
+    bytes4 private constant DEPOSIT0 = bytes4(keccak256("deposit()"));
+    bytes4 private constant DEPOSIT1 = bytes4(keccak256("deposit(uint256)"));
+   
+    bytes4 private constant WITHDRAW0 = bytes4(keccak256("withdraw()"));
+    bytes4 private constant WITHDRAW1 = bytes4(keccak256("withdraw(uint256)"));
+    bytes4 private constant WITHDRAW3 = bytes4(keccak256("withdraw(uint256,address,uint256)"));
```
juniset: Shouldn't the filter accept `WITHDRAW2 = bytes4(keccak256("withdraw(uint256,address)"));` with the default `maxLoss` and block `WITHDRAW3` where one can pass a large `maxLoss` and potentially decrease the wallet's value?

juniset (changes requested):
Would be good to have a deployment script.

contracts/infrastructure/dapp/YearnV2Filter.sol
olivdb: Blocking `WITHDRAW3` is a good idea, and means that whales may have to use guardians to liquidate large positions, which is reasonable.

`WITHDRAW1` already provides a default `maxLoss` so no need for `WITHDRAW2`.

juniset (approved):

PR #250, opened 2021-05-14
Curve Filter

olivdb:
Filter to add/remove liquidity on Curve


contracts/infrastructure/dapp/CurveFilter.sol
```diff
@@ -27,6 +27,15 @@ contract CurveFilter is BaseFilter {
 
     bytes4 private constant EXCHANGE = bytes4(keccak256("exchange(int128,int128,uint256,uint256)"));
     bytes4 private constant EXCHANGE_UNDERLYING = bytes4(keccak256("exchange_underlying(int128,int128,uint256,uint256)"));
+   
+    bytes4 private constant ADD_LIQUIDITY2 = bytes4(keccak256("add_liquidity(uint256[2],uint256)"));
```
juniset, line 34: Maybe add a comment to note that only one of the method is present on a given pool.

juniset (approved):

PR #255, opened 2021-07-09
Integration tests

delaaxe:


test-integration/filter-aaveV1.js
```diff
@@ -151,33 +68,20 @@ contract("AaveV1 Filter", (accounts) => {
...
-      const masterMinter = "0xe982615d461dd5cd06575bbea87624fda4e3de17";
-      await usdcToken.configureMinter(accounts[0], web3.utils.toWei("10000"), { from: masterMinter });
-      await usdcToken.mint(wallet.address, 1000);
-      const usdcTokenBalance = await usdcToken.balanceOf(wallet.address);
-      expect(usdcTokenBalance).to.eq.BN(1000);
-
-      const transactions = encodeCalls([
-        [usdcToken, "approve", [aaveLendingPoolCore, 1000]],
-        [aaveLendingPool, "deposit", [usdcToken.address, 1000, ""]]
+      expect(await argent.USDC.balanceOf(wallet.address)).to.gte.BN(1000);
```
olivdb: we could add `.be` to the chain to "sound" better: `expect(...).to.be.gte.BN(1000)`

test-integration/filter-aaveV1.js
```diff
@@ -1,147 +1,64 @@
...
-        "multiCall",
-        [wallet.address, transactions],
-        wallet,
-        [owner],
-        1,
-        ETH_TOKEN,
-        relayer);
-
-      const { success, error } = utils.parseRelayReceipt(txReceipt);
+      const receipt = await argent.multiCallRaw(wallet, transactions);
```
olivdb: we could use `argent.multiCall` instead of `argent.multiCallRaw` wherever possible to reduce verbosity

test-integration/filter-curve.js
```diff
@@ -0,0 +1,67 @@
+/* global artifacts */
+
+const ArgentContext = require("../utils/argent-context.js");
+
+const CurvePool = artifacts.require("CurvePoolMock");
```
olivdb, line 7: At some point it would be cool to be able to remove all these mock sources from the repo and instead get the ABIs of these contracts via precompiled artifacts. We currently do this only for a few Uniswap contracts (see `./scripts/provision_lib_artifacts.sh`) and for some of our legacy contracts (see `./test/basewallet.js`), but we could extent that to all third party contract dependencies.

test-integration/filter-lido.js
```diff
@@ -157,39 +56,28 @@ contract("Lido Filter", (accounts) => {
...
       data = curve.contract.methods.exchange(1, 0, 99, 1).encodeABI();
-      transaction = encodeTransaction(curve.address, 0, data);
-      transactions.push(transaction);
+      const curveTransaction = utils.encodeTransaction(curve.address, 0, data);
+      const transactions = [lidoTransaction, curveTransaction];
 
       const before = await utils.getBalance(wallet.address);
-      const txReceipt = await manager.relay(
-        module,
+      const txReceipt = await argent.manager.relay(
```
olivdb: could we use `argent.multiCall` here instead?

test-integration/filter-weth.js
```diff
@@ -0,0 +1,60 @@
...
+const WETH = artifacts.require("WETH9");
+
+contract("WETH Filter", (accounts) => {
+  let argent, wallet;
+  let weth;
+
+  before(async () => {
+    argent = await new ArgentContext(accounts).initialize();
+
+    weth = await WETH.at("0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2");
```
olivdb: we could use `argent.WETH.address`, or use `argent.WETH` if we use the `WETH` abi  instead of the `ERC20` abi in `argent-context.js`

test-integration/filter-yearn-v2.js
```diff
@@ -0,0 +1,44 @@
...
+
+    const filter = await YearnV2Filter.new();
+    await argent.dappRegistry.addDapp(0, vault.address, filter.address);
+  });
+
+  beforeEach(async () => {
+    wallet = await argent.createFundedWallet({ DAI: "1" });
+  });
+
+  const deposit = async () => argent.multiCall(wallet, [
```
olivdb: could we add before/after balance check?

test-integration/filter-yearn-v2.js
```diff
@@ -0,0 +1,44 @@
...
+
+  it("should allow deposits", async () => {
+    const { success, error } = await deposit();
+    assert.isTrue(success, `deposit failed: "${error}"`);
+  });
+
+  it("should allow withdrawals", async () => {
+    await deposit();
+    const { success, error } = await argent.multiCall(wallet, [
+      [vault, "withdraw", []]
```
olivdb: would be good to also test the other possible variations of deposit/withdraw methods that we support in our filter

utils/argent-context.js
```diff
@@ -0,0 +1,143 @@
...
+    return this;
+  }
+
+  // transfer tokens from mainnet whale addresses to a test address we control
+  async initializeTokens() {
+    this.DAI = await ERC20.at("0x6b175474e89094c44da98b954eedeac495271d0f");
+    this.WETH = await ERC20.at("0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2");
+    this.USDC = await ERC20.at("0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48"); 
+
+    if (tokensInitialized) {
```
olivdb: shouldn't this be at the beginning of the function?

utils/argent-context.js
```diff
@@ -0,0 +1,143 @@
...
+    await this.DAI.transfer(this.tokensAddress, web3.utils.toWei("10000"), { from: "0x6B175474E89094C44Da98b954EedeAC495271d0F" });
+    await this.WETH.transfer(this.tokensAddress, web3.utils.toWei("10000"), { from: "0x2F0b23f53734252Bda2277357e97e1517d6B042A" });
+    await this.USDC.transfer(this.tokensAddress, "10000000000", { from: "0x39AA39c021dfbaE8faC545936693aC917d5E7563" });
+
+    tokensInitialized = true;
+  }
+
+  // amounts: tokens amounts in natural units
+  async createFundedWallet(amounts) {
+    amounts = amounts || {};
```
olivdb: you could use a default 'empty object' value: 
```
async createFundedWallet(amounts = {})
```

utils/argent-context.js
```diff
@@ -0,0 +1,143 @@
...
+    const wallet = await BaseWallet.at(walletAddress);
+
+    // fund wallet in ETH
+    await wallet.send(web3.utils.toWei(amounts.ETH || "0.1"));
+
+    await utils.initNonce(wallet, this.module, this.manager, SECURITY_PERIOD);
+
+    // optionally fund wallet in ERC-20's
+    if (amounts.USDC) {
+      const amount = new BN(amounts.USDC).mul(new BN(1e6)).toString();
```
olivdb: We generally try to always use amounts in "wei" units in the code. So here your `amounts` could be expressed in weis so that you don't have to multiply them by `10^decimals`

utils/argent-context.js
```diff
@@ -0,0 +1,143 @@
...
+class ArgentContext {
+  constructor(accounts) {
+    [
+      this.infrastructure,
+      this.owner,
+      this.guardian1,
+      , // eslint-disable-line
+      this.relayer, // 4
+      , // eslint-disable-line
+      this.tokensAddress, // 6
```
olivdb: this could maybe be renamed into something like `tokensHolder` to make the intend clearer

utils/argent-context.js
olivdb: That would also allow you to avoid having distinct branches of code for USDC vs other tokens

test-integration/filter-aaveV1.js
delaaxe: Done

test-integration/filter-aaveV1.js
delaaxe: Done

test-integration/filter-curve.js
delaaxe: Good point, will add this to my todo

test-integration/filter-lido.js
delaaxe: Done

test-integration/filter-weth.js
delaaxe: done, used the WETH abi

test-integration/filter-yearn-v2.js
delaaxe: sure

test-integration/filter-yearn-v2.js
delaaxe: ok

utils/argent-context.js
delaaxe: here we want to do the transfer only once, but still instantiate the tokens for every test file. I've renamed the variable for more clarity

utils/argent-context.js
delaaxe: done

utils/argent-context.js
delaaxe: done, switched all units to wei

utils/argent-context.js
delaaxe: done

test-integration/filter-ens.js
```diff
@@ -0,0 +1,189 @@
...
+    ensRegistry = await ENSRegistry.at(await argentEnsManager.ensRegistry());
+    ensResolver = await ArgentENSResolver.at(await argentEnsManager.ensResolver());
+    ensReverse = await ReverseRegistrar.at(await argentEnsManager.getENSReverseRegistrar());
+
+    const filter = await Filter.new();
+    await argent.dappRegistry.addDapp(0, argentEnsManager.address, filter.address);
+    await argent.dappRegistry.addDapp(0, recipient, ZERO_ADDRESS);
+  });
+
+  describe("ENS Manager", () => {
```
juniset: I have the impression that all these tests are unit tests of the ArgentENSManager which should stay in `/test/ens.js` as they are today. Here we should only test the integration with a wallet (i.e. the calls initiated by a wallet and that go through the filter).

juniset (commented):
I think the dapp unit tests in `/test` are no longer necessary and can be removed.

test-integration/filter-ens.js
delaaxe: Makes sense, this one doesn't need to be duplicated

delaaxe:
Let me check that I haven't missed anything on the integration side

delaaxe:
> I think the dapp unit tests in `/test` are no longer necessary and can be removed.

I compared unit tests vs. integration tests:

- Similar:
    - Aave v2
    - Compound
    - Gro
    - Maker
    - Paraswap (some paths involving defiswap/linkswap may be less thoroughly tested)
    - Uniswap
    - Yearn v1
- Incomplete integration tests:
    - Balancer (missing various tests)
    - Curve (missing tests for methods: add_liquidity, remove_liquidity, exchange_underlying)
    - Yearn v2 (missing ETH deposit/withdrawal)
- Only in integration tests:
    - Aave v1
    - Lido
    - WETH

If you confirm, I can remove the unit tests for the "similar" dapps while I finish the incomplete ones.

@juniset @olivdb

delaaxe:
Removing dapp unit tests in #263 to avoid coverage drop

test-integration/filter-ens.js
```diff
@@ -0,0 +1,54 @@
...
+    await argentEnsManager.addManager(argent.infrastructure, { from: ARGENT_ENS_OWNER_ADDRESS });
+
+    ensRegistry = await ENSRegistry.at(await argentEnsManager.ensRegistry());
+    ensResolver = await ArgentENSResolver.at(await argentEnsManager.ensResolver());
+
+    const filter = await ArgentEnsManagerFilter.new();
+    await argent.dappRegistry.addDapp(0, argentEnsManager.address, filter.address);
+  });
+
+  it("should register an ENS name", async () => {
```
juniset, line 39: This tests that a manager of the `ArgentENSManager` can register the ENS of a wallet. That test should be in `test/ens.js` .

However, in `test-integration` we should test that a wallet can register his ENS via a multicall and the dapp registry. Something that is currently in `test/ens.js` under the `ENS Integrations` section...

utils/relay-manager.js
```diff
@@ -67,10 +67,13 @@ class RelayManager {
       + 4 * empty calldata bytes
       + 70k buffer
     */
-    let gas = gasLimit + 21000 + nonZeros * 16 + zeros * 4 + 70000;
+    let gas = gasLimit + 21_000 + nonZeros * 16 + zeros * 4 + 70_000;
```
juniset, line 70: Wow, I didn't know javascript supported numeric separators :-)

juniset (changes requested):

test-integration/filter-ens.js
delaaxe: Pushed new changes

utils/relay-manager.js
delaaxe: It gets tricky when there are lots of 0's :)

test-integration/filter-aave-v2.js
```diff
@@ -0,0 +1,83 @@
...
+    pool = await AaveV2LendingPool.at("0x7d2768dE32b0b80b7a3454c06BdAc94A69DDc7A9");
+    const filter = await AaveV2Filter.new();
+    await argent.dappRegistry.addDapp(0, pool.address, filter.address);
+  });
+
+  beforeEach(async () => {
+    wallet = await argent.createFundedWallet({ DAI: amount });
+  });
+
+  const deposit = async (beneficiary) => argent.multiCall(wallet, [
```
olivdb, line 34: We could check that the balances have been updated as part of the `deposit` and `withdraw` methods.

test-integration/filter-balancer.js
```diff
@@ -0,0 +1,83 @@
...
+    pool = await BPool.at("0x1eff8af5d577060ba4ac8a29a13525bb0ee2a3d5");
+    const filter = await BalancerFilter.new();
+    await argent.dappRegistry.addDapp(0, pool.address, filter.address);
+  });
+
+  beforeEach(async () => {
+    wallet = await argent.createFundedWallet({ WETH: amount });
+  });
+
+  const deposit = async () => argent.multiCall(wallet, [
```
olivdb, line 28: We could check that the balances have been updated as part of the `deposit` and `withdraw` methods.

utils/utilities.js
```diff
@@ -312,7 +314,23 @@ const utilities = {
 
     const event = await utilities.getEvent(tx.receipt, factory, "WalletCreated");
     return event.args.wallet;
-  }
+  },
+
+  checkBalances: async (wallet, providedToken, receivedToken, body) => {
```
olivdb: Maybe worth using a more explicit name like `callContractAndAssertTokenExchange` ?

juniset (approved):

test-integration/filter-gro.js
```diff
@@ -0,0 +1,173 @@
...
+  });
+
+  beforeEach(async () => {
+    wallet = await argent.createFundedWallet({
+      DAI: daiAmount,
+      WETH: ethAmount,
+    });
+  });
+
+  const deposit = async (isPwrd) => argent.multiCall(wallet, [
```
olivdb, line 53: We could wrap this in your `checkBalances` function. Idem for the withdraw functions.

utils/paraswap/swap-helper.js
```diff
@@ -0,0 +1,282 @@
+const { assert, expect } = require("chai");
```
olivdb: Paraswap helper was named `sell-helper.js` because Paraswap codebase distinguishes a `buy` (acquiring a fixed amount of tokens by sending a variable amount of tokens) from a `sell` (sending a fixed amount of tokens to acquire a variable amount of tokens). Our app only uses the `sell` functions for now. So you could merge the content of `swap-helper` (which relates to `sell` functions) into `sell-helper`.

delaaxe: I agree, so can we remove the `buy-helper`?

Also whenever a file reaches 200-300 lines I think it ok to split over multiple files.

olivdb: Yes, `buy-helper` can be removed.

Agree with splitting large files but here it's used as a library of related methods so I think it's fine to keep it as one.

PR #272, opened 2021-09-10
cleaned the config in S3 to remove reference to filter

juniset:
Updated the configs in S3 to remove all the filter related properties that have been move to https://github.com/argentlabs/argent-trustlists.

Updated `set_tradable_tokens.js` accordingly.


delaaxe (changes requested):
Awaiting your changes to the config schema

delaaxe (approved):

